PR #13, opened 2018-10-19
Dockerfile for master

LynxAbraxas:
PR provides a `Dockerfile` to build and run (included `run.sh`, or e.g. [x11docker](https://github.com/mviereck/x11docker)) the game from a docker image, provided the necessary data from the CD (as described in the docs) is put into `ctp2CD/`. This can be done through a private submodule, which has to be created by one's own and then referenced by a new commit to track the different SHA of the new submodule. If both repos are pushed to GitLab, the docker image is build by GitLab CI according to the given `.gitlab-ci.yml`.
Currently, the Dockerfile is single staged (and not tuned for size) for debugging purposes and therefore results in a 700MB docker image, which could be reduced in size a lot when debugging is not needed any more. 

Current status:
Game starts (also fullscreen, i.e. `./run.sh ./ctp2 fullscreen`), supports sound and music, graphics resolution can be changed and will be saved in `userprofile.txt` to take effect the next time the game is started. `userprofile.txt` is mapped to `~/.civctp2/userprofile.txt` to persist over docker containers, however the initial settings concerning the map plugins saved by ctp2 (for an existing but empty `userprofile.txt`) are inappropriate, which leads to a assertion message window when a game's map is created. Therefore, start the game without mapping `userprofile.txt` and copy the `userprofile.txt` (created where the ctp2 executable resides) to `~/.civctp2/userprofile.txt` and then enable the mapping again.
Saved games are mapped to ~/.civctp2/save/ but ctp2 terminates (with 255) when loading a game, probably due to some assertion code. This needs some debugging.
If firefox is running when the game is started with `./run.sh` the ctp2 window is distorted, closing firefox solves this, can possibly be avoided with x11docker.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "ctp2CD"]
-	path = ctp2CD
-	url = ../ctp2CD.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ addons:
 compiler:
     - g++
 
+services:
+    - docker
+
 #Build steps
 before_install:
     - sudo unlink /usr/bin/gcc && sudo ln -s /usr/bin/gcc-5 /usr/bin/gcc
@@ -26,3 +29,4 @@ script:
     - ./autogen.sh
     - CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -w -fuse-ld=gold" ./configure --enable-silent-rules
     - make
+    - ./docker_build.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,15 @@ before_install:
     - sudo unlink /usr/bin/gcc && sudo ln -s /usr/bin/gcc-5 /usr/bin/gcc
     - sudo unlink /usr/bin/g++ && sudo ln -s /usr/bin/g++-5 /usr/bin/g++
 
-script: 
-    - ./autogen.sh
-    - CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -w -fuse-ld=gold" ./configure --enable-silent-rules
-    - make
-    - ./docker_build.sh
+jobs:
+  include:
+    - stage: build
+      name: "direct build"
+      script:
+        - ./autogen.sh
+        - CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -w -fuse-ld=gold" ./configure --enable-silent-rules
+        - make
+    - stage: build
+      name: "docker build"
+      script:
+        - docker build --pull .
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ RUN chown -R $USERNAME:$USERNAME /opt/
 USER $USERNAME
 
 COPY --chown=diUser:diUser ./ /ctp2/
-COPY --chown=diUser:diUser ctp2CD/ /opt/ctp2/
 
 RUN cd /ctp2 && \
     ./autogen.sh && \
@@ -30,11 +29,3 @@ RUN cd /ctp2 && \
     make && \
     make install
 
-
-RUN cp -r /ctp2/ctp2_data/ /opt/ctp2/
-
-RUN cp -v /ctp2/ctp2_code/mapgen/.libs/*.so /opt/ctp2/ctp2_program/ctp/dll/map/
-
-WORKDIR /opt/ctp2/ctp2_program/ctp/
-
-CMD ["./ctp2"]
--- a/docker_build.sh
+++ b/docker_build.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+tag="civctp2/civctp2:build"
+echo "Building $tag"
+docker build -t "$tag" . -f Dockerfile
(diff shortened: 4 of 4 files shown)


LynxAbraxas:
It seems the current travis config fails on checking out the (non-existent, to be private) ctp2CD submodule. Is it possible to disable this specific submodule checkout for ctp2CD in travis?

DEVoytas:
Hi @LynxAbraxas , thank you for sharing your `Dockerfile` !
I looked at some of the commits (not all of them) and posted some comments/question there. Please have a look if you have a while.
However after fetching this PR I realized that final version of this 'Dockerfile' does not work on my system. So I made few adjustment to make it work in my environment.
I will create alternative PR with mine adjustments so other can check if it works for them and possibly comment on it. I would be grateful if you find a while to have a look and tell what you think.

DEVoytas:
Oh!
I see that all my commit-specific comments appeared here!
I was actually expecting, that those will only be seen when you click the the link to the commit they belong to.

Anyway, the issue I was getting with original Dockerfile were:

```
$ git log --oneline -1
4609b56ab52c Merge branch for Dockerfile
$ docker build .
Sending build context to Docker daemon 315.8 MB
Step 1/14 : FROM ubuntu:16.04 as builder
Error parsing reference: "ubuntu:16.04 as builder" is not a valid repository/tag: invalid reference format
```
even after this change:

````
-FROM ubuntu:16.04 as builder
+FROM ubuntu:16.04
````

I still got error:

```
$ docker build .                                                                                                                                                                                                        
Sending build context to Docker daemon 315.8 MB
Step 1/14 : FROM ubuntu:16.04
 ---> 4a689991aa24
[...] # cut out some text here
 ---> 4a612f6fb98a
Step 8/14 : COPY --chown=diUser:diUser ./ /ctp2/
Unknown flag: chown
```
So new PR with my adjustments (that works for me) is here:
https://github.com/civctp2/civctp2/pull/15

I removed the part that is dependent on original CD data, so it's only building; no running/playing at the moment. We will need to figure out best approach for adding running/playing when it comes to adding external dependencies. I considered two options:
- passing arguments to docker with the location of the data from original CD
- creating separate Dockerfile project (so no extra external dependencies here) that would use build artifacts from builder docker.

Anyhow, I think this article may be useful for improving and  further development  of the container: https://docs.docker.com/develop/develop-images/multistage-build/

Please let me know your opinion on all of that.

LynxAbraxas:
It seems to me that your version of docker is older than mine and that of gitlab CI, which knows about `--chown` flag and the `as builder` statement for multi stage builds (intended for later work to reduce the docker image size).
My ctp2DF repos uses the submodule approach to avoid stale caches for both the ctp2 code repos and a relative and private ctp2CD repos. When merging that into civctp2 for a PR, I removed the submodule dependency on ctp2, since that is then given by the repos itself (see https://github.com/civctp2/civctp2/pull/13/commits/4609b56ab52c74af4ec97f4fdfbaf3156d73f959#diff-3254677a7917c6c01f55212f86c57fbfR20 or better in gitk where you get a three-way merge view). I preserved the submodule approach for ctp2CD because the CD data must not be made public. If you create a ctp2CD repos with the necessary CD data and push that and the civctp2 repos with the contribution of this PR to GitLab, it will automatically build a docker image that is even playable. I chose GL because it allows private repos for free and to access those as private submodules (in case of relative path) inside the docker environment without making the content of the private repos publicly available.
cpt2 is build and installed in the DF as an unprivileged user, because the executing user has to have write access to the install dir (dirty but that's the current state). This only works if the data copy is owned by that user as well, which is ensured by the  `--chown` flag. You could do a `RUN chown -R ...` in a later step, but that yields a very big docker layer again, which is nicely avoided with the   `--chown` flag to `COPY` and since GL's docker was fine with that, I did not consider to change that.

LynxAbraxas:
To clarify the reasons for the chosen strategy:
- The commits originate from my ctp2DF repos, which aims to provide a Dockerfile to build the sources and to run the game from the docker image. This allows to test the game (e.g. screenshot from the start screen) and allows to make sure that the working docker image is not replaced by a broken one. So if dependencies are broken (e.g. due to an update to the distro), there is still the old image left to use as long as docker can handle it. 
I chose GitLab because it offers CI, docker registry, and private repose all in one. This is necessary because a running version for testing and playing needs some data from the game CD. This resides in ctp2CD, which cannot be shared due to the policies on the initial release of the ctp2 sources. GitLab allows to have private repos, so if you create and populate your own ctp2CD repos on GL, sync the submodule in ctp2DF with that and then push all, GL will automatically build, test and release privately.
- By merging (instead of rebasing or squashing all relevant commits) I can preserve the relation to ctp2DF for offering further PRs easily.

If preferred, I can squash down all changes into one commit here for the PR, but even if I combined all comments into one, that info would hardly make any sense because the context each refers to would be lost. I have often found info about what did not work (and in what way it was tried) very helpful to avoid doing the same over and over again and in git many tiny commits have no significant effect on performance.

DEVoytas:
> If preferred, I can squash down all changes into one commit here for the PR, but even if I combined all comments into one, that info would hardly make any sense because the context each refers to would be lost. I have often found info about what did not work (and in what way it was tried) very helpful to avoid doing the same over and over again and in git many tiny commits have no significant effect on performance.

After your explanation, I have more sympathy to this arguments. However my main concern is still introducing git submodules, even if removed in subsequent commit, due to fact that git does not seem to handle well that use case (as explained in my other comment).

LynxAbraxas:
Then let's go for #23. If the history is needed it can be found in https://github.com/LynxAbraxas/ctp2DF (though the exact dependency will not be easily found any more).

PR #14, opened 2018-10-19
Dockerfile for linux-branch (as a reference, master now compiles and runs on linux)

LynxAbraxas:
PR similar to https://github.com/civctp2/civctp2/pull/13 just for the linux branch, which needed some additional commits, like ptitSeb's cifm and making use of it (to avoid a vfat partition inside docker).

Status:
Compiles, and game can be started from docker image, but suffers the same problem concerning the loading of a saved game.
SDL_ffmpeg does not compile with the changed API of libavformat, libswscale of ubuntu:14.04. Either compilation of a much older version of ffmpeg (< 1.0.0) or adjustments to SDL_ffmpeg according to the new API are needed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,26 @@
+image: docker:stable
+
+services:
+  - docker:dind
+
+before_script:
+  - docker login -u gitlab-ci-token -p $CI_JOB_TOKEN $CI_REGISTRY
+
+stages:
+  - build
+
+variables:
+  GIT_SUBMODULE_STRATEGY: recursive
+  IMAGE_TAG: $CI_REGISTRY_IMAGE/$CI_COMMIT_REF_NAME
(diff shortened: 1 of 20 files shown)


LynxAbraxas:
Also terminates on the creation of a new game. This was not the case 2008 or 2012, so my guess is that either newer compiler/libs cause some assertion without error message, or that I missed to pick some of the cifm changes in https://github.com/civctp2/civctp2/commit/3285b196b128d9. @ptitSeb would you have a look if https://github.com/civctp2/civctp2/pull/14/commits/54064c932cf9e3a16c108b29246d581474df9a84 is missing some.

ptitSeb:
I'm in vacation with limited computer access.  I'll check when I'm back next week.

LynxAbraxas:
Closed in favor of #37.

PR #15, opened 2018-10-24
Add support for docker builds on master branch

DEVoytas:
This is based on @LynxAbraxas 's PR https://github.com/civctp2/civctp2/pull/13
with few adjustments that makes it work on my system.
This is only build stage for now, since running requires external data files (original CD data).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ addons:
 compiler:
     - g++
 
+services:
+    - docker
+
 #Build steps
 before_install:
     - sudo unlink /usr/bin/gcc && sudo ln -s /usr/bin/gcc-5 /usr/bin/gcc
@@ -26,3 +29,4 @@ script:
     - ./autogen.sh
     - CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -w -fuse-ld=gold" ./configure --enable-silent-rules
     - make
+    - ./docker_build.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+################################################################################
+# builder
+################################################################################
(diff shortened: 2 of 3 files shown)


LynxAbraxas:
Thanks for looking into docker builds with travis. I think that you can use my DF as is for travis; docker on your system might be a bit out of date, see also my comments on `--chown` at: https://github.com/civctp2/civctp2/pull/13#issuecomment-433048150. If you just copy over the final Dockerfile from my PR, you loose all the records/history on trial-and-error experience, which might be helpful to avoid going through those again in the future. How about first merging #13 and then rebase your changes/removes onto that? 
Also, even if the `.gitlab-ci.yml` has no meaning to GH, keeping it would be great for forks pushed to GL to make use of building a playable docker image for private use, also see my comments at: https://github.com/civctp2/civctp2/pull/13#issuecomment-433048150.

DEVoytas:
> docker on your system might be a bit out of date

Yes, you are exactly right! It was outdated. After updating, `COPY --chown` and build stage naming works as expected. Thanks!

DEVoytas:
> How about first merging #13 and then rebase your changes/removes onto that?

Sounds like a good idea, but I have two concerns:

1. I do not want make any decisions about what PR to merge, this should be up to the maintainer. And I think @ptitSeb agreed to temporary role at the beginning of  November so very soon.
1. I am of the opinion that no commit that fails any checks (in our case only travis at the moment) should be merged to master branch. So maybe it would be better idea to make this change on top of your PR and then merge the result (that passes travis check) to master. What do you think? Still final decision is up to maintainer IMO :)

LynxAbraxas:
Included in #13 and its squashed version #23, which both succeed on travis.

DEVoytas:
Closing, as this is now obsoleted by #23 and #43

PR #16, opened 2018-10-25
Docker builder fix

DEVoytas:
For some reason travis-ci docker build failed (despite building fine on
other system) with the following error [1]:

```
      CXX      nowin32.lo
    libtool: Version mismatch error.  This is libtool 2.4.6 Debian-2.4.6-0.1, but the
    libtool: definition of this LT_INIT comes from libtool 2.4.2.
    libtool: You should recreate aclocal.m4 with macros from libtool 2.4.6 Debian-2.4.6-0.1
    libtool: and run autoconf again.
    Makefile:448: recipe for target 'nowin32.lo' failed
```

[1] https://travis-ci.com/civctp2/civctp2/builds/89085604

Fix based on this SO post:
https://stackoverflow.com/questions/3096989/libtool-version-mismatch-error

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ addons:
 compiler:
     - g++
 
+services:
+    - docker
+
 #Build steps
 before_install:
     - sudo unlink /usr/bin/gcc && sudo ln -s /usr/bin/gcc-5 /usr/bin/gcc
@@ -26,3 +29,4 @@ script:
     - ./autogen.sh
     - CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -w -fuse-ld=gold" ./configure --enable-silent-rules
     - make
+    - ./docker_build.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+################################################################################
+# builder
+################################################################################
+FROM ubuntu:16.04 as builder
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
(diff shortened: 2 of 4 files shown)


LynxAbraxas:
Hm, thanks for finding a solution to that. Since docker is supposed to always give a well defined environment, this made me wonder what causes the problem. I think that running `./docker_build.sh` https://github.com/civctp2/civctp2/pull/16/commits/a05c01dbe113ffe9760c654f5c3c4d5acb6c1514#diff-354f30a63fb0907d4ad57269548329e3R32 after the make has the following side-effect:
- travis sets up a different environment than the DF
- ctp2 is build in the travis environment
- the base dir with all build files is passed to docker as the context
- files built under travis environment are rebuilt in the docker environment, which (due to different versions of libtool) lead to that error.

I'm not sure how to best set up building with docker in travis but for the chosen approach, I would recommend to stick to the former approach of my [ctp2DF](https://github.com/LynxAbraxas/ctp2DF) where a clean code base is ensured by using the submodule approach also for the code.
A very simple workaround would be to move  `./docker_build.sh` as the first line after `script:` https://github.com/civctp2/civctp2/pull/16/commits/a05c01dbe113ffe9760c654f5c3c4d5acb6c1514#diff-354f30a63fb0907d4ad57269548329e3R28 under the assumption that travis does no other changes to the base dir before it is passed to docker as the context.

LynxAbraxas:
Solved in #13 and its squashed version #23 (both succeed on travis) where the docker build has its on job and therefore an untouched/clean environment.

DEVoytas:
Closing as this is solved by #23

PR #17, opened 2018-10-25
Parallel build dependencies

LynxAbraxas:
This PR contains changes that resolve build dependencies for a parallel build with e.g. `make -j"$(nproc)"` and remove some redundant rules identified during the process.


LynxAbraxas:
Noticed during the compile tests of late that there are still some circular dependencies and that sometimes make breaks due to some missing file, so there are some dependencies that still need to be resolved.

PR #18, opened 2018-10-25
loading saved game now works

LynxAbraxas:
The crash was caused due to different definitions of size_t used to read e.g. the file size and inconsistent use of their specific counter part for writing.

### DIFF
--- a/ctp2_code/gs/fileio/GameFile.cpp
+++ b/ctp2_code/gs/fileio/GameFile.cpp
@@ -720,7 +720,7 @@ uint32 GameFile::Restore(const MBCHAR *filepath)
 
 	g_theProgressWindow->StartCountingTo( 70 );
 
-	size_t ulLen = 0;
+	uint32 ulLen = 0;
 	n = c3files_fread(&ulLen, sizeof(ulLen), 1, fpLoad);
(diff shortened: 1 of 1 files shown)


ctp2_code/gs/fileio/GameFile.cpp
```diff
@@ -720,7 +720,7 @@ uint32 GameFile::Restore(const MBCHAR *filepath)
 
 	g_theProgressWindow->StartCountingTo( 70 );
 
-	size_t ulLen = 0;
+	uint32 ulLen = 0;
```
DEVoytas: For long-term solution, it would probably be better to change other places to `size_t`[[1]](https://www.embedded.com/electronics-blogs/programming-pointers/4026076/Why-size-t-matters), to match what we had here, but that would probably much more work, so this is probably good enough for now since it fixes the issue. Good job! :)

[1] For rationale please see: [Why size_t matters](https://www.embedded.com/electronics-blogs/programming-pointers/4026076/Why-size-t-matters) article.

LynxAbraxas: Yes, it's just a quick fix. There are many, many size_t which all would need checking...

Issue #21, opened 2018-11-01
Can not get docker builds passing travis-ci check


DEVoytas:
PR https://github.com/civctp2/civctp2/pull/15 is currently (commit bf01e6cf68ecc2d3fbbed8de940e2ff3b517b606) failing travis-ci checks during docker build, with the following error message[[1]](https://travis-ci.com/civctp2/civctp2/builds/89882381#L4893):
```
make[3]: Entering directory '/ctp2/ctp2_code/os/nowin32'
  CXX      nowin32.lo
libtool: Version mismatch error.  This is libtool 2.4.6 Debian-2.4.6-0.1, but the
libtool: definition of this LT_INIT comes from libtool 2.4.2.
libtool: You should recreate aclocal.m4 with macros from libtool 2.4.6 Debian-2.4.6-0.1
libtool: and run autoconf again.
make[3]: *** [nowin32.lo] Error 63
make[2]: *** [all-recursive] Error 1
```
[1] link to travis build: https://travis-ci.com/civctp2/civctp2/builds/89882381#L4893

I think it may be related to hard-coded `libtool` version 2.4.6 in generated `ltmain.sh` file (see issue #20).

To investigate it, I created a [debug PR #19](https://github.com/civctp2/civctp2/pull/19)  that prints out `libtoolize` version used by travis-ci system, and used inside docker.

It turns out that travis-ci version is `2.4.2` (https://travis-ci.com/civctp2/civctp2/builds/89892005#L966)
and docker version is `2.4.6` https://travis-ci.com/civctp2/civctp2/builds/89892005#L2320

Docker version matches what is hard-coded in `ltmain.sh`.
Failure seem to be the result of mismatch with travis `libtoolize` version.

But why the travis `libtoolize` version appears during build inside the docker container?
Based on the above investigation, I would much rather expect direct build to fail (version mismatch) than docker failure (version match).

ptitSeb:
I'll try to have a look at this, but not sure I can before this WeekEnd.

DEVoytas:
Thank you @ptitSeb ! No rush.
BTW, do you still sustain your consent to be the acting maintainer, until we find someone else to step in?

LynxAbraxas:
Did you have a  look at my comment https://github.com/civctp2/civctp2/pull/16#issuecomment-433065489 on your initial PR? I think this is due to a not clean docker context after the travis build.
The docker build should have a clean environment and ideally its own travis job. That's what I tried to achieve in .

[...]

Issue #20, opened 2018-11-01
Generated ltmain.sh file is stored in the repository


DEVoytas:
Currently generated file `ctp2_code/libs/freetype-1.3.1/ltmain.sh` is committed to the repository.
It's usually considered good practice to generate this kind of files during build time, rather than committing generated files to the repository.

This script has hard-coded following version info:
```
PACKAGE=libtool
VERSION="2.4.6 Debian-2.4.6-0.1"
package_revision=2.4.6
```
which, can potentially lead to the issues when different `libtoolize` version is installed in the system.

Specifically, it may be the root-cause of this failure:
https://travis-ci.com/civctp2/civctp2/builds/89882381#L4895
when trying to test docker builds in travis-ci.

Error message:
```
libtool: Version mismatch error.  This is libtool 2.4.6 Debian-2.4.6-0.1, but the
libtool: definition of this LT_INIT comes from libtool 2.4.2.
libtool: You should recreate aclocal.m4 with macros from libtool 2.4.6 Debian-2.4.6-0.1
libtool: and run autoconf again.
```
comes from `ltmain.sh` file:
https://github.com/civctp2/civctp2/blob/1cbc9256b1d297720ca9f1e98c952a1c8224b70f/ctp2_code/libs/freetype-1.3.1/ltmain.sh#L2219-L2237

LynxAbraxas:
See https://github.com/civctp2/civctp2/pull/16#issuecomment-433065489

PR #22, opened 2018-11-05
Try to fix issue of Docker with freetype build

ptitSeb:
Should improve things about Docker build (issue #20 and #21)


LynxAbraxas:
Not needed for Docker build with #13 and its squashed version #23.

MartinGuehmann:
Is this actually still needed. And I am starting whether the freetype is actually used by CTP2, the Linux branch does not have it. The manuals don't seem to mention it.

LynxAbraxas:
This PR is not needed since #23 was merged and can be closed.

Freetype is needed for both master and linux branch under linux, and has to be built separately for the linux branch, see https://github.com/LynxAbraxas/ctp2DF/blob/linux/Dockerfile#L27-L40. The inclusion in the master branch works such that this is not necessary for the master branch, see https://github.com/LynxAbraxas/ctp2DF/blob/master/Dockerfile.

MartinGuehmann:
I see that the Dockerfile builds it from the repository. And if I build the Linux branch directly, which I haven't done so far do I have to install it myself. So far I haven't found it in the instruction of the readme.linux.

PR #25, opened 2018-11-17
revert some changes from commit 88eb70652910a3bbbae that cause a segfault

LynxAbraxas:
changes to:
ctp2_code/robot/aibackdoor/dynarr.h : caused a segfault of ctp2 when a new or a saved game was loaded (see tests of ctp2DF @ 6e70a4a96e82 and c45f8e9f474a)


LynxAbraxas:
@DEVoytas would you mind merging this? Without this PR, the executable from the linux branch is basically unusable and it is essential that this bug doe not get ported to master.

Issue #33, opened 2018-11-19
Accepted diplomatic proposals seem to have no effect


LynxAbraxas:
It seems that *diplomatic proposals* that get accepted seem to have no effect, in the master branch under linux. They work as expected in the linux branch. While this should not be related to the linux port, it would be great if someone could confirm that this probem does not exist under windows. This might be the actual cause for #29.
Rejections seem fine, at least a message about the rejection is presented in the *diplomatic proposals* window, while the window just closes when the proposel seems accepted. (I also noticed that the processing of computer players already continues before I close the proposal window, though it seems to wait until I hit "send", which suffices for proper logic processing.)

@DarkDust Do you think you could drop an email to the ctp2 code masters of old (like Martin Gühmann, ctplinuxfan, Maquiladora, Fromafar, richardh, J Bytheway) to point them at the new repos and the issue?

LynxAbraxas:
Seems to be solved with #35, but even with #36 the diplomatic proposal window closes after clicking Send without waiting for a click on OK if the diplomatic proposal was initiated by an AI player. The agreement seem to be effective though. In a way the OK button only seems to serve as a "close window" button while the actual processing of agreements already happens when Send is clicked, still I find this behavior a bit irritating.

LynxAbraxas:
It never happened to me with ctp2 on linux that an AI player offered to share advances, which happened a lot in ctp1. Has the AI lost the ability to do that or has it just become unlikely? Initiating and executing exchange of advances as a normal player does work though.

MartinGuehmann:
@LynxAbraxas, I don't think that it has anything to with any recent or any code change at all. This is, I think, CTP2 out of the box behavior. As far as I can see in automatic AI testing, is that the AIs ask for map exchanges, pollution reduction, cease fire and piece treaties. (These things pop up even if the visible player is a robot.) That's all, and nothing has changed here.

Actually, this is to be expected since we have worked from little to nothing on diplomacy.

LynxAbraxas:
Many thanks @DarkDust for the call!

Issue #2, opened 2018-10-12
Project maintainer wanted


civctp2:
Contribution to the project are on hold, until maintainer is found.
For details please refer to this discussion: https://github.com/ptitSeb/ctp2/issues/3#issuecomment-427801575

DEVoytas:
Since @ptitSeb said in this comment: https://github.com/civctp2/civctp2/issues/21#issuecomment-441194286 that we will not be able to act as a temporary maintainer, I propose two alternative solutions (until real maintainer is found):
- either @LynxAbraxas agrees for this role (he crated many very useful PRs so he is most active and motivated person in this project at the moment)
- I will start merging pending PRs, but since I (as stated many times before) do not feel authorized to make such decisions, I would only merge those, that we all agree on (but take into account I may have periods of lower activity)

@LynxAbraxas, @ptitSeb what do you think about it? Do yo have any other ideas?

ptitSeb:
I'm ok with that @DEVoytas .

LynxAbraxas:
Yes, pleas go ahead @DEVoytas! My current New Years resolution for 2019 is to not work on this any more (for at least a year). It has infected me again...

LynxAbraxas:
@MartinGuehmann: Would you mind to be maintainer? You have profound knowledge of civctp2 source and of its evolution.

MartinGuehmann:
For sure, I can do this, I just have to get familiar with GitHub.

So the tasks are merging the branches. We have the two mentioned here and there is also trunk it is 11 commits ahead of master. I don't exactly know what these commits are about.

MartinGuehmann:
> @MartinGuehmann that is great! Now we need to figure out how can I pass `civctp2` account credentials to you. Maybe email?

@DEVoytas, I guess email is fine. You can make GitHub display the email address, publically. But of course in the age of spam-bots, that's not a good idea. I can't make GitHub display it only to some people, so let's do it like this:

guehmann
at
t-online.de

> 
> PS.
> I understand the need of getting familiar with GitHub, but are you more or less familiar with git in general?

I played around with the Git stuff in the last days. So it is more complex than Subvision, but that shouldn't be much of a problem.

The first goal is to bring master into a releasable stage, at least for Windows. Of course master should also compile on Linux, but I haven't managed to do this, yet. It stuck at wldgen.cpp. I haven't checked the compiler. I just used what came with Kubuntu.

[...]

PR #35, opened 2018-11-20
Debug diplomacy

LynxAbraxas:
The cause for #33 seem to be the `intptr_t` changes from https://github.com/civctp2/civctp2/commit/3285b196b128d975bc560a8b731011399050d949. Those are not needed if "-fpermissive" is in CXXFLAGS for modern compilers (second commit) and then accepted diplomatic proposals work again.

### DIFF
--- a/ctp2_code/ui/interface/diplomacywindow.cpp
+++ b/ctp2_code/ui/interface/diplomacywindow.cpp
@@ -25,7 +25,7 @@
 // Modifications from the original Activision code:
 //
 // - Keep the embargo and war buttons enabled until confirmed by the player.
-// - Initialized local variables. (Sep 9th 2005 Martin Gï¿½hmann)
+// - Initialized local variables. (Sep 9th 2005 Martin Gühmann)
 //
 //----------------------------------------------------------------------------
 
@@ -556,9 +556,9 @@ void DiplomacyWindow::UpdateProposalList(ctp2_ListBox *propList, bool toPlayer)
 		const Response *response = NULL;
 		if(oldItem) {
 			if(toPlayer) {
-				oldSelectedPlayer = (intptr_t)oldItem->GetUserData();
+				oldSelectedPlayer = (sint32)oldItem->GetUserData();
 			} else {
-				oldSelectedPlayer = (intptr_t)oldItem->GetUserData();
+				oldSelectedPlayer = (sint32)oldItem->GetUserData();
 			}
 		}
 
@@ -575,13 +575,36 @@ void DiplomacyWindow::UpdateProposalList(ctp2_ListBox *propList, bool toPlayer)
 					sender = i;
 					receiver = player;
 					response = &Diplomat::GetDiplomat(player).GetResponsePending(i);
+
 					if (*response == Diplomat::s_badResponse)
+
 						prop = &Diplomat::GetDiplomat(i).GetNewProposalPending(player);
 					else
+
 						prop = &Diplomat::GetDiplomat(i).GetMyLastNewProposal(player);
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
 					if(*prop == Diplomat::s_badNewProposal)
 						continue;
+
 					propData = prop->detail;
+
 					if(!response) {
 						response = &Diplomat::GetDiplomat(i).GetResponsePending(g_selected_item->GetVisiblePlayer());
 						if(*response == Diplomat::s_badResponse) {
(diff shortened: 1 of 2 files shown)


LynxAbraxas:
@DEVoytas would you mind merging this? In contrast to all other current PRs this one actually adresse a serious flaw in the game logic that will even effect windows builds. Basicall master is buggy since https://github.com/civctp2/civctp2/commit/3285b196b128d975bc560a8b731011399050d949.

ctp2_code/ui/interface/diplomacywindow.cpp
```diff
@@ -591,16 +614,32 @@ void DiplomacyWindow::UpdateProposalList(ctp2_ListBox *propList, bool toPlayer)
...
+
+
+
+
+
+
+
+
+
+
```
DEVoytas: why so many blank lines above?

ctp2_code/ui/interface/diplomacywindow.cpp
```diff
@@ -2969,22 +3008,29 @@ void DiplomacyWindow::TabPanelActionCallback(aui_Control *control, uint32 action
 
 	if(action != static_cast<uint32>(ctp2_Tab::ACTION_ACTIVATED))
 		return;
-	sint32 section = (intptr_t)cookie;
+
+	sint32 section = (sint32)cookie;
+
```
DEVoytas: some of those blank lines (above and below) seem to be redundant to me, although I admit that this is very subjective (especially in absence of some sort of coding standard)

DEVoytas:
@LynxAbraxas would you mind if we let @ptitSeb to weight in before I do? Please see my comment here https://github.com/civctp2/civctp2/issues/2#issuecomment-442920028 why.

Meanwhile, maybe you could have a look at some minor comments/questions I added.

Also, I would really like if we could follow those [rules for commit messages](https://chris.beams.io/posts/git-commit/#seven-rules).

I always try to persuade everyone to follow suggestions this two links:
- [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) (longer)
- [Writing good commit messages](https://github.com/erlang/otp/wiki/writing-good-commit-messages)

because from my experience, it really helps when reading git history by others (or even oneself, later)

ptitSeb:
Did you check thoses changes on a 64bits build?

Also, if `-fpermissive` is now mandatory (all those `intptr_t` transtyping was there to avoid the use of this flag), I think it should be forced in `configure` script.

But a solution without the use of `-fpermissive` would be preferable. And I don't understand why those change have this effect, I would like to understand what's wrong.
Is there a simple way of reproducing the issue? Or a save somewhere?

ctp2_code/ui/interface/diplomacywindow.cpp
LynxAbraxas: Because they were there before https://github.com/civctp2/civctp2/commit/3285b196b128d975bc560a8b731011399050d949 ;-) Sadly the source is full of inappropriate white space. There are some commits that try to remove some, but do not catch all. I just reverted to the former state, so the white space came in again. If preferred, I can add another that removes them but I would not do it in this commit to really just revert that file to the former state. In the end, luckily white space does not introduce bugs in C/C++ ;-)

ctp2_code/ui/interface/diplomacywindow.cpp
LynxAbraxas: Sure they are, same reaseon as above. (Actually you can remove any new line in C as e.g. in this "International Obfuscated C Code Contest" (2004) winner:
https://www.ioccc.org/2004/anonymous.c
https://www.ioccc.org/2004/anonymous.hint
)

LynxAbraxas:
Yes, checked on 64-bit system and 64 bit build, i.e. docker image from this [DF](https://github.com/LynxAbraxas/ctp2DF/blob/11213337c99fa5851e09fc5f6c5176ae906e9c89/Dockerfile). 
It's not absolutely clear to me either why it matters but I could imaging a similar reason to why `size_t` is problematic, i.e. an `intptr_t` can be arch dependent while the game uses non arch dependent types (likely a de-reference of those pointers) at other places. Changing only some will introduce bugs. What puzzles me most is that it did not lead to a segfault but just silently continues. 
I would welcome changes that avoid `-fpermissive` (in case support for that flag is dropped in future) but they have to be done very carefully and tested well as this bug shows. Even though I can create quite many unit smoke test with sikuli (https://github.com/LynxAbraxas/ctp2DF/tree/master/tests), this one is particularly difficult to catch because I'm not sure if the RNG of ctp2 is deterministic such that a saved game leading to a diplomatic agreement will always succeed, such that e.g. you could check if a shared map appears on the screen.

MartinGuehmann:
@ptitSeb
 
> Also, if `-fpermissive` is now mandatory (all those `intptr_t` transtyping was there to avoid the use of this flag), I think it should be forced in `configure` script.
> 
> But a solution without the use of `-fpermissive` would be preferable. And I don't understand why those change have this effect, I would like to understand what's wrong.
> Is there a simple way of reproducing the issue? Or a save somewhere?

I don't know exactly what is wrong with this. But if you look at the design of the whole thing, then we use a void* to store whatever we like in it. That's the opposite of type safety. So the way to fix it is to fix the design. The first thing is to wonder whether we really want to do this. Unfortunately, it is Activision's design, so they already decided and it is a lot of work to change it. So keep the fundamental idea and implement it properly. This means you take a union and add all the needed fields so that you can store all the stuff you wanna store without casting. Unfortunately, this is also a lot of work since you have to change the function's signature. And since this function was inherited, you have to change a lot.

I already tried fixing the design, but I did not use a union. So, I changed a lot of files until I could compile it again and guess what it also broke something. So in the end, I abandoned that working copy.

PR #37, opened 2018-11-23
Dockerfile for linux-branch (as a reference, master now compiles and runs on linux)

LynxAbraxas:
PR similar to #14 (but no segfaults any more) just for the linux branch, which needed some additional commits, like ptitSeb's cifm and making use of it (to avoid a vfat partition inside docker).

Status:
Compiles, and game can be started (and tested) from docker image.
SDL_ffmpeg does not compile with the changed API of libavformat, libswscale of ubuntu:14.04 but it is possible to compile a much older version of ffmpeg (< 1.0.0) and SDL_ffmpeg (https://github.com/LynxAbraxas/ctp2DF/commit/b6947c06d4aef87ca424a4c2292a115b1fdda52b) to get also the in-game videos playing.


LynxAbraxas:
@MartinGuehmann This PR should not have been merged either. As there is only one commit of yours on top (https://github.com/civctp2/civctp2/commit/12be0bbfc5855c127b7f3f1def3c54c1b0cc66f6) I would suggest to do a hard reset of the linux branch to https://github.com/civctp2/civctp2/commit/5db46abfabbf44efa23557d789b01a28c81e9593 and cherry-picking https://github.com/civctp2/civctp2/commit/12be0bbfc5855c127b7f3f1def3c54c1b0cc66f6 on to that, i.e. the linux branch should not contain https://github.com/civctp2/civctp2/commit/9a2aa4864bef1256a9eea0ace8720ce391d03f1a and https://github.com/civctp2/civctp2/commit/f34c3feecbe5146c31e7c45234278862dd2eee87, which create a dependency on a git-submodule. 
A DF and GL-CI should be implemented from https://github.com/LynxAbraxas/civctp2/tree/docker4linux similar to #84.

MartinGuehmann:
Thank you for pointing me to this. Since, this is in the closed pull request section. I won't see it. And GitHub isn't here a great help, because I have to manually order these things by the last change, so that it won't show up at the top, where I would catch it. And of course, these email notifications aren't a great help either if I get too many.

This PR was a bit confusing, and I didn't see anything in it not to merge it. So it will be the hard reset. Unfortunately, I still uploaded it directly from my Linux working copy, with a PR thinks would be a bit nicer.

MartinGuehmann:
So I reset my Linux working copy, force pushed it to my own clone first for testing and since everything looked fine, I force pushed it to civctp2. For the visual studio project files, I created pull request #98. So that this goes in via the supposed way.

LynxAbraxas:
Looks right now. If #84 gets accepted, I can create a similar PR that will enable GL-CI for the linux branch, it's mostly the DF that differs.

MartinGuehmann:
Fine, it is just the question whether we need it for the Linux branch. At least that should then be easier. :)

PR #43, opened 2018-12-04
Minor Dockerfile updates to increase readability

DEVoytas:
This PR includes following changes to (hopefully) increase Dockerfile readability:
- try to obey 80-characters per line limit
- move '&&' shell operators the to new line, so it is easier to see
  where the next command starts
- use extra indentation for the parameters when dealing with multi-line
  commands
- change the username from to 'builder': it is not obvious to me what the `di` in `diUser ` stands for (docker image perhaps?), but by convention UNIX usernames are all lower-case, and also `builder` seems to be more explanatory
- explain in the comments why `ENV USERNAME` is not used in `COPY --chown=`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,29 +3,32 @@
 ################################################################################
 FROM ubuntu:16.04 as builder
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


DEVoytas:
Thank you for ACK @LynxAbraxas ! I will merge it than.

PR #58, opened 2019-01-22
Restore the old way to copy the build queue in CityData

MartinGuehmann:
This is a bug relating to copying CityData, my copy constructor wasn't clean, and makes the game crash when for instance you found a new city.

Actually, the whole copying is ugly
..\ctp2_code\gs\gameobj\BldQue.h
..\ctp2_code\gs\gameobj\bldque.cpp

Fixed an assert due to an rounding error
..\ctp2_code\gs\world\WrlEnv.cpp

Added the database indices in the comments
..\ctp2_data\default\gamedata\goods.txt


MartinGuehmann:
Plus it adds some comments to goods.txt: The database indices of the goods. And it fixes a rounding error that triggers an assert while adding goods to the world. And if the assert is now truly triggered, it reports which goods had a summed probably of more than 100% on which terrain.

PR #59, opened 2019-01-22
Add CONTRIBUTING.md file

DEVoytas:


MartinGuehmann:
So that is the file where we can put the more specific guideline.

What I don't get is why it still shows that some checks haven't completed, yet. If I look at the details the it seems to have passed the checks.

And actually, why does it trigger a build check? I thought this is a non-source file, so it wouldn't change anything on the build status.

MartinGuehmann:
OK, I merged it on GitHub. But still, when I uploaded project files of Visual Studio to the Linux branch, so that it is easier for me to browse on Windows the code, it rejected a test.

Issue #57, opened 2019-01-21
Commit f228215 broke the build


DEVoytas:
From the build history: https://travis-ci.com/civctp2/civctp2/builds it looks like commit https://github.com/civctp2/civctp2/commit/f22821505d20ff5bbb698dc3db9828c145b53f1e is the one that broke the build: https://travis-ci.com/civctp2/civctp2/builds/97939556

Build logs:
- https://travis-ci.com/civctp2/civctp2/jobs/171647537
- https://travis-ci.com/civctp2/civctp2/jobs/171647538

@MartinGuehmann, I think that the reason it went unnoticed is because the commits were pushed directly to master branch. The better workflow is to create the pull requests that are verified by travis-ci and only merge them to master branch when they pass all the checks.
Alternatively, if one does want to push commits directly rather than creating  PRs 
for every single commit, it's better to have dedicated branch for this (maybe 'dev'), that is also verified by travis-ci, and only merge it back to master when it's in known good state.
Other workflows are also possible but I think the main goal should be to keep master in known healthy state, that can be always trusted to produce good binary.

MartinGuehmann:
Of course master should be kept in a healthy state. Looks like a beginner mistake. But it shows a problem with the freetype library: Currently, if I clean the project the freetype library is not cleaned. So without a fresh install I couldn't have caught it. The other problem, I could have caught if I had build it on Linux. Of course, this shows that travis-ci can check whether it builds on Linux, but so far it can check Windows builds, which is also be needed. It is quite a pain in the ass to keep the thing compiling on all platforms, even if you can access them.

Ideally, I would like to have something like: Push to whatever designated branch. Check whether it builds, if yes push to master, automatically. That would minimize bureaucracy as much as possible. Well, I haven't tried so far but would a pull request open again if something new is pushed to that branch?

DEVoytas:
There already is the issue for setting up Windows automatic checks: https://github.com/civctp2/civctp2/issues/6.
One issue with merging to master automatically is, that it does not allow you to have any review or discussion on incoming changes.

MartinGuehmann:
Thanks for pointing that out. Since, having these many issues here, and I think it is still a moderate number,  it is hard to catch them all. And especially, what is in, exactly.

[...]

Issue #55, opened 2019-01-16
Linux version loads save games created on Windows and creates save games that load on Windows


MartinGuehmann:
Now, the Linux version creates and loads save games that are created on Windows. Unfortunately, this breaks save games you created while testing.

This is an issue with binary compatibility. Since these save games you created are unofficial, they won't be supported, unless you go into the hassle to do it yourself.

To do so I had to implement the struct GUID, which has to be 16 byte big. _MAX_PATH has to be 260, as on Windows. Activision did something with it that is not good. Maybe it has some unwanted side effects. But we will see.

LynxAbraxas:
The compatibility between OSs is great (though I do miss my late save game...). How about a save game file in the repos (commit on separate branch for a PR for testing first) as a standard for testing loading on linux and win?
Would saving the loaded game lead to a bit-exact result again that could be compared to the standard by a checksum or would there be time-stamps or random seeds in it that would change?

MartinGuehmann:
> How about a save game file in the repos (commit on separate branch for a PR for testing first) as a standard for testing loading on linux and win?

Sure, I already thought about having one in it. I would donate my test game. It is a game from 4000 BCE, which I usually load and run for 200 or 400 turns. Lately, I also did for 800 turns. These AIs like Trumpy Bumpy: Global warming is bull****. I agree raising sea levels and deserted land really sucks. And still these AIs don't get, let's have our spies planting A-bombs everywhere, yeah.

The question is were to put. I use Activision's developer setup, which creates the executable ..\ctp2_code\ctp\ on Windows and that is also the working directory. The save games are stored in a subdirectory. That is by the way the reason for the Run.sh, but I think I don't need it anymore. On Linux you have other possibilities, especially if you want to use the home directory. And actually, storing save games in the exe directory or a subdirectory isn't standard in Windows 10 either, that is Win95/Win98.

Then what would be the name of such a directory, I already have a directory among my savegames called Test, so maybe DevTest.

> Would saving the loaded game lead to a bit-exact result again that could be compared to the standard by a checksum or would there be time-stamps or random seeds in it that would change?

I am not sure, whether I understand the question. Obviously, you get an error that something is wrong with the save game. They have some magic numbers, at least they call it magic numbers.

[...]

PR #63, opened 2019-01-23
WIP: PR to collect commits needed to make master compile and pass tests again

LynxAbraxas:
With the changes of this PR, master compiles again with SDL_ffmpeg and jobs up to game play succeed, i.e.
https://github.com/LynxAbraxas/ctp2DF/blob/master/.gitlab-ci.yml#L9-L13

```
stages:
  - build
  - testSprite
  - spriteBuild
  - spriteInclude
```

from the play tests 
- `load-game` fails with `Messagebox(Load Game Error): 980424-JRL-015 - The size of the save game file is incorrect`
- `play-game_build-city` fails also, reason unclear, needs investigation

### DIFF
--- a/ctp2_code/ui/aui_common/aui_movie.cpp
+++ b/ctp2_code/ui/aui_common/aui_movie.cpp
@@ -43,6 +43,11 @@
 #include "soundmanager.h"		// g_soundManager
 #include "aui_sdlsurface.h"
 
+#ifdef USE_SDL_FFMPEG
+#include <SDL_ffmpeg.h>
+SDL_ffmpegFile* film;
+#endif
+
 Mix_Chunk achunk;
 
 int m_moviechannel;


LynxAbraxas:
@MartinGuehmann is the error `The size of the save game file is incorrect` to be expected with your changes of late? 
Is it possible to convert existing game files (e.g. those in https://github.com/LynxAbraxas/ctp2DF/tree/master/.civctp2/save/games/Julius) to the new format?

LynxAbraxas:
Ogg playing (on linux) is broken for current master @ https://github.com/civctp2/civctp2/commit/0eee854da0ad1f249701269a2efc65f59543f8ed (was working with https://github.com/civctp2/civctp2/commit/74c4f072fa185716eef73fd0393689a5112a27ad).

MartinGuehmann:
Load game is #55. It is supposed to load official games. Which include any game created on Windows including the scenarios. What happened is Activision saved something like this in Game.cpp:

n = c3files_fread(&info->gameSetup, sizeof(nf_GameSetup), 1, saveFile);

You store the whole thing in the file including vtables. Then it includes an object from the anet library, which has some members of the type long. The long is long not some typedef like sint32. Which would be correct on both 32 and 64 bit programs. So it is 32 bit on Windows 32 bit and 64 bit on my Linux compiler. Somewhere in was also a pointer stored 32 bit vs. 64 bit.

Well, there are a few other things causing incompatibilities. In principle you could have it so that the game loads that unofficial game format and saves it in the official format. But note we have at least, three formats out there, the official Windows format and at least a Linux 32 bit and a Linux 64 bit format. There might be other formats out there. That depends how the compiler aligns the members.

play-game_build-city that is PR #58. I am not sure whether the reason for that is just stupidity or lies on a deeper. If it is a deeper problem then it is how CityData handles the copying of BuildQueue. Well, I think the whole copy thing is quite bad. I think CityData should not be copied at all, but I think it is not a trivial fix either. I haven't merged #58 it to give you a change to discuss it, but I guess I should have merged it, immoderately. But here I need some feedback, how quickly you like to have me merged these things.

MartinGuehmann:
> @MartinGuehmann is the error `The size of the save game file is incorrect` to be expected with your changes of late?
> Is it possible to convert existing game files (e.g. those in https://github.com/LynxAbraxas/ctp2DF/tree/master/.civctp2/save/games/Julius) to the new format?

That error message is expected, I got it often enough with my saves. I would still say it is better than just segfaulting, when I am just opening the load screen.

OK, you convinced me, I make a converter version for the save games.

MartinGuehmann:
> Ogg playing (on linux) is broken for current master @ [0eee854](https://github.com/civctp2/civctp2/commit/0eee854da0ad1f249701269a2efc65f59543f8ed) (was working with [74c4f07](https://github.com/civctp2/civctp2/commit/74c4f072fa185716eef73fd0393689a5112a27ad)).

Maybe I have an idea what is, but not today.

MartinGuehmann:
Sorry, for the misunderstanding, I commented in #57 about the WIP.

LynxAbraxas:
> Load game is #55.

Looks like it. 

> OK, you convinced me, I make a converter version for the save games.

That would be very welcome.

> play-game_build-city that is PR #58.

Not sure if the failure of `play-game_build-city` is related to #58. Will have to investigate that, possible will find some time for that the next weekend.

MartinGuehmann:
> > play-game_build-city that is PR #58.
> 
> Not sure if the failure of `play-game_build-city` is related to #58. Will have to investigate that, possible will find some time for that the next weekend.

Well, I tried to build a city, the game just crashed, and #58 is the result of it.

LynxAbraxas:
> Well, I tried to build a city, the game just crashed, and #58 is the result of it.

Thanks, that solved the `play-game_build-city` test problem (I though it was already included before 0eee854d, but it came afterwards).

LynxAbraxas:
Created a new test game for `load-game` test, which now succeeds again.

PR #64, opened 2019-01-24
Make the AI compare the city limit of it current government, when it …

MartinGuehmann:
…is looking for a better government

The AI has a list of governments it will use. If it has too many cities it will look whether it contains a government that allows more cities. If so it will switch to. However, it contained a bug instead comparing the governments from the list to the current government it would compare it to another government in the list. Of course that does not allow the AI to find a better government. But it may also crash the game if the current government's index is higher than the list is long. This is a problem that occurs in the late game with developed AIs.
..\ctp2_code\ai\CityManagement\governor.cpp

Added the database indices of the governments in the comments, that is useful for debugging such as the previous problem.
..\ctp2_data\default\gamedata\govern.txt

If a city bankrupts it does not try to sell building if the maintenance of the remaining buildings is zero, such as the capitol. This actually just fixes an assert in the debug version, and give a little speed, but probably not noticeable.

Removed some outcommented unused code (by the way commit)
..\ctp2_code\ai\strategy\scheduler\Scheduler.h
..\ctp2_code\ai\strategy\scheduler\scheduler.cpp

### DIFF
--- a/ctp2_code/ai/CityManagement/governor.cpp
+++ b/ctp2_code/ai/CityManagement/governor.cpp
@@ -400,7 +400,7 @@ sint32 Governor::ComputeBestGovernment() const
 		if (player_ptr->GetNumCities() > newCityLimit)
 		{
 			sint32 const oldCityLimit =
-				strategy.GetGovernment(player_ptr->GetGovernmentType())->
+				g_theGovernmentDB->Get(player_ptr->GetGovernmentType())->
 					GetTooManyCitiesThreshold();
 
 			if (newCityLimit > oldCityLimit)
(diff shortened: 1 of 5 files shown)


LynxAbraxas:
Testing this, seems to be an important PR. 
@MartinGuehmann I would suggest though to split into 3 commits (database change together with the first), possibly even to make 3 PRs of this and to tell in the commit message for the last change why the code is unused and why it was needed before (if possible).

MartinGuehmann:
Those are now #80, #81, and #82.

LynxAbraxas:
Would suggest to close this PR in favour of the other 3.

MartinGuehmann:
So, the the others can go in?

LynxAbraxas:
Yes, no objections from my side.

Issue #66, opened 2019-01-25
Issues with master under linux after recent merges


LynxAbraxas:
In relation to #63, while playing up to 3450BC (22 turns) on linux from DI, I had about 6 dead-locks of the game (compiled at https://github.com/civctp2/civctp2/commit/9371b6c1094c65c40493cd4f8f86e9c5b08ec589), which is far, far more often than I had crashes (no dead-locks) before (one in about 3 hours).
Also the OGG playing (that used to work with https://github.com/civctp2/civctp2/commit/74c4f072fa185716eef73fd0393689a5112a27ad) is broken, game sounds and video (with audio) are functional though.
@MartinGuehmann any idea what commits could be the source for these changes?

MartinGuehmann:
The dead-locks are resolved in PR #61. Now it is crashing again. I merged them quite quickly, when I realized that you and others shouldn't test these things. I guess these PRs are a bit special.

For the OGG playing, I have no idea right now, for that I would use TortoiseGit, which is not available on my Kubuntu. I got a debug version compiling with some of the debug tools. However for now it trows some Asserts that aren't thrown on Windows and then crashes. However, compiling a final version seems to be fine. Probably, I have now enough for a WIP pull request.

Obviously it is something with sound, I edited the sound manager, but probably there are some other places to look for.

LynxAbraxas:
> The dead-locks are resolved in PR #61. Now it is crashing again. 

Sadly not, the game at https://github.com/civctp2/civctp2/commit/9371b6c1094c65c40493cd4f8f86e9c5b08ec589 already includes #61. I think #61 has not reverted all changes. I'll take a look.
Odd though that the dead-locks appear much, much more often than the former crashes.

MartinGuehmann:
I looked it again, looks like I misunderstood what TortoiseGit was showing me, so that I just undid one of your edits on sdl_blitter.cpp:

```
        // lock both surfaces
        // note that there is a chance of a deadlock if a different
        // thread locks the two surfaces in the opposite order
        /*SDL_LockMutex(sdlDest->m_bltMutex); //http://www.libsdl.org/cgi/docwiki.cgi/SDL_5fBlitSurface says don't lock surface! mutex lock OK but seems not needed
        SDL_LockMutex(sdlSrc->m_bltMutex);*/
```

Looks like the dead locks are a well known problem.

Well, at least I should now have got it, it is PR #67.

MartinGuehmann:
[...]

Issue #30, opened 2018-11-17
irregular segfaults during game play


LynxAbraxas:
Observed irregular segfaults that happen after playing a few up to many rounds (master branch run under linux, sometimes not occurring for hours) no obviously related to some specific game state. Probably needs gdb debugging during playing.

LynxAbraxas:
Tested if SDL surface locking is causing this with the changes from (https://github.com/LynxAbraxas/civctp2/commit/6a96bd28cb425fb09bdcde718b0878454d39f698) and (https://github.com/LynxAbraxas/civctp2/commit/a6998b7b264da780e21c2500238dca45b4185021) the crashes seem to be replaced by dead-locks of the game.

MartinGuehmann:
Hey, I got a segfault while it was running on the debugger, must have something to do with the mouse and the blitter. Concerning that I changed the blitter thing on the Windows version, this does not seem so unsurprising.

#0 ??() at 
#1 SDL_LowerBlit() at /usr/lib/x86_64-linux-gnu/libSDL-1.2.so.0
#2 SDL_UpperBlit() at /usr/lib/x86_64-linux-gnu/libSDL-1.2.so.0
#3 aui_SDLBlitter::Blt16To16(aui_Surface*, tagRECT*, aui_Surface*, tagRECT*, unsigned int)() at 
#4 C3Blitter::Blt16To16(aui_Surface*, tagRECT*, aui_Surface*, tagRECT*, unsigned int)() at 
#5 aui_Blitter::Blt(aui_Surface*, int, int, aui_Surface*, tagRECT*, unsigned int)() at 
#6 aui_Mouse::BltDirtyRectInfoToPrimary()() at 
#7 aui_UI::Draw()() at 
#8 aui_UI::Process()() at 
#9 CivApp::ProcessUI(unsigned int, unsigned int&)() at 
#10 CivApp::Process()() at 
#11 CivMain(int, char**)() at 
#12 main() at 

Unfortunately, I haven't figured out how to use the debugger in kdevelop. It shows me the frame stack but I can not click on the function and it sends me to the place in the source file if it is a cpp file. However, it does it if it is a c file.

The only thing I would do before really debugging is to merge in the changes from the Linux branch. It might be fixed there. Or the problem might be different. But I think that is could also be related to the mouse artifacts.

LynxAbraxas:
[...]

PR #67, opened 2019-01-26
Undo the changes of aui_sdlblitter.cpp, I should not have included, second trial

MartinGuehmann:
Looks like, I didn't catch all the changes in aui_sdlblitter.cpp, when I included the remote branch debug-random-segfault. This should now be everything.
..\ctp2_code\ui\aui_sdl\aui_sdlblitter.cpp

### DIFF
--- a/ctp2_code/ui/aui_sdl/aui_sdlblitter.cpp
+++ b/ctp2_code/ui/aui_sdl/aui_sdlblitter.cpp
@@ -40,8 +40,8 @@ AUI_ERRCODE aui_SDLBlitter::Blt16To16(
         // lock both surfaces
         // note that there is a chance of a deadlock if a different
         // thread locks the two surfaces in the opposite order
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Tested this PR and it seems to revert back to the former behaviour: no dead-locks but crashes, but far less often than the dead-locks.
Since this state seems to be the lesser of to evils, I would suggest merging this until a proper solution is found to #30 (see also https://github.com/civctp2/civctp2/issues/66#issuecomment-457781784).

MartinGuehmann:
> Tested this PR and it seems to revert back to the former behaviour: no dead-locks but crashes, but far less often than the dead-locks.
> Since this state seems to be the lesser of to evils, I would suggest merging this until a proper solution is found to #30 (see also [#66 (comment)](https://github.com/civctp2/civctp2/issues/66#issuecomment-457781784)).

Good, since I don't really expect comments from other developers, I can merge it.

PR #69, opened 2019-01-27
revert "ESC show the option window again" from 6e795dc2c7efb042

LynxAbraxas:
Showing the option window should have another key than ESC.
ESC is very commonly used for closing a window, which is an action that is far more often needed during game play (e.g. showing unread messages with TAB and closing them with ESC) than opening the option window.

### DIFF
--- a/ctp2_code/ui/aui_ctp2/keypress.cpp
+++ b/ctp2_code/ui/aui_ctp2/keypress.cpp
@@ -363,8 +363,8 @@ sint32 ui_HandleKeypress(WPARAM wParam, LPARAM lParam)
 						data->GetButton(0)->Callback();
 					}
 				}
-			} else if (g_chatBox->IsActive()) {
-				g_chatBox->SetActive(false);
+			} else if (g_chatBox) {
+			        g_chatBox->SetActive(false);
 			} else if (g_optionsWindow && g_c3ui->GetWindow(g_optionsWindow->Id())) {
 
 				optionsscreen_removeMyWindow(AUI_BUTTON_ACTION_EXECUTE);


MartinGuehmann:
So what would be the alternative key? I am quite used to it, using it?

LynxAbraxas:
Hm, which would you like of those that are free? How about Ctrl+o?

The way I understand the code of https://github.com/civctp2/civctp2/blob/05eb30eb88b669ddc73441d8f919327e1cb9be5a/ctp2_code/ui/aui_ctp2/keypress.cpp#L343-L382 is that ESC should close (most) open windows and only open the option screen if none is open any more. However it does not close the option screen, so hitting ESC once to often leaves you in the option screen and forces you to use the mouse to close it, which I find non-practical. Any idea how to solve this in another way?

MartinGuehmann:
The thing is if I am put the game on auto turn, then the escape key allows me to escape from auto turn and the option window is a visual cue that I have escaped.

In principle it could also be some sound that is played, then I got my feedback. And that would also be feed back for people that they have escaped enough. I thing the original idea is that the screen also includes quit game. In that sense it is also some escape.

I think I am happy if you give me some sound and put the option screen on Ctrl+o.

LynxAbraxas:
Just checked: The option screen is by default on F10, wich I guess is simpler than Ctrl+o.

I see, your use of ESC is special. I never used ESC for that, because the auto-turn is not (yet) available under linux.
It isn't clear to me why ESC does not close the option screen. Shouldn't
https://github.com/civctp2/civctp2/blob/05eb30eb88b669ddc73441d8f919327e1cb9be5a/ctp2_code/ui/aui_ctp2/keypress.cpp#L370
or
https://github.com/civctp2/civctp2/blob/05eb30eb88b669ddc73441d8f919327e1cb9be5a/ctp2_code/ui/aui_ctp2/keypress.cpp#L371
close it?

MartinGuehmann:
Well, it depends whether you have the chat box defined. You should only close it if it is open and that is the thing I fixed here. The chat box is initialized on game init. This means this is always true:

`else if (g_chatBox)`

So if you want to change the key behavior, first revert the file, that is one line and then change what is in:

```
		} else { 
  
 			optionsscreen_Initialize(); 
 			optionsscreen_displayMyWindow(1); 
 		} 
```

Actually, optionsscreen_Initialize is done in optionsscreen_displayMyWindow if necessary.

LynxAbraxas:
> The chat box is initialized on game init. This means this is always true

That's what I though as well, but it wasn't working that way before.
I tested this again, and it works as expected for the current master, i.e. ESC closes open windows and if none is open anymore it opens the option screen, which can then be closed with ESC again. Possibly, this had something to do with the key handling under Linux that got fixed by some of the merges. 
I'm fine with the current behaviour as long as ESC closes open windows first, so closing this.

Issue #49, opened 2018-12-13
pirating trade seems to have no (direct) effect


LynxAbraxas:
Pirating trade shows the pirate flag animation but no (direct) gold increase.

LynxAbraxas:
@MartinGuehmann What effect would pirating trade routes have on windows? I would expect to get the trade route gold right away, like it happens for e.g. *selling indulgences*. Could this be a linux problem?

MartinGuehmann:
I am not sure, what it exactly does and what it is supposed to do. For sure it is supposed to annoy the player, and of course it does. As far as I can see the AI cancels trade routes that are being pirated for a few turns, I don't know for how many. As a player I haven't cared so much about it. I am a builder so I build all the stuff what you can build in a city and then go to the rest. So trade rather come late in the game and then the AIs aren't a thread anymore at least in Activision's version.

I recently tried out the tutorial. There, I discovered the problem with the CityData and when I fixed I played on for some 170 turns. You know just one more turn and the tutorial told me to build the caravans and found cities here, because right to West we have some tea. And that on the debug version.

I would say most likely it is neither Windows nor Linux, but just a bug in the game. Either it comes from Activision, or we introduced it at some point.

LynxAbraxas:
> As far as I can see the AI cancels trade routes that are being pirated for a few turns, I don't know for how many.

Never observed that in a game. Perhaps the bug is somewhere before that so the AI actually never acts upon pirating trade because it never sees any effect from it.

LynxAbraxas:
@MartinGuehmann Any idea roughly where in the code the problem might be?

LynxAbraxas:
> ordered to pirate and then left there for the following turns. Gold will only be assigned at the beginning of each new turn if the unit stays there.

Oberseved another problem with this: If a unit is ordered to pirate, then the game is saved and reloaded, the pirating event seems not to be taken into account for the next turn. Pirating only works for me between trurns if the game is not reloaded.
@MartinGuehmann Do you see a chance to change the pirating code such that it just works like a cleric selling indulgences? I.e. the gold is taken from the victim and assigned to the attacker at once (leaving it to the player to remember to repeate the attack for each turn).

MartinGuehmann:
@LynxAbraxas, as far as I can see you have to remember to pirate, anyway. Of course you can change the code to whatever you want, the only problem we have right now is that we don't know exactly how the gold thing works, if I understood you correctly in #76.

[...]

PR #71, opened 2019-01-27
message for pirating civ about the gold captured

LynxAbraxas:
This PR introduces a game message to tell the pirating civilization what amount of gold was pirated from with trade route. This is in relation to #49.
Pirating with ctp2 style works in the way that a unit is placed on a trade route, ordered to pirate __and then left there for the following turns__. Gold will only be assigned at the __beginning of each new turn__ if the unit __stays there__.

### DIFF
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -1656,9 +1656,22 @@ void ArmyData::BeginTurn()
             for(i = 0; i < cell->GetNumTradeRoutes(); i++) {
                 TradeRoute route = cell->GetTradeRoute(i);
                 if(route->GetPiratingArmy().m_id == m_id) {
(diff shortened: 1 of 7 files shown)


LynxAbraxas:
The messages from CTP1 that I think your refer to seem to randomly give away who it was in the message to the player suffering the pirating.
This message is only for the player doing the pirating to have some feedback that the gold arrived.

MartinGuehmann:
OK, I see could you still add the message to info_str.txt from the other languages, no translation is needed, except for the German file. The translation is:

`TRADE_PIRATE_GOLD			"Wir erbeuteten {gold[0].value} Gold als wir den Handelsweg von {city[0].name} nach {city[1].name} von {player[0].civ_name_plural#DAT} überfiehlen."`

MartinGuehmann:
The English info_str.txt now contains the message, twice.

LynxAbraxas:
Ah, yes, many thanks for noticing that mistake, it's always good to have a review.
If you find the new commit OK, it would be ready for merging from my side.

Issue #75, opened 2019-01-28
victim of piracy is not getting a message/info about this fact


LynxAbraxas:
In relation to #49 and PR #71, it seems that the victim of piracy is not getting a message about this fact, if playing with the ctp2 style of pirating rule. (This only seems to happen if ctp1 style of pirating rule is used.)
Additionally, as I understand the code, the trade gold of the route that is getting pirated is not removed from the victim (only additionally added to the attacker).

LynxAbraxas:
I have the impression that the pirating code is not at an ideal position possibly wrong there:
https://github.com/civctp2/civctp2/blob/e0edb94b97916c979857a712ab8530779f9055b4/ctp2_code/gs/gameobj/ArmyData.cpp#L1650-L1681
There is also InterceptTrade:
https://github.com/civctp2/civctp2/blob/e0edb94b97916c979857a712ab8530779f9055b4/ctp2_code/gs/gameobj/ArmyData.cpp#L5862
I noticed that pirating only succeeds sometimes, but I cannot find any randomization in the code concerning success.
Additionally, I get a message if I pirate a route from an AI I'm not at war with, but when the AI pirates a route of mine, I do not get a message (#76 should amend that) nor is the pirating listed in the column of the trade window. 

@MartinGuehmann what do you think about the pirating code?
Do you see any randomization concerning its success?
Why is it different to that of e.g. a cleric selling indulgences (https://github.com/civctp2/civctp2/blob/e0edb94b97916c979857a712ab8530779f9055b4/ctp2_code/gs/gameobj/ArmyData.cpp#L4330-L4428)?
Is the code missing the subtraction of the pirated gold from the owner of the route?

MartinGuehmann:
@LynxAbraxas, the code is a bit odd, I would put it in the ArmyBeginTurn event, but rather at the code when the prirating is happining, it could be in UnitData::InterceptTrade() or in the GEV_SetPiratingArmy event. I know these events are a bit tricky, first you have to track down the code and you have to keep in mind that the events are executed asynchronously. For instance GEV_SetPiratingArmy is added after current, but it will only executed after current directly if nothing afterwards has been added after current.

GEV_SetPiratingArmy leads you to tradeevent.cpp, and there it just sets the pirating army. Now you have to track what is done if there is a pirating army.

LynxAbraxas:
Thanks @MartinGuehmann  for taking a look.
> Now you have to track what is done if there is a pirating army.

And that leads me back to:
[...]

PR #76, opened 2019-01-28
Message for the civ whose trade route got pirated

LynxAbraxas:
Regarding #75, this PR implements that the victim gets a message about being pirated. 
It does not include an implementation for removing the gold of that trade route (yet), because I'm not sure if that is taken into account at some place already.

### DIFF
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -1670,13 +1670,21 @@ void ArmyData::BeginTurn()
 		    fromCity = route.GetSource();
 		    toCity = route.GetDestination();
 
-		    SlicObject * so = new SlicObject("044TradePirateGold");
-		    so->AddRecipient(GetOwner());
-		    so->AddGold(pgold) ;
-		    so->AddCity(fromCity);
-		    so->AddCity(toCity);
-		    so->AddCivilisation(fromCity.GetOwner());
-		    g_slicEngine->Execute(so);
+		    SlicObject * so1 = new SlicObject("044TradePirateGold");
+		    so1->AddRecipient(GetOwner());
+		    so1->AddGold(pgold) ;
+		    so1->AddCity(fromCity);
+		    so1->AddCity(toCity);
+		    so1->AddCivilisation(fromCity.GetOwner());
+		    g_slicEngine->Execute(so1);
+
+		    SlicObject * so2 = new SlicObject("045TradePirated");
+		    so2->AddRecipient(fromCity.GetOwner());
+		    so2->AddGold(route->GetValue()) ;
+		    so2->AddCity(fromCity);
+		    so2->AddCity(toCity);
+		    so2->AddCivilisation(GetOwner());
+		    g_slicEngine->Execute(so2);
                 }
             }
             if(piratedByMe < 1) {
--- a/ctp2_data/chinese/gamedata/info_str.txt
+++ b/ctp2_data/chinese/gamedata/info_str.txt
@@ -151,7 +151,7 @@ WAR_OVER 				"{player[0].civ_name_plural}与{player[1].civ_name_plural}战争结束。
 
 
 ## Trade Messages ##
-TRADE_PIRATED 				"{player[0].sir_cap}，我{city[0].name}至{city[1].name}的贸易路线遭到{player[1].civ_name_plural}的抢劫。"
+TRADE_PIRATED 				"，我{city[0].name}至{city[1].name}的贸易路线遭到{player[0].civ_name_plural}的抢劫。 and {gold[0].value} gold is lost"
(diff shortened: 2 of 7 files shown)


LynxAbraxas:
This works as expected if tested as described in https://github.com/civctp2/civctp2/issues/75#issuecomment-480625947, so can go in on its own since the PR is not for fixing the general issues concerning piracy.

MartinGuehmann:
Well, then let's put it in. You still have this from the intro:

> It does not include an implementation for removing the gold of that trade route (yet), because I'm not sure if that is taken into account at some place already.

But I think it is taken into account at some place in the code I checked. So that issue should be addressed.

LynxAbraxas:
@MartinGuehmann many thanks for merging and pointing out the missing part from the intro, had forgotten about that. 
Where did you find it in the code you checked? I was not able to find that so far only the place where it is added to the attacker
https://github.com/civctp2/civctp2/blob/e0edb94b97916c979857a712ab8530779f9055b4/ctp2_code/gs/gameobj/ArmyData.cpp#L1660
and I would expect the subtraction close by.

It is also not clear to me if multiple armies/units are pirating the same trade route all of them get the gold (which would be inappropriate) and if it subtracted multiple times (which would be inappropriate even more), which might be related to #75.

MartinGuehmann:
> Where did you find it in the code you checked? I was not able to find that so far only the place where it is added to the attacker

For the sake, I just assumed but it is in CityData.cpp

```
sint32 CityData::CalculateGoldFromResources()
{
	m_goldFromTradeRoutes = 0;
	m_goldLostToPiracy    = 0;

	for(sint32 i = 0; i < m_tradeSourceList.Num(); i++)
	{
		if(!m_tradeSourceList[i]->IsBeingPirated())
		{
			m_goldFromTradeRoutes += m_tradeSourceList[i]->GetValue();
		}
		else
		{
			m_goldLostToPiracy += m_tradeSourceList[i]->GetValue();
		}
	}

	sint32 wonderTradeBonus = wonderutil_GetMultiplyTradeRoutes(m_builtWonders);

	if(wonderTradeBonus > 0)
	{
		m_goldFromTradeRoutes += sint32(double(m_goldFromTradeRoutes) *
										(double(wonderTradeBonus) * 0.01));
		m_goldLostToPiracy += sint32(double(m_goldLostToPiracy) *
										(double(wonderTradeBonus) * 0.01));
	}

	return m_goldFromTradeRoutes;
}
```

Actually, you don't "subtract" you just "do not add". However, m_goldLostToPiracy is just calculated but not used, actually it should show up in the trade manager under advice. So if you want to add a message it should go to the pirate event and the loss shall be the value of the trade route. Obviously, that is quite obscure.

> [civctp2/ctp2_code/gs/gameobj/ArmyData.cpp](https://github.com/civctp2/civctp2/blob/e0edb94b97916c979857a712ab8530779f9055b4/ctp2_code/gs/gameobj/ArmyData.cpp#L1660)
> 
> Line 1660 in [e0edb94](/civctp2/civctp2/commit/e0edb94b97916c979857a712ab8530779f9055b4)
> 
>  g_player[m_owner]->AddGold(pgold); 
> 
> and I would expect the subtraction close by.

Nope, as you can see it is somewhere else, and not even a subtraction.

> It is also not clear to me if multiple armies/units are pirating the same trade route all of them get the gold (which would be inappropriate) and if it subtracted multiple times (which would be inappropriate even more), which might be related to #75.

That's easy you already cited that code:
https://github.com/civctp2/civctp2/blob/e0edb94b97916c979857a712ab8530779f9055b4/ctp2_code/gs/gameobj/ArmyData.cpp#L1658

When you pirate then the pirating army is set. Let's say player 1 pirates. Then on your next turn you get the gold, and the victim does not earn anything from the trade route. Now after your turn, player 2 pirates the same trade then the pirating army is a different army, not yours but of player 2. After player 2 eventually, it will be your turn again, then you won't get any gold and if you don't pirate the route again then player 2 gets on his turn the gold. But if you pirate the route again then the pirating army is your and so player 2 won't get any gold.

So if two players are pirating the same trade route over turns then nobody gets the bounty. Of course, still the trade route does not give any gold to the owner.

LynxAbraxas:
> Nope, as you can see it is somewhere else, and not even a subtraction.

Many thanks @MartinGuehmann for pointing that out. The fact that it is not a subtraction is the reason that I did not find it (search through all `SubGold`).

> So if two players are pirating the same trade route over turns then nobody gets the bounty. Of course, still the trade route does not give any gold to the owner.

Many thanks for that explanation. I think this is likely the reason for the effect that is bugging me, and which I think should be improved, but I'm not sure how so far.

And the explanation above is probably the reason why the message for the civ whose trade route got pirated is not show, and also why the trade manager does not show the pirating square and also why the diplomatic window does not expose the option to ask for "stop pirating" even if you see that player pirating your routes.

I've a test game ready, just need to write up the interaction to reproduce the problem, will try to get to that tomorrow.

PR #78, opened 2019-01-29
Make c3.h to includes what it is supposed to include

MartinGuehmann:
On Linux c3.h is supposed to include ctp2_config.h, which includes config.h, but only if HAVE_CONFIG_H has been defined. This is however defined in windows.h, which was included afterwards. ctp2_config.h was already included by ctp2_inttypes.h, since both files have multiple inclusion guards, things that should be defined in the translation unit won't be defined.

For instance if you build a debug build, _DEBUG should be defined, but because of this you can't be sure whether it is really defined. This is also true for USE_SDL.

Obviously, this small change can change what is compiled at various places. Actually, c3.h is supposed to be a precompiled header that is supposed to be included in every source. However, this is not true, even so it must be done to make sure that config.h is used in every translation unit.


LynxAbraxas:
Changes make sense, currently testing with ctp2DF. There are other places as well where the include directives are quite odd but probably not so critical as here.

LynxAbraxas:
Passes all tests of ctp2DF.

MartinGuehmann:
So, I guess I can include it. Unless you still wanna test more.

LynxAbraxas:
Also tested with newly converted save-games from Linux. Good to go from my side.

MartinGuehmann:
Then let's put it in.

PR #79, opened 2019-01-31
Enable debug for linux

MartinGuehmann:
The Windows debug version has an extended log, a memory leak reporter, reports the times for the sub steps needed on loading and the debug asserts, which tell where something might be wrong.

Currently the code compiles and runs with --enable-debug configured. The debug asserts work, however they are not as comfortable as on Windows to press the ignore button via the keyboard, I have to press alt+i instead of i alone.

On game load we have a lots of asserts, which are absent on Windows, because the function _fullpath returns more often NULL then it does on Windows, and actually it is not supposed to return NULL in this instance.

Neither the extended log nor the memory leak reporter are working. Also the load times are not shown.


MartinGuehmann:
The logs are now working, the memory leak reporter is running, but does not produce any useful information, yet. The problem with _fullpath on Linux wasn't actually a problem, so that assert is disabled on Linux. And increased the buffer size for the debug assert message to prevent it from crashing.

However, now I have a problem for some reason the last debug assert message box does not close.

MartinGuehmann:
Now, the debug asserts print a nice stack trace to the log files and also the leak reporter, gives information that can be used.

LynxAbraxas:
I once tried to configure (and compile) with ` --enable-debug` https://github.com/LynxAbraxas/ctp2DF/commit/8eab8dc3a039db8202aac5272d5fd01bd2b288b4
which did not succeed under Linux at that state of ctp2, but I will try again to follow up with https://github.com/LynxAbraxas/ctp2DF/commits/enableDebug.

MartinGuehmann:
> I once tried to configure (and compile) with ` --enable-debug` [LynxAbraxas/ctp2DF@8eab8dc](https://github.com/LynxAbraxas/ctp2DF/commit/8eab8dc3a039db8202aac5272d5fd01bd2b288b4)
> which did not succeed under Linux at that state of ctp2

That isn't so trivial either. I had to fix the includes in c3.h, and then there is quite something that had to fixed and also be ported. Otherwise, you don't really need the debug version. At least Travis builds. It just doesn't check for Windows and thus I didn't see it before.

I sped it up more, but I think it is still slower then on Windows. The backtrace function doesn't seem to be that fast. Therefore, the assembler code would be better, but for that I would need a fast and reliable method to figure out whether I am at the bottom of the stack. Segfaulting as the indented way of function sounds just crappy.

MartinGuehmann:
I think this could still be improved and also the splash thingy is still missing, but that shall be something for another pull request. So far it is already useful, that it should go into master.

LynxAbraxas:
OK, ctp2DF shows that without `--enable-debug` all compiles fine and test succeed (https://github.com/LynxAbraxas/ctp2DF/commit/ec3518f33cf6225d6b9fb51e14a6483b5463a14f).

With  `--enable-debug` (https://github.com/LynxAbraxas/ctp2DF/commit/9cc4d34ffd0713c417107e7a446fafbe700f2c97) all compiles fine now (many thanks @MartinGuehmann for all that effort) but `load-game` fails with:
```
Assertion (!HasCached()) Failed in File:ui/aui_common/aui_bitmapfont.cpp, Line:330
Messagebox(Assert): Assertion (!HasCached()) Failed in File:ui/aui_common/aui_bitmapfont.cpp, Line:330
```
(probably related to #87)
and `play-game_build-city` fails with:
```
ui/aui_sdl/aui_sdlblitter.cpp L57: Blit failed: Surfaces must not be locked during blit
Assertion (errcode == AUI_ERRCODE_OK) Failed in File:ui/aui_common/aui_mouse.cpp, Line:764
Messagebox(Assert): Assertion (errcode == AUI_ERRCODE_OK) Failed in File:ui/aui_common/aui_mouse.cpp, Line:764

Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:276
Messagebox(Assert): Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:276
```

MartinGuehmann:
> With `--enable-debug` ([LynxAbraxas/ctp2DF@9cc4d34](https://github.com/LynxAbraxas/ctp2DF/commit/9cc4d34ffd0713c417107e7a446fafbe700f2c97)) all compiles fine now (many thanks @MartinGuehmann for all that effort) but `load-game` fails with:
> 
> ```
> Assertion (!HasCached()) Failed in File:ui/aui_common/aui_bitmapfont.cpp, Line:330
> Messagebox(Assert): Assertion (!HasCached()) Failed in File:ui/aui_common/aui_bitmapfont.cpp, Line:330
> ```
> (probably related to #87)

That is what it fixes.

> and `play-game_build-city` fails with:
> 
> ```
> ui/aui_sdl/aui_sdlblitter.cpp L57: Blit failed: Surfaces must not be locked during blit
> Assertion (errcode == AUI_ERRCODE_OK) Failed in File:ui/aui_common/aui_mouse.cpp, Line:764
> Messagebox(Assert): Assertion (errcode == AUI_ERRCODE_OK) Failed in File:ui/aui_common/aui_mouse.cpp, Line:764

Well, if you retest it it might not fail, this is a very erratic one.

> 
> Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:276
> Messagebox(Assert): Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:276
> ```

I haven't seen that. Could you post the stack trace. It is written to ../ctp/logs/civ3log000.txt. Just note that these problems were already in. The debug version just shows these problems.

LynxAbraxas:
>   I haven't seen that. Could you post the stack trace. It is written to ../ctp/logs/civ3log000.txt.

Enabled exporting all files in `logs/` as artrifacts (https://github.com/LynxAbraxas/ctp2DF/commit/e2c94d04d203c928f00eda22d7ce9e5779d5e560).
@MartinGuehmann the console output of a `play-game_build-city` job is here:
https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/162576907
and all log-files of that job:
https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/162576907/artifacts/browse/logs/
which you should be able to see with the reporter status, let me know if not.

LynxAbraxas:
When ctp2 starts I also get an `sh: 1: nm: not found`. Is  nm called by ctp2 when compiled in debug mode?

LynxAbraxas:
OK together with #87 only the 
```
Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:276
Messagebox(Assert): Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:276
```
remains as a problem showing up in the ctp2DF tests. @MartinGuehmann can you access the job artifacts above? Do you have an idea what could be the problem with this assertion?

MartinGuehmann:
> When ctp2 starts I also get an `sh: 1: nm: not found`. Is nm called by ctp2 when compiled in debug mode?

I used nm to extract the symbols from the executable. The thing came with my Linux distro. Obviously, I cannot assume that is available in every distro. The thing is also a bit hackey. So I would also be happy about a different solution.

From the console output:

> sh: 1: nm: not found
> Home directory not accessible: Permission denied

I find the thing a bit odd that you cannot access the home directory. As a normal user, I expect that my home directory is accessible. Of course that could be still a valid setup.

Anyway, that it did not find nm means that the stack trace is meaning less.

By the way, it is also generating two files in the executable folder: CTP_LEAKS_99999.TXT, CTP_LEAKS_ALT_99999.TXT. If it did not record any leaks then CTP_LEAKS_99999.TXT says "None." and CTP_LEAKS_ALT_99999.TXT is empty. However, the Linux version still reports something. And actually, the logs folder should also contain a CivCTP_debug000.txt, however on the Linux version the files seems to be created, but not saved, properly. At least it disappears, again.

For the assert itself comes from CivPaths::MakeSavePath of course that could refer to your home directory as well and if that is not writable then you have a problem as well.

LynxAbraxas:
Fixed `Home directory not accessible: Permission denied`, still assertion is there, with some printfs in https://github.com/LynxAbraxas/civctp2/commit/64bf524bc4201635d9c197af8dd26183cdcf0ed1
I get:
```
nowin32.cpp L48: ./../../ctp2_data/english/graphics/pictures/uptg20i.tga ./../../ctp2_data/english/graphics/pictures/uptg20i.tga !
nowin32.cpp L48: ./../../ctp2_data/default/graphics/pictures/uptg20i.tga ./../../ctp2_data/default/graphics/pictures/uptg20i.tga !
nowin32.cpp L48: setup/data/max/ctp_data/videos/../../ctp2_data/english/graphics/pictures/uptg20i.tga setup/data/max/ctp_data/videos/../../ctp2_data/english/graphics/pictures/uptg20i.tga !
nAssertion (s != NULL) Failed in File:CivPaths.cpp, Line:277
Messagebox(Assert): Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:277
```
see https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/162858981. Do those pahts make sense? Could not find those files nor any similar ones.

MartinGuehmann:
As far as I could see the ctp2CD project contains under ../ctp2_data/german/graphics/pictures/ the file pic565.zfs, I would assume that the it is looking for in there. However, you set this copy to English, so it looks for the file in ../ctp2_data/english/graphics/pictures/, but it is not there. Either you can get the English version of pic565.zfs or copy the German version to that folder. Or you could edit civpaths.txt to use the German version.

Anyway, I think I should edit CivPaths.cpp, because the Linux version of _fullpath returns NULL if the path does not exist. The Windows just produces the path, whether it exists or not. That means the situation here is valid on Linux but not on Windows.

LynxAbraxas:
>  Either you can get the English version of pic565.zfs or copy the German version to that folder. Or you could edit civpaths.txt to use the German version.

Well spotted! Yes, my CTP2CD is German and I switched to English for the tests to be international in case the testing suite get's included. 
Just trying if symlinks suffice. 
Do you know if the English versions of those files differ from other languages?

MartinGuehmann:
> > Either you can get the English version of pic565.zfs or copy the German version to that folder. Or you could edit civpaths.txt to use the German version.
> 
> Well spotted! Yes, my CTP2CD is German and I switched to English for the tests to be international in case the testing suite get's included.
> Just trying if symlinks suffice.
> Do you know if the English versions of those files differ from other languages?

I only have a German version, too. But I looked into the file with Read ZFS File. And they contain some language specific CTP1 art, in English. So these files are superfluous, and uptg20i.tga is not in there. It is the zfs file in the default folder as uptg20i.rim. It is the lower right corner of a Window. But it is odd why it is reported in your setup and not in mine.

MartinGuehmann:
The assert is addressed in pull request #99.

LynxAbraxas:
Tried with German language settings, but am still getting that error:
https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/163523122

```
nowin32.cpp L57: setup/data/max/ctp_data/videos/../../ctp2_data/german/graphics/pictures/uptg20e.tga setup/data/max/ctp_data/videos/../../ctp2_data/german/graphics/pictures/uptg20e.tga !
nowin32.cpp L57: setup/data/max/ctp_data/videos/../../ctp2_data/default/graphics/pictures/uptg20e.tga setup/data/max/ctp_data/videos/../../ctp2_data/default/graphics/pictures/uptg20e.tga !
nowin32.cpp L57: ./../../ctp2_data/german/graphics/pictures/uptg25f.tga ./../../ctp2_data/german/graphics/pictures/uptg25f.tga !
nowin32.cpp L57: ./../../ctp2_data/default/graphics/pictures/uptg25f.tga ./../../ctp2_data/default/graphics/pictures/uptg25f.tga !
nowin32.cpp L57: setup/data/max/ctp_d
Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:277
Messagebox(Assert): Assertion (s != NULL) Failed in File:CivPaths.cpp, Line:277
```

>But it is odd why it is reported in your setup and not in mine.

I only happens in the `play-game_build-city` test when sikuli enters `b` (for building a city) and then adding a Warrior in the build queue and exits the city-manager. First I suspected that auto-save might be the cause, but considering that it is part of the window decoration might point to showing the city-manager window. 

> Anyway, I think I should edit CivPaths.cpp, because the Linux version of _fullpath returns NULL if the path does not exist. The Windows just produces the path, whether it exists or not.

I guess that should come first, before going on debuggint this.

LynxAbraxas:
Got the bugger. When the build-manager is opened, `save/queues` is checked, which did not exist in ctp2DF until https://github.com/LynxAbraxas/ctp2DF/commit/c4f9065ba8a5dd30492df6445496dd3c311ba361
So it was not a language specific problem, this appearance was actually caused by the debug messages sent to stdout, which is buffered. I created #100 where these stdout are changed to stderr (which is not buffered) to avoid such confusion in future.

MartinGuehmann:
> Got the bugger. When the build-manager is opened, `save/queues` is checked, which did not exist in ctp2DF until [LynxAbraxas/ctp2DF@c4f9065]

But I would expect that the game creates those.

LynxAbraxas:
> But I would expect that the game creates those.

True, perhaps after the check?

MartinGuehmann:
> > But I would expect that the game creates those.
> 
> True, perhaps after the check?

Not after the check, they should have been created before. That's why I won't accept pull request #99. There should also be some details about it.

PR #80, opened 2019-01-31
Make the AI compare the city limit of it current government, …

MartinGuehmann:
when it is looking for a better government

The AI has a list of governments it will use. If it has too many cities it will look whether it contains a government that allows more cities. If so it will switch to. However, it contained a bug instead comparing the governments from the list to the current government it would compare it to another government in the list. Of course that does not allow the AI to find a better government. But it may also crash the game if the current government's index is higher than the list is long. This is a problem that occurs in the late game with developed AIs.
..\ctp2_code\ai\CityManagement\governor.cpp

Added the database indices of the governments in the comments, that is useful for debugging such as the previous problem.
..\ctp2_data\default\gamedata\govern.txt


LynxAbraxas:

PR #81, opened 2019-01-31
Make the AI not trying to sell buildings while bankrupting if it does not help

MartinGuehmann:
If a city bankrupts it does not try to sell building if the maintenance of the remaining buildings is zero, such as the capitol. This actually just fixes an assert in the debug version, and give a little speed, but probably not noticeable.


LynxAbraxas:

PR #82, opened 2019-01-31
Remove unused load/save (serialize) methods from Scheduler

MartinGuehmann:
This is leftover code from CTP1 and was used to save the state of the scheduler. However, the schdeuler is now regenerated from the game state and is therefore not be needed to be saved. Additionally, saving it would require more additions, because the code is already outcommented and incomplete, so that it would not compile anyway. Additionally, to use it again would require to change the savegame format, which should however stay as it is for backward compatibility. To sum it up it this code will never used again and therefore can go.

Additionally, it clutters the files for two files it is not a problem, but so far the code produces save games that can be loaded on Windows versions produced by Visual Studio and Linux versions produced by gcc. Both operating systems are little endian systems. On big endian systems the save game format is not compatible. It can only be made compatible by modifying the serialize methods of a round 160 classes, which would be identified by file search. And here it would help if two files less showed up in the results.
..\ctp2_code\ai\strategy\scheduler\Scheduler.h
..\ctp2_code\ai\strategy\scheduler\scheduler.cpp

### DIFF
--- a/ctp2_code/ai/strategy/scheduler/Scheduler.h
+++ b/ctp2_code/ai/strategy/scheduler/Scheduler.h
@@ -115,11 +115,6 @@ public:
 
 	static void ResizeAll(const PLAYER_INDEX & newMaxPlayerId);
 
-#if 0
-	static void LoadAll(CivArchive & archive);
-	static void SaveAll(CivArchive & archive);
-#endif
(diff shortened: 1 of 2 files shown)


LynxAbraxas:

PR #83, opened 2019-02-02
music only needs either CD or OGGs (with SDL)

LynxAbraxas:
There is a bug in the logic for playing music, which aborts if there is no CD or no OGGs, https://github.com/civctp2/civctp2/blob/a21b96cf46d7eb1b89b2c21519f21e5a5db436dc/ctp2_code/sound/soundmanager.cpp#L903. This used to work because before https://github.com/civctp2/civctp2/commit/9446560cae32b09f31831728300334c1675c3403#diff-0039129f4311ecee53c3184b02d234d5L522 CD status was always true. However, with SDL we can have music with either CD or with OGGs (even if there is no CD), which this PR takes into account.

### DIFF
--- a/ctp2_code/sound/soundmanager.cpp
+++ b/ctp2_code/sound/soundmanager.cpp
@@ -900,7 +900,8 @@ void SoundManager::StartMusic(const sint32 &InTrackNum)
 #if !defined(USE_SDL)
 	if (!g_theProfileDB->IsUseRedbookAudio() || !c3files_HasCD()) return;
 #else
-	if (!g_theProfileDB->IsUseRedbookAudio() || !c3files_HasCD() || !m_useOggTracks) return;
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
> if I have my CD in it uses the OGGs and has no idea how to use the music from the CD but this is another pull request.

The OGGs have precedence. At least in the linux branch, the CD is used for music (but not for videos).

@MartinGuehmann if you are fine with the PR, I'd say it can go in.

MartinGuehmann:
> The OGGs have precedence. At least in the linux branch, the CD is used for music (but not for videos).

That explains a few things. The OGGs are only implemented for Linux in master, they require SDL. As far as I can tell is that the CD is not used.

> @MartinGuehmann if you are fine with the PR, I'd say it can go in.

Sure, I just left it open over night, in case something would come to our minds.

PR #89, opened 2019-02-08
Make the Linux version load Windows save games if messages are stored

MartinGuehmann:
The Linux version still could not load save games created on Windows if the message list contained some messages. Probably, only a few messages caused this problem, so that deleting all the messages before switching to this version may already convert old Linux save games.
..\ctp2_code\gs\slic\SlicArray.cpp

PR #90, opened 2019-02-08
Update the save game converter for pull request #89

MartinGuehmann:
Another incompatibility was found for Windows savegames loaded on 64bit versions of CTP2, which includes the current Linux version but no Windows version, because so far all Windows versions are 32 bit. It is a problem that only occurs if the player has messages stores, and it only occurs with a subset of messages. Therefore, save games could be converted by saving them without any stored messages. But of course for more convenience here is an update for the save game converter. It loads save games in the old 64 bit format and saves them in the format generated by 32 bits.
..\ctp2_code\gs\slic\SlicArray.cpp

### DIFF
--- a/ctp2_code/gs/slic/SlicArray.cpp
+++ b/ctp2_code/gs/slic/SlicArray.cpp
@@ -126,9 +126,15 @@ void SlicArray::Serialize(CivArchive &archive)
 			archive.PutUINT8(static_cast<uint8>(m_structTemplate->GetType()));
 		}
 
-		if(m_type == SS_TYPE_INT) {
-			archive.Store((uint8*)m_array, m_arraySize * sizeof(SlicStackValue::m_int));
-		} else {
+		if(m_type == SS_TYPE_INT)
+		{
+			for(i = 0; i < m_arraySize; i++)
+			{
+				archive.PutSINT32(m_array[i].m_int);
+			}
+		}
+		else
+		{
 			Assert(m_type == SS_TYPE_SYM);
 			for(i = 0; i < m_arraySize; i++) {
 				haveSym = m_array[i].m_sym != NULL;
--- a/ctp2_code/gs/slic/SlicArray.cpp
+++ b/ctp2_code/gs/slic/SlicArray.cpp
@@ -127,7 +127,7 @@ void SlicArray::Serialize(CivArchive &archive)
 		}
 
 		if(m_type == SS_TYPE_INT) {
-			archive.Store((uint8*)m_array, m_arraySize * sizeof(SlicStackValue));
+			archive.Store((uint8*)m_array, m_arraySize * sizeof(SlicStackValue::m_int));
 		} else {
 			Assert(m_type == SS_TYPE_SYM);
 			for(i = 0; i < m_arraySize; i++) {
@@ -171,7 +171,6 @@ void SlicArray::Serialize(CivArchive &archive)
 
 BOOL SlicArray::Lookup(sint32 index, SS_TYPE &type, SlicStackValue &value)
 {
-
 	type = m_type;
 
 	if(index < 0 || index >= m_arraySize) {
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
> by saving them without any stored messages.

How do you clear the stored messages? I noticed that happens for `/reloadslic` but is that the way you refer to?

Does the 64-bit format have any advantage over the old 32-bit format?

MartinGuehmann:
> > by saving them without any stored messages.
> 
> How do you clear the stored messages? I noticed that happens for `/reloadslic` but is that the way you refer to?

That is a possibility, but you can also right click on the message in the list, this way you can get rid of them very, quickly.

> Does the 64-bit format have any advantage over the old 32-bit format?

Nowadays, computers are 64 bit systems. They can handle more memory. A 32 bit program can only use 2³² bits of memory, that are some 2 GB, which is something the Windows version already reached on my system, but that was due to the memory leak in the CityData destructor. When CTP2 runs out of memory it does exit(-1). Of course, for the gamer it looks like a crash. But it does not create a crash.txt.

Also, if you go to the limit, it will be harder for the program to find free memory and thus slows down. That are all effects you can notice on Windows. And the downside of going to 64 bit are obviously those incompatibilities and all the libraries need also be 64 bit including the Miles Sound Library, and I don't know whether we are allowed to distribute it. Right now we use the one from the CD, which is 32 bit.

LynxAbraxas:
I was only referring to the advantage of the 64-bit format of ctp2 save-games (not 64-bit in general), i.e. what is the advantage of this:

> It loads save games in the old 64 bit format and saves them in the format generated by 32 bits.

Does the 32-bit save-game format bind the ctp2 executable to be 32-bit? In that case it would be quite inconvenient to have to have all the needed libraries to be 32-bit as well.

MartinGuehmann:
> I was only referring to the advantage of the 64-bit format of ctp2 save-games (not 64-bit in general), i.e. what is the advantage of this:
> 
> > It loads save games in the old 64 bit format and saves them in the format generated by 32 bits.
 
There is no advantage. If you compile CTP2 as 32 bit program then it stores the games in a certain way. If you compile it as a 64 bit program then it stores the save games in another way. CTP2 was not designed in a way that it ensures portability. Even so the programmers had portability in mind. They are using everywhere these typedefs: sint32, uint32, etc. Well, almost everywhere.

> Does the 32-bit save-game format bind the ctp2 executable to be 32-bit? In that case it would be quite inconvenient to have to have all the needed libraries to be 32-bit as well.

No the save game format does not bind the CTP2 executable to be 32-bit. Switching to 64-bit just produces some incompatibilities like in the save game format. But those just need to be found and fixed. Originally, I imagined that the problem might be much worse, but it can be handled. The problem I have is that the Miles sound library binds the CTP2 executable to be 32-bit, but that is only a Windows problem.

> Or putting it the other way round, if this PR does not restrict ctp2 under linux to be 32-bit, it's OK.

Well, this PR actually allows CTP2 to be 64-bit on Linux and in principle also on Windows without disturbing save game compatibility of official save games.

Well, so far it is just the purpose. I just discovered something in the logic, which makes the game loading but not saving the data properly.

LynxAbraxas:
> Well, this PR actually allows CTP2 to be 64-bit on Linux and in principle also on Windows without disturbing save game compatibility of official save games.

That I would welcome.

> I just discovered something in the logic, which makes the game loading but not saving the data properly.

Not having investigated the save-game code, I can't judge on this well. 
It makes me wonder though why I was able to convert my old save-games and load successfully without these changes. Were I just lucky or have I just not noticed that something was not converted correctly?

MartinGuehmann:
> Not having investigated the save-game code, I can't judge on this well.
> It makes me wonder though why I was able to convert my old save-games and load successfully without these changes. Were I just lucky or have I just not noticed that something was not converted correctly?

As far as I can see it only concerns a few messages, that means if those are not on the message list, it is fine. Concerning compatibility, also GoodMod is not free of incompatibilities. I never noticed it, because the final version does not crash or complain about it. Only the debug version complains. Fromafar noticed about it and tried to fix it, but that does not work under all circumstances. The symptom is that after a reload all the trade routes were ruby trade routes. The AI cancels such trade routes and creates new trade routes. In that case, I consider it a minor problem that doesn't need to be fixed, but of course can be fixed.

LynxAbraxas:
ctp2DF does not complain, so I'd say it can go in.

MartinGuehmann:
Then we'll just merge it.

MartinGuehmann:
Did you test it for this PR, since it is the one for the save game converter and not for master? Well, I just merged it, so then #89 should also go in.

PR #93, opened 2019-02-09
optimized animation for camel sprite

LynxAbraxas:
The animation of the camel sprite had sliding feets and was lacking some transitions between the actions.


MartinGuehmann:
I guess nothing more will come, so that I can just merge.

LynxAbraxas:
Tested the new SPR, works as expected under Linux, so can be merged as is.

Issue #92, opened 2019-02-08
Use SDL 2.0 or higher


MartinGuehmann:
Currently, we use SDL 1.2, which handles non British keyboard layouts not very, well. I am trying to open the chat Window, but nothing happens. And the best is if I set the operating system to use a British keyboard layout then all programs do it, but not CTP2. I hit the ä-key, that is the key that is labled on a British keyboard with '. And still nothing happens. I hit + and I only get +, with shift, too.

It seems that SDL 2.0, can handle localized keyboards much better.

LynxAbraxas:
This might solve #30.

williamd67:
I started working on the migration to SDL 2.0. I started by compiling the source code with SDL2 while isolating the parts that did not compile by introducing defines. This resulted in 4 areas that caused issues:
1. screen drawing (SKIP_SDL2_SCREEN_ISSUES)
2. event handling (SKIP_SDL2_EVENT_ISSUES)
3. cd-rom (SKIP_SDL2_CDROM_ISSUES)
4. debug-memory (SKIP_SDL2_DEBUG_MEMORY_ISSUE)
Currently I fixed the SKIP_SDL2_SCREEN_ISSUES and the initial blitting to the screen works. I cannot pass the initial menu as the events do not work properly so I expect some extra work as soon as the events work.
Another big issue is that SDL2 does not support CD-rom any more which mean that there is no migration possible. What is your opinion about this? Can we drop support for CD-rom under Linux or do we need to find another solution?
Next step is to fix the event issues.

LynxAbraxas:
This is great, thanks for working on this @williamd67.
In my opinion we could drop CD-ROM support, since checking `texture.dat` from CD was dropped to support GOG, which is without CD. Those that have CTP2 on CD just need to copy the data over, but it is not necessary any more to have the CD inserted for starting the game. @MartinGuehmann what do you think?

williamd67:
I propose to drop CD-ROM support for Linux as it is not supported by SDL2. Agreed?

LynxAbraxas:
I agree, @MartinGuehmann what do you think? Do you foresee any problems with that?

MartinGuehmann:
> I agree, @MartinGuehmann what do you think? Do you foresee any problems with that?

Well, not really my favorite, but my laptop has no CD/DVD-ROM drive, either. For my taste it is a bit early to pull CD-support. But I guess it is more hassle then it is worth it.

MartinGuehmann:
[...]

PR #94, opened 2019-02-10
Enhance Easter eggs, including the /rnd command for Linux

MartinGuehmann:
This enhances the Easter eggs, from the chat window.

First, it adds the chat window to the options menu, so that I can access on Linux with a non British keyboard. Even if I switch the keyboard to a British layout CTP2 ignores it.

Then, I improved the feedback of the Easter Eggs, if the command involves a player then the feedback messages is in the color of the player given. For instance for the /attach command it is easier to see that the right player was turned into a robot.

Then the /rnd command can now be used on Linux. Just enter /rnd x and the game is running for x turn on auto turn end. Make sure that you turned the visible into a robot, before. If you want to stop auto end turn just hit the escape key.

And finally, some white space clean up with some minor code clean up. No functional change.


MartinGuehmann:
> I noticed though, that the diplomacy window was not always closed before the rounds went on, possibly related to [#33 (comment)](https://github.com/civctp2/civctp2/issues/33#issuecomment-445467570) in the other way round?

That is right, it is closed on the next turn. Just keep in mind that this is a debug tool, and as a developer you don't care so much about it. In principle, in an AI only game, the diplo-window shoudln't pop up at all.

MartinGuehmann:
Merged current master in and waiting for Travis, then it'll go in.

Possibly, the opening of the diplo-window is related to #33, obviously it is strange that the AI is affected by a mere interface change.

LynxAbraxas:
>  I might be tempted to create a GL-CI job that records a video of some rounds in /rnd mode so we'd get not only screen-shots but also animation and sound tested

Trying that (https://github.com/LynxAbraxas/ctp2DF/tree/vidTest), but have troubles getting ffmpeg to record cpt2 in GL-CI under xvfb. Using ffmpeg and x11grab works locally on normal Xorg with dockerimage from GL. It also works for e.g. xclock started in GL-CI under xvfb and it works for e.g. java based GUIs started from a dockerimage under GL-CI with xvfb. 
However, cpt2 from dockerimage in GL-CI under xvfb seems not to work: if xclock runs first, it is visible in the screen-cast but as soon as ctp2 starts the video is black (https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/166482352/artifacts/file/play-game_build-city.mpg)  and the log shows (https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/166482352):
```
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":99.0"
      after 289 requests (90 known processed) with 0 events remaining.
ctp2: Fatal IO error 0 (Success) on X server :99.0.
```
Possibly xvfb, ffmpeg and SDL-1.2 do not work together. @MartinGuehmann any idea what could be the problem?

MartinGuehmann:
> > I might be tempted to create a GL-CI job that records a video of some rounds in /rnd mode so we'd get not only screen-shots but also animation and sound tested
> 
> Trying that (https://github.com/LynxAbraxas/ctp2DF/tree/vidTest), but have troubles getting ffmpeg to record cpt2 in GL-CI under xvfb. Using ffmpeg and x11grab works locally on normal Xorg with dockerimage from GL. It also works for e.g. xclock started in GL-CI under xvfb and it works for e.g. java based GUIs started from a dockerimage under GL-CI with xvfb.
> However, cpt2 from dockerimage in GL-CI under xvfb seems not to work: if xclock runs first, it is visible in the screen-cast but as soon as ctp2 starts the video is black (https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/166482352/artifacts/file/play-game_build-city.mpg) and the log shows (https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/166482352):
> 
> ```
> XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":99.0"
>       after 289 requests (90 known processed) with 0 events remaining.
> ctp2: Fatal IO error 0 (Success) on X server :99.0.
> ```
> Possibly xvfb, ffmpeg and SDL-1.2 do not work together. @MartinGuehmann any idea what could be the problem?

I haven't really worked with that stuff. From the error message, I would guess it has something to do with the (virtual) machine on GitLab. If it is just the video then you can start it without the intro video and see what happens. For that pass to ctp2 via the command line: nointromovie.

If you think that SDL-1.2 is the problem you could try SDL-2.0. I just don't know what you have to adjust for it maybe you can just install it instead of SDL-1.2. Or you need to adjust the include directory paths for it. If it is just that it would be great, because with SDL-2.0 we could support non-UK keyboards, properly.

LynxAbraxas:
> If it is just the video then you can start it without the intro video and see what happens.

Already doing so: https://github.com/LynxAbraxas/ctp2DF/blob/vidTest/.gitlab-ci.yml#L104

> If it is just that it would be great, because with SDL-2.0 we could support non-UK keyboards, properly.

No, sadly not: https://wiki.libsdl.org/MigrationGuide I already thought of SDL-2 but the migration does not seem to be simple. @MartinGuehmann Do you know of any SDL-1.2 test program that is part of ubuntu or alpine linux and that does not use openGL? All test programs I found so far did use openGL.

MartinGuehmann:
> > If it is just that it would be great, because with SDL-2.0 we could support non-UK keyboards, properly.
> 
> No, sadly not: https://wiki.libsdl.org/MigrationGuide I already thought of SDL-2 but the migration does not seem to be simple.

The problematic stuff seems to be the graphics that needs to be re-implemented. Of course the keyboard too. But that isn't so much and it would improve the function. For the graphics it sounds to be really worth it since it also supports modern graphics cards. Audio is almost unchanged, so most likely it will work without major changes or with minor changes. The CD support is missing, but it could come from a library.

> @MartinGuehmann Do you know of any SDL-1.2 test program that is part of ubuntu or alpine linux and that does not use openGL? All test programs I found so far did use openGL.

Well, I would have to search, too. But does it really need to be a specific distro? The test program I found uses openGL but those are only two functions, which possibly could be changed easily: https://gist.github.com/miguelmartin75/6946310

LynxAbraxas:
#104 for the follow-up, as well as #92.

PR #84, opened 2019-02-03
Gl-CI integration for master

LynxAbraxas:
In regard to #68, this PR integrates the configuration for GL-CI from https://github.com/LynxAbraxas/ctp2DF/ into https://github.com/civctp2/civctp2 to enable automated GUI-smoke tests for new commits (possibly only for master).
It is adjusted to avoid git submodules and use GL-CI artifacts for providing `ctp2CD/` instead.

PR #95, opened 2019-02-11
Allow optimizations for GCC builds

MartinGuehmann:
Right now the project readme explains how to build CTP2 under Linux without optimizations. This way, we have a bigger executable and it runs more slowly than it could. However using optimization flags for GCC makes the game crash, when save games are reloaded, that means you load a savegame when already another game has been loaded. This can actually be the same savegame. The first commit fixes this:

If CTP2 is compiled by GCC and optimization flags are used such as -Os or -Og, GCC removes the NULL assignments from the destructors. Since the memory for the Order objects comes from a pool, these parts of memory can be reused and the destructor for that is called again. That is the design, whether it is good or bad is another question. However, pointers in this design must be NULLed for proper code function. So for the Order destructor optimizations are disabled.
..\ctp2_code\gs\gameobj\Order.h

In a second commit, I would like to update the readme, to include the optimize options in the build instructions. I just don't know whether it should go into CFLAGS or CXXFLAGS or both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ The build itself is pretty classing and straight forward:
 
 ```
 ./autogen.sh
-CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -fuse-ld=gold" ./configure --enable-silent-rules
+CFLAGS="$CFLAGS -w -O3 -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -O3 -fuse-ld=gold" ./configure --enable-silent-rules
 make
 ```
 
@@ -40,9 +40,11 @@ If you want to build a debug version it is:
 
 ```
 ./autogen.sh
-CFLAGS="$CFLAGS -w -fno-omit-frame-pointer -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -fno-omit-frame-pointer -fuse-ld=gold" ./configure --enable-silent-rules --enable-debug
+CFLAGS="$CFLAGS -w -Og -fno-omit-frame-pointer -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -Og -fno-omit-frame-pointer -fuse-ld=gold" ./configure --enable-silent-rules --enable-debug
 make
 ```
+
+The option -Og optimizes bulds for debugging, this speeds up the debug function and should in general not change anything of the behavior of the executable, however it still could do so. In that case remove the -Og option.
 
 You can look at `./configure` for options, but there aren't many.
 
--- a/ctp2_code/gs/gameobj/Order.h
+++ b/ctp2_code/gs/gameobj/Order.h
@@ -64,7 +64,17 @@ public:
 		m_gameEventArgs = NULL;
 	}
 
+#if defined(__GNUC__)
+	// GCC optimizes away the NULLing in the destructor
+	// Since the Order objects comes from a pool it is
+	// effectively deleted, twice. That is the design,
+	// whether that is good or bad is another question.
+	__attribute__((optimize("-O0"))) ~Order();
+#elif defined(WIN32)
 	~Order();
+#else
+	~Order();
+#endif
 
 	void *operator new(size_t size);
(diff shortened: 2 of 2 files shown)


LynxAbraxas:
Will try to test this with ctp2DF, also using -Os or -Og. 
@MartinGuehmann what flags do you recommend in addtion to https://github.com/LynxAbraxas/ctp2DF/blob/a0ce3654ba4161c2444e06f6ea2ca98d61ff10ec/Dockerfile#L65-L72 ?

MartinGuehmann:
> Will try to test this with ctp2DF, also using -Os or -Og.
> @MartinGuehmann what flags do you recommend in addtion to https://github.com/LynxAbraxas/ctp2DF/blob/a0ce3654ba4161c2444e06f6ea2ca98d61ff10ec/Dockerfile#L65-L72 ?

In the readme that came with the debug version, I have right now this:

`CFLAGS="$CFLAGS -w -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -fuse-ld=gold" ./configure --enable-silent-rules`

And for the debug version:

`CFLAGS="$CFLAGS -w -fno-omit-frame-pointer -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -fno-omit-frame-pointer -fuse-ld=gold" ./configure --enable-silent-rules --enable-debug`

-fno-omit-frame-pointer is supposed for speeding up backtrace. I just don't know what the difference of CFLAGS and  CXXFLAGS. So I put into both, but I assume it is just needed for one.

LynxAbraxas:
OK, but what optimization flags did you use/test so far in conjunction with this PR, only -Os/-Og?

MartinGuehmann:
> OK, but what optimization flags did you use/test so far in conjunction with this PR, only -Os/-Og?

Only those, there are also -O1, -O2, and -O3, but I would assume that those also cut NULLing in destructors.

LynxAbraxas:
Tested with cpt2DF with -O3 (https://github.com/LynxAbraxas/ctp2DF/commit/83f4aaf0be01757cc38eca53c5b84f767ef1dd5d) compiles and tests all fine, also when manually loading another game when one is already loaded. The test time of play-game_build-city test was reduced from 2:02 (without -O3 https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/165947192) to 1:54 (with -O3 https://gitlab.com/LynxAbraxas/ctp2DF/-/jobs/165944396) i.e. 6s which is not much but when testing the game manually, it did make the impression that it runs faster.
PR can go in in my opinion. Many thanks @MartinGuehmann for this PR making it possible to use -Ox which does improve the gaming experience.

LynxAbraxas:
@MartinGuehmann Do you want to add anything else to this PR or should it be merged (and WIP removed)?

MartinGuehmann:
> @MartinGuehmann Do you want to add anything else to this PR or should it be merged (and WIP removed)?

As I put it in the intro:

> In a second commit, I would like to update the readme, to include the optimize options in the build instructions. I just don't know whether it should go into CFLAGS or CXXFLAGS or both.

The idea is to give the debug options also in the project readme and I still haven't figured out what the difference between CFLAGS and CXXFLAGS is. I could put the optimize flags into both flag collections, but I would prefer to put into the one that is needed.

LynxAbraxas:
As far as I know CFLAGS is for compiling plain c and CXXFLAGS for c++, since we have both in ctp2 I specified -O3 for both: https://github.com/LynxAbraxas/ctp2DF/commit/83f4aaf0be01757cc38eca53c5b84f767ef1dd5d

MartinGuehmann:
Updated the readme accordingly, and then this can go in after the formal build check. Or shall I still add these options to .travis.yml?

MartinGuehmann:
Actually, Travis just checks whether it can be build with gcc. And building with optimizations would only increase the build time, but whether it builds doesn't change. So let's merge.

LynxAbraxas:
> Actually, Travis just checks whether it can be build with gcc. And building with optimizations would only increase the build time, but whether it builds doesn't change.

Yes, Travis only checks building, but with #84 GL-CI now also checks if the game can be run, so in regard to the motivation for this PR I can offer #106.

PR #96, opened 2019-02-14
Fix disasters

MartinGuehmann:
When the worlds is flooded, terrain improvements are not removed properly, when only swamps are converted to shallow water, especially roads. Also ruins are not removed. The AI does not resettle the now empty space. Additionally, when a tile improvement is removed then now also the neighbor tile improvements are redrawn so that roads do not connect to no-where.

Additionally, there is only indented a certain amount of disasters happening, which is now indeed the behavior.

Finally, I added a /gw command to the Easter eggs, for triggered a disaster so that I can better debug it. Therefore this pull request is based on pull request #94.


LynxAbraxas:
The description sounds great, never got up to that state, will test with /gw.

MartinGuehmann:
Usually, the AIs aren't that advanced at that stage, I just let it run for 800 turns to see that.

MartinGuehmann:
Then this can go in as, well.

PR #97, opened 2019-02-16
Fix memory leaks

MartinGuehmann:
I found some leaks, the one in CityData was introduced by redesigning the CityData class. And the one of RadarMap only showed up after opening the Gaia Controller screen.
..\ctp2_code\ui\aui_ctp2\radarmap.cpp

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -1316,6 +1316,7 @@ CityData::CityData(CivArchive &archive)
 }
 
 CityData::CityData(CityData *copy)
+: m_name(NULL)
 {
 	m_distanceToGood = new sint32[g_theResourceDB->NumRecords()];
 	m_happy = new Happy;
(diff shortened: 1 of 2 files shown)


LynxAbraxas:
Seems reasonable, but did not test yet with ctp2DF. @MartinGuehmann  How did you notice those leaks?

MartinGuehmann:
> Seems reasonable, but did not test yet with ctp2DF. @MartinGuehmann How did you notice those leaks?

The debug version generates on shut-down two files in the working directory, by default the executable directory: CTP_LEAKS_99999.TXT and CTP_LEAKS_ALT_99999.TXT.

I should have seen the one in CityData earlier. The one from the Gaia controller screen you would only see if you have opened the Gaia controller screen, which is something you don't see so often.

MartinGuehmann:
This has been long enough around so that it can also go in.

LynxAbraxas:
@MartinGuehmann, sorry, I'm currently not able to keep up with testing all your cool contributions. With #84 you should now get a basic feedbacks similar to those you got from me, except for the manual testing of the game run from the DI, which you can also do yourself, see https://github.com/civctp2/civctp2/pull/84#issuecomment-466803970.

PR #99, opened 2019-02-19
Fixes an assert in the Linux debug version

MartinGuehmann:
This assert is triggered in a situation that can legally happen on Linux but not on Windows
..\ctp2_code\gs\fileio\CivPaths.cpp

### DIFF
--- a/ctp2_code/gs/fileio/CivPaths.cpp
+++ b/ctp2_code/gs/fileio/CivPaths.cpp
@@ -273,9 +273,9 @@ MBCHAR *CivPaths::MakeSavePath(MBCHAR *fullPath, MBCHAR *s1, MBCHAR *s2, MBCHAR 
 		sprintf(tempPath, "%s%s%s%s%s", s1, FILE_SEP, s2, FILE_SEP, s3);
 
 		s = _fullpath(fullPath, tempPath, _MAX_PATH);
-		Assert(s != NULL);
 
 #ifdef WIN32
+		Assert(s != NULL);
 		r = _stat(s, &tmpstat);
 #else
 		r = stat(CI_FixName(s), &tmpstat);


ctp2_code/gs/fileio/CivPaths.cpp
```diff
@@ -273,9 +273,9 @@ MBCHAR *CivPaths::MakeSavePath(MBCHAR *fullPath, MBCHAR *s1, MBCHAR *s2, MBCHAR
 		sprintf(tempPath, "%s%s%s%s%s", s1, FILE_SEP, s2, FILE_SEP, s3);
 
 		s = _fullpath(fullPath, tempPath, _MAX_PATH);
-		Assert(s != NULL);
 
 #ifdef WIN32
+		Assert(s != NULL);
```
LynxAbraxas: Looks like you moved the assertion from linux to win code. Is that intended? I'd say either in both or in neither.

MartinGuehmann: That's indented. We have a custom implementation of _fullpath on Linux. _fullpath turns a path such as ..\ctp2_code\ctp\..\..\ctp2_data\ into ..\ctp2_data\. As far as I can see the Windows version does not care whether the resulting path exists or not. This is then handled by _stat. On Linux already _fullpath checks whether the path exists and returns NULL if it doesn't.

As far as I understand it: stat has no problem with it. In principle we could also drop stat here.

LynxAbraxas: >  On Linux already _fullpath checks whether the path exists and returns NULL if it doesn't.

Not sure about the existance but it retruns NULL if something went wrong. The question is, do we want an assertion in case a path (like `saves/queues`) does not exist?

> stat has no problem with it. In principle we could also drop stat here.

Not absolutely sure about that. The man page for https://linux.die.net/man/3/canonicalize_file_name points to https://linux.die.net/man/3/realpath in contrast to that my local man page says `all but the last component must exist` so probably better leave it in.

MartinGuehmann: Well, then we just think about the assert. I looked more into it. The game is supposed to create the save folders if they do not exist, when it creates the CivPath object. That means, we should get an assert at that place in this situation. Possibly, this is because the place where these directories where supposed to be created were not writable. I assume that it was related to the fact that the home directory was not writable.

Originally, I thought it was related to the image file in that case it was reasonable not to have an assert here. Since, the file would have been found at another place. But this way, this pull request is void and I will close it.

PR #100, opened 2019-02-19
output error/debug messages to stderr, which is unbuffered (in contrast to stdout)

LynxAbraxas:
The (false) assumption in #79 that language specifics are the problem that the `play-game_build-city` test failed was due to debug messages were sent to stdout, which is buffered. This PR changes these stdout to stderr (which is not buffered) to avoid such confusion in future.


LynxAbraxas:
Builds and tests without failures in ctp2DF.

MartinGuehmann:
Thanks, @LynxAbraxas. This is a good one. I also encountered it, but had some other priorities than fixing it. The files look good, you have tested it, and I agree that it is very needed so it can on.

Issue #91, opened 2019-02-08
odd AI behaviour when encountering other player's stealth units


LynxAbraxas:
The AI always seems to attack other player's stealth units when encountered within its own borders without taking diplomatic actions and state into account. I often experienced that with e.g. clerics, the AI attacks (breaks standing peace contracts) and just in the following round the AI wanted to agree on peace again.

MartinGuehmann:
I assume this is addressed by commit 412aa81 in pull request #101. Looks like this is not only about stealth units but about any units.

LynxAbraxas:
Seems to be resolved with #101, this issue should be closed when #101 is merged.

LynxAbraxas:
@MartinGuehmann, possibly only resolved for clerics (see #224)?

Issue #54, opened 2019-01-12
AI able to employ special attact directly from (unit on) a ship but normal player not


LynxAbraxas:
After playing for a while I notice that ships (or loaded units) from the AI seem to be able to perform some special attacks directly from the water (i.e. without unloading any units like cleric, slaver etc). The only thing I could see are some special symbols during the execution of the AI moves (like $ sign on one of my towns) when an AI ship is just beside.
Is that correct? If so, how to do that as a normal player? Possibly, there is some porting to linux missing here.

MartinGuehmann:
Is it really a dollar sign or the cross out sign (red circle with a tilted bar). If it is the cross out sign, then the AI is trying to execute the order from board, but cannot do it. Anyway, it shouldn't be able to do it, neither should it try.

LynxAbraxas:
It is definitely not the "action not possible" sign (red circle with a tilted bar), it seems to be some action was successful sign (with a cash sound), but I am neither able to notice any difference nor do I get a message.

MartinGuehmann:
What I get from speceffectid.txt is that it could be this: 
SPECEFFECT_CORPORATE	 		7	#swirling money franchise

But of course, I would need to see the thing, again the sprite editor.

LynxAbraxas:
I will try to catch it and post more details, tough it only appeared in my game test at an early stage (some time before 1000BC).

PR #101, opened 2019-02-19
Fix the AI

MartinGuehmann:
This pull request collects a couple of AI fixes that I noticed during testing. These things usually change small portions of code. And only related by affecting the AI, and because they showed up consecutively in the same game. For instance in the game where I looked at the Gaia controller victory, I noticed that the AI settled at placed where it did not want to settle. Once I fixed that problem and I looked at the following turns the AI declared war, just to sign a cease fire on the next turn.

Right now this pull request includes:

The AI does not use resources on science if there is nothing to research anymore
The AI fixes all dead tiles within the city borders
The AI builds the Gaia Controller
The AI does not settle at places where it does not want to settle, just because it cleared orders
Event logging has been sped up, which was useful for fixing the issue above, and might also be useful for declare war and cease fire on the next turn

I expect that there are more issues to collect, which can be bundled and then at a certain number be included and then we get to the next collection.


LynxAbraxas:
Very much looking forward to this, will try to test this at the WE.

LynxAbraxas:
I'm not sure I correctly understand the commit message of https://github.com/civctp2/civctp2/pull/101/commits/aa11d30d93c05d7482772eb59d5ed69be12f17ae, why would you want that lawyer sue units at home?

MartinGuehmann:
> I'm not sure I correctly understand the commit message of [aa11d30](https://github.com/civctp2/civctp2/commit/aa11d30d93c05d7482772eb59d5ed69be12f17ae), why would you want that lawyer sue units at home?

That would be corporate branches or other lawyers that are in your territory. Obviously, those make more trouble, since they are more likely, to attack your cities. Actually, this is only a moderate increase, but it should bias the lawyers, care more about enemy lawyers.

LynxAbraxas:
Ah, I see, so basically you mean: Make lawyers prefer to sue enemy units in home territory than abroad.

LynxAbraxas:
@MartinGuehmann Did some play-testing with a DI based on https://github.com/civctp2/civctp2/pull/101/commits/c85b37b1dafa48205ec4ff502fb439fb7b2a4a1e, worked all well so far. I can confirm that this seems to resolve #91.

LynxAbraxas:
@MartinGuehmann How is, what the commit message of https://github.com/civctp2/civctp2/pull/101/commits/a766e96c6477b7505e953b5226bc555631a638a9 describes, accomplishable any way? Does it need special units that can attack directly from a ship/tranporter? Would that possibly be related to #54?

MartinGuehmann:
Even so I commented at the line at the diff, GitHub doesn't give the line here. So it is goals.cpp@1551 that was removed. Basically, a little thing that screws it up.

MartinGuehmann:
> @MartinGuehmann How is, what the commit message of [a766e96](https://github.com/civctp2/civctp2/commit/a766e96c6477b7505e953b5226bc555631a638a9) describes, accomplishable any way? Does it need special units that can attack directly from a ship/tranporter? Would that possibly be related to #54?

I think, I misunderstand this. To conquer a city from a ship you need units who can beach assault, or to put it more into the terms of the unit database, the unit record needs the flag CanBeachAssault, which aircrafts and rockets have, but those cannot conquer cities because they have the CantCaptrueCity flag. Marines are the only land units that can conquer a city from a ship, and that is how it is designed.

That's what the Great Library says about:

```
[UNIT_MARINE_GAMEPLAY]
What sets the Marine apart from the <L:DATABASE_UNITS,UNIT_MACHINE_GUNNER>Machine Gunner<e> is its ability to move and attack immediately after unloading from a Naval <L:DATABASE_ORDERS,ORDER_BOARD_TRANSPORT>Transport<e> vessel.  Whereas other infantry units must first land before beginning their assault, the Marine hits the ground running, striking targets with alacrity.
[END]
```

And here is the German version:

```
[UNIT_MARINE_GAMEPLAY]
Der Marine unterscheidet sich vom <L:DATABASE_UNITS,UNIT_MACHINE_GUNNER>MG-Schützen<e> in seiner Fähigkeit, <L:DATABASE_CONCEPTS,CONCEPT_CITY>Städte<e> direkt von einem <L:DATABASE_ORDERS,ORDER_BOARD_TRANSPORT>Transportfahrzeug<e> aus angreifen zu können. Während andere Infanterieeinheiten zunächst an Land gehen müssen, bevor sie einen Angriff durchführen können, stürmt der Marine das feindliche Gebiet und schlägt sofort bereitwillig gegen die Zielstädte zu.
[END]
```

The German version mentions that marines can attack directly from a transporter. The English version isn't so clear about it. Anyway, I wouldn't consider it as the best German what you have in the Great Library and also the English could be improved. And here I mean the style for instance you put the action into the verb. E.g. for "ability" you say "can", as well as in German: Instead of "Fähigkeit" you say "können". That is a style of language meant to impress people of the fancy words you know and the fancy ways how you can say things. But all what it does is to blow up the language with air. Also a transport vessel is just a transporter.

MartinGuehmann:
> Would that possibly be related to #54?

I don't think so.

LynxAbraxas:
Thanks @MartinGuehmann for the follow up. I see, will check while play testing.
Would #112 fall into the scope of this PR? Unless the reason for #112 is a bug, I would guess that game logic for the AI has not been implemented for ctp2.

MartinGuehmann:
> Thanks @MartinGuehmann for the follow up. I see, will check while play testing.
> Would #112 fall into the scope of this PR? Unless the reason for #112 is a bug, I would guess that game logic for the AI has not been implemented for ctp2.

The problem is that the AI does not build diplomats, no diplomats no embassies, and no receptions.  In principle #112 is covered by this pull request.

However concerning what I wrote at the start:

> I expect that there are more issues to collect, which can be bundled and then at a certain number be included and then we g[o] to the next collection.

So, I am thinking to cut here. (Maybe one or two more.) We are now at 37 commits. As far as I can see that nothing is in that stops a merge such as the problem from e9876bd.

I am still not that happy with AI conquering, but when I looked more closely, the AI just hadn't the resources according to the specifications. And it now takes sometimes enough troops and sometimes not enough, which I think is a good balance.

It could still balance the sliders better. And it could build other units, I saw it building ones a cleric, but I only saw lawyers and spies, but no corporate branches, I saw some aircraft, but that isn't a regular thing, no airfields or listening posts, either, it builds aircraft carries, but no aircraft to carry. I haven't seen rockets, even so spies are planting nukes. But no nano infection, plague (maybe sometimes), no soothsaying. Most of these special attacks are missing. At the beginning, the AI is now nicely exploring but it still could focus more on the goody huts, settling next to a goody hut that then gives you a city with incomplete first ring is a no-go. Then the AI could use less memory, but that is another thing.

I saw a cross ocean invasion.

LynxAbraxas:
Cool, the AI was really improved a lot, such that it is now fun to play against the AI alone. It can really come up with something challenging. 
Many thanks for that!

I do agree that multiple small PRs are generally better than just one big one. For one, multiple PRs are easier to test concerning the specific change they propose.

MartinGuehmann:
> Cool, the AI was really improved a lot, such that it is now fun to play against the AI alone. It can really come up with something challenging.

That is the point of a good AI. It should challenge you. But it should keep the balance. Actually, this also depends on the terrain type, it is more challenging if you start in a forest then on grassland. And that is what makes AI players more successful.

> I do agree that multiple small PRs are generally better than just one big one. For one, multiple PRs are easier to test concerning the specific change they propose.

Of course, I am also experimenting here. The thing with the single commits is that they fix a specific problem but that may not be too obvious, what it it does if tested alone. Some of the commits fix AI transporter handling, all of these points keep the AI from getting its units over the ocean. And the result that the AI does not attack you.

Then I went for AI exploration with 967695e, it turned out that it used other units for city defense then it assigned for garrison before, so even so it had the units for exploration it didn't use them. And this is also a problem if the AI has to assign units for conquest, in principle.

In principle, you could have put 587673d into its own pull request, since it is not directly related to conquest, which I think is the focus of this PR. But then less dead tiles mean the AI has more production can produce more units for conquest. So there, we are back.

And then there is the nature of testing these things themselves, to fix a thing I must see it in the game and the best way is to stop and fix it immediately, so that it doesn't get lost.  So that how ae13d31 happened, also something I would only see in the late game with quite some destruction.

And of course this way there is more stuff to find, if you fix one bug then the next bug gets obvious.

I guess now it is time to start FixTheAI2.

LynxAbraxas:
Play-tested https://github.com/civctp2/civctp2/pull/101/commits/efe5a732813021592c5a9009e9b5cb986f3e2411 for some hours and observed no unwanted side-effects, so I'd say this PR is good to go in.

MartinGuehmann:
> Play-tested [efe5a73](https://github.com/civctp2/civctp2/commit/efe5a732813021592c5a9009e9b5cb986f3e2411) for some hours and observed no unwanted side-effects, so I'd say this PR is good to go in.

Well, then let's merge.

PR #102, opened 2019-02-22
Improve the German translation

MartinGuehmann:
The translators of CTP2 had a love for bureaucratise. A language that says things more complicated than necessary, and they mistranslated the order injoin, it is better translated as "Abmahnen" than "Handelsembargo verhängen". Handelsembargo translates into English as trade embargo, which is not meant here.

Then "sue" can be just translated as "verklagen" or just "klagen" then "Klage erheben". The latter is the bureaucratic way to say it. This does not really help to make it comprehensible.

Additionally, I fixed a few more of such problems in the Great Library, where things are said with more words than necessary.
..\ctp2_data\german\gamedata\Great_Library.txt
..\ctp2_data\german\gamedata\gl_str.txt
..\ctp2_data\german\gamedata\info_str.txt
..\ctp2_data\german\gamedata\ldl_str.txt


MartinGuehmann:
This I think is uncontroversial and can then go in.

PR #103, opened 2019-02-22
Make Lawyers and Corporate Branches expellable

MartinGuehmann:
Lawyers and Corporate Branches are special units and civilians and should be expellable as every other civilian and special unit. I just do not know whether it is intended that those cannot be expelled. For sure these units can be sued.
..\ctp2_data\default\gamedata\Units.txt

### DIFF
--- a/ctp2_data/default/gamedata/Units.txt
+++ b/ctp2_data/default/gamedata/Units.txt
@@ -970,6 +970,7 @@ UNIT_CORPORATE_BRANCH {
    IgnoreZOC
    NoZoc
    CanBeSued
+   CanBeExpelled
    CantCaptureCity
    DeathEffectsHappy
    Advertise
@@ -2342,6 +2343,7 @@ UNIT_LAWYER {
    NoZoc
    CanSue
    CanBeSued
+   CanBeExpelled
    CantCaptureCity
    DeathEffectsHappy
    IsSpecialForces


MartinGuehmann:
Set to work in progress, since I am not sure whether this is the way to go or make the AI actually, sue those lawyers and corporate branches.

LynxAbraxas:
Finally, I've hit this situation and am not sure if it is intended that those two cannot be expelled but it have the impression from my game testing, that the lack of being expellable weakens the AI, since then it cannot at all defend itself against Corporate Branches until it reaches the necessary advance to built Lawers. 
The same holds for freeing a city from franchise. While conversions can be removed by any military unit the AI has no way to free one of its cites from a franchise until it can built lawers. Possibly, military units should also be able to do so but in order to keep a meaning for lawers those might be given a better chance or that the removal costs less than when done with a military unit. Or removal by a military unit could cause more unhappiness.

I'm not sure what would be the best way to go but the way it currently works is not good for keeping the AI a challenging opponent. @MartinGuehmann any ideas on how else this could be adjusted?

MartinGuehmann:
@LynxAbraxas, I am not really sure what to do here, either. Therefore, I put the WIP label on it.

The thing is that Lawyers and Corporate Branches cannot expelled by design. That is supposed to be like that. Of course, the question is whether it is a good idea to handle it like this.

One way to help the AI would to make it build Lawyers, so that they can sue Corporate Branches other Lawyers. The alternative would be to allow Lawyers to sue all special units, this way they would become more powerful, and thus we could make them expellable.

Of course, we still would need an alternative way to get rid of franchises. Indeed, suing them by everybody would be a way to go, but that seems to be a bit too easy. But I think that conversions are also too easy to remove. However, I haven't really seen them in action since the AI doesn't use them in the original game, since it is not long enough in Theocracy.

LynxAbraxas:
> One way to help the AI would to make it build Lawyers

Yes, but still the AI would be powerless if it does not own the advance to built them.
Another idea: How about restricting franchising attacks to only those cities whose civ has the advance to built Lawers or Corporate Branches? That way it would be quite fair the way the logic is and it would also be more realistic since a franchise in a civ still in the Middle Ages could not be established, franchising needs some agreement on international law.

LynxAbraxas:
> How about restricting franchising attacks to only those cities whose civ has the advance to built Lawers or Corporate Branches?

@MartinGuehmann what do you think of this idea? How difficult do you think this would be to realize?

In my current game, I've established a franchise in nearly every foreign city unhindered, since the AI cannot expell Corporate Branches. I can even block rail-roads of other civs with a single Corporate Branche. This is very unrealistic and bad for the AI to stay a challenging and interesting opponent.
I'm now considering to use the cheat-mode to give every foreign civ  some lawer to get rid of the franchises. So far I did not continue the main game when I used the cheat mode to test something. @MartinGuehmann what does the warning that the score will not be saved mean in that case? Can I otherwise use the chat window to give lawers to each civ?

LynxAbraxas:
Tested this by becoming an AI player. The expelling works with the PR, however I have never observed that the AI expelled any of my Corporate Branches. @MartinGuehmann would that need extra treatmeant to "teach" this the AI?

PS: How do you rid a city of an established franchise? From the docs in the GL I would expect that I need to use the sue action of a Lawyer and click on a city with a foreign franchise, but that does not work for me.

MartinGuehmann:
> Tested this by becoming an AI player. The expelling works with the PR, however I have never observed that the AI expelled any of my Corporate Branches. @MartinGuehmann would that need extra treatmeant to "teach" this the AI?

Well, the AI has to see the Corporate Branches to expel them. They are special forces and invisible for normal units, so as long as the AI has no special unit next to them it won't expel them. When I tested this pull request the AI expelled Lawyers. They are also quite lethal for small civs if it cannot get rid of them. At that point the AI did not build any Corporate Branches. I think right now we are at the Corporate Branch side and the AI does not really build Lawyers, which we probably have to adjust.

> PS: How do you rid a city of an established franchise? From the docs in the GL I would expect that I need to use the sue action of a Lawyer and click on a city with a foreign franchise, but that does not work for me.

Well, I think this is the way it is supposed to.

LynxAbraxas:
I finnaly observed that the AI expelled a Corporate Branche, so this works.

> Well, I think this is the way it is supposed to.

But does that work for you? I only get an "invald action" sign and that's it.

MartinGuehmann:
> I finnaly observed that the AI expelled a Corporate Branche, so this works.
> 
> > Well, I think this is the way it is supposed to.
> 
> But does that work for you? I only get an "invald action" sign and that's it.

I haven't tried this, but sometimes I got the invalid action sign when I was supposed to execute the action. That was testing with the cheat editor so I put another unit onto the map and that could then execute the special action.

LynxAbraxas:
> I haven't tried this, but sometimes I got the invalid action sign when I was supposed to execute the action. That was testing with the cheat editor so I put another unit onto the map and that could then execute the special action.

Hm, I'm not able to rid any franchise. @MartinGuehmann  Do you mind testing again with the current state of the game?

If I order an Arabian Lawer to rid an Arabian city (Ad Dawhah) from the Persian franchise
ss_2019-07-07_22:11:06

it only leads to the invalid-action sign and nothing else happens
ss_2019-07-07_22:11:32

Came across this problem when investigating solutions for #158 in the scope of #166.

LynxAbraxas:
Under consideration that Franchises seem to end after some turns by default, which works with https://github.com/civctp2/civctp2/pull/167/commits/0f12a2ac2d7c0647445639627c11c3015c3debd3 of #167,  it might be interesting to keep the original behaviour that Lawyers and Corporate Branches cannot be expelled in case the civ that wants to expell already has the advance to built Lawers but to allow civs that cannot yet built Lawers to be able to expell. Or to restrict Franchising to opponents that can built Lawers.

LynxAbraxas:
> Under consideration that Franchises seem to end after some turns by default, which works with 0f12a2a of #167

Now in #168.

MartinGuehmann:
As in line with the final comments of #177, this can now go in.

LynxAbraxas:
From https://github.com/civctp2/civctp2/pull/177#issuecomment-515704330
> > I think then #103 should go in as well. Handling that more specifically, i.e. depending on wether a civ can built Lawyer or not will not be so simple.
> 
> In principle, it devalues suing. But eventually we have to decide and not all ideas of Activision are necessarily the best. So then it goes in.

Well, suing destroys, expelling just sents the unit home. I'll try to think of a way to realize that expelling is only possible for civs that cannot yet built Lawyers, that way we would have both concepts without weakening slow AI civs badly.

PR #106, opened 2019-02-25
using `-O3` for compiling ctp2 in GL-CI

LynxAbraxas:
In regard to https://github.com/civctp2/civctp2/pull/95#issuecomment-467116205, this PR offers compilation with `-O3` for GL-CI. While the compilation takes a bit longer, the test seem to run a bit quicker, however the overall time consumtion is quite similar.
This PR is more interesting for testing if the game actually still functions when compiled with `-O3`, i.e. to catch cases like #95.


MartinGuehmann:
Definitely something we should have.

PR #108, opened 2019-02-25
Sprite building with GL-CI

LynxAbraxas:
This PR contains adjustments and sikuli instructions to automate the building of new sprites based on blender files. If accepted this should not be merged into master but a separate branch.
Using camel.blend from https://github.com/LynxAbraxas/sprites as an example to avoid git submod dependency (resulting SPR already included since #93).

PR #109, opened 2019-02-25
sprite building not part of GL-CI config for civctp2

LynxAbraxas:
Minor change that was missed in #84, sprite building offered in #108.


MartinGuehmann:
Then let's have also the sprite building for testing.

Issue #88, opened 2019-02-06
SDL_ffmpeg for videos


MartinGuehmann:
Currently, we are using SDL_ffmpeg for videos. However, right now it an old version and it is not developed actively anymore.  We have in pull request #62 some discussion about.

Probably, we need to update the code of SDL_ffmpeg and its use in CTP2 anyway. An alternative could be SDL since it has a header about videos.

LynxAbraxas:
>   Yes, I had updated SDL_ffmpeg for ctp2!

@ptitSeb Did you finish? Do you have any instructions on using/compiling your version? What version of ffmpeg does it build on?

I would then try to use it with the instruction to continue on https://github.com/LynxAbraxas/ctp2DF/commits/videos4master for automated testing.

ptitSeb:
Yes, I finished. But it still use some "deprecated" function, so maybe it isn't working anymore with current ffmpeg version.
It's been more than a year ago, so I may have forgoten things.
There is nothing spectacular for the build, it's a regular cmake project, so create a `build` folder, use `cmake` (or `ccmake` to check it founds all `avXXXX` libs) then make and that's it.

LynxAbraxas:
Thanks @ptitSeb. So was it working with the ffmpeg version that was current when you finished? It took me a lot of time finding out the compatible version of ffmpeg for the old SDL_ffmpeg that ctp2 was originally using.
Did you check if your updated version of SDL_ffmpeg was compatible with the way it is called/used by ctp2? I had the impression, that ctp2 would need adjustments even for the last release of @arjanhouben.

ptitSeb:
Yes, SDL_ffmpeg was compatible with current version (last year), but I haven't tried with current current version.
Also, I hadn't started really pluggin that version in CTP2, so I cannot say if it needs adjustments or not.

LynxAbraxas:
It is even difficult to find out what was the current version of ffmpeg at the time of your last commit (https://github.com/ptitSeb/SDL_ffmpeg/commit/32767d4e223090396995aae84730331bf1d39cfb). @ptitSeb would you have a look at:  https://ffmpeg.org/olddownload.html https://ffmpeg.org/download.html and https://github.com/FFmpeg/FFmpeg/releases?after=n3.0.11 and check what version of ffmpeg you used?

MartinGuehmann:
I guess this can now be closed.

LynxAbraxas:
Yes, deprecated with #298.

PR #110, opened 2019-02-25
WIP: updating to SDL_ffmpeg that works with modern ffmpeg

LynxAbraxas:
Regarding #88 this PR concerns updating to SDL_ffmpeg that works with modern ffmpeg. So far it contains the changes to GL-CI for testing this, however building https://github.com/civctp2/civctp2/commit/4d2aee2ad647af5f316d862b1bbdd35333b8a381 fails because [SDL_ffmpeg@FFmpeg-2018](https://github.com/LynxAbraxas/SDL_ffmpeg/tree/FFmpeg-2018) demands corrections to its use in the code of ctp2 in 
https://github.com/civctp2/civctp2/blob/132f78b67c1997c64e034c888d02f62457ca2ce5/ctp2_code/ui/aui_common/aui_movie.cpp#L233
and elsewhere.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,7 +71,7 @@ COPY ctp2_data/  /ctp2/ctp2_data/
 
 RUN cd /ctp2 \
     && ./autogen.sh \
-    && CPPFLAGS="-I/usr/local/include/SDL/" \
+    && CPPFLAGS="-I/usr/include/SDL/ -I/usr/local/include/SDL/" \
     CC=/usr/bin/gcc-5 \
     CXX=/usr/bin/g++-5 \
     CFLAGS="$CFLAGS -w -fuse-ld=gold" \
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Only the docker built uses SDL_ffmpeg, so that is the only one testing it and the only one that fails: https://travis-ci.com/civctp2/civctp2/jobs/180436538#L5535
GL also tests SDL_ffmpeg but that is not triggered here (would need a manual push of the PR commits to GL, see https://github.com/civctp2/civctp2/pull/84#issuecomment-467424925) and should work when the travis docker build works so no need for that until the PR passes the build tests.

LynxAbraxas:
@ptitSeb, @MartinGuehmann  Please go ahead to adjust the ctp2 code to make use of the new SDL_ffmpeg if you find the time, I won't be able to do so in the near future.

LynxAbraxas:
Not needed any more with the great work in #298 of @williamd67.

PR #113, opened 2019-03-19
Fix the AI2

MartinGuehmann:
The second batch of commits for fixing the AI. Right now it mainly deals with AI grouping and rallying, but also with conquering one tile islands and one tile peninsulas if the access is blocked by a foreign unit. Additionally, it contains also AI debug message improvments.


LynxAbraxas:
Tested this PR for some time now, no crashes or unexpeced behaviour so far, so I'd say it's good to go in.

MartinGuehmann:
Then, let's put it in. This means I have to sort out a few commits in my local copy, but least it makes also sense doing it by topic.

PR #114, opened 2019-03-28
Televangelist "faith heal" message expects spec. of civ

LynxAbraxas:
Televangelist message is sightly different to that of clerics. Either the messages need to be adjusted or the civ specified for the message to avoid in-game errors/warinings. 
This PR goes for the latter.


MartinGuehmann:
This can go in as well.

PR #115, opened 2019-04-05
Allow unloading 12 units from stacked transporters at once

MartinGuehmann:
For instance the coracle can only transport one unit, if you want to bring 12 land units to a target tile you would use 12 coracles and group. However, in that scenario you could not even unload one unit, because first the units are unloaded at the position of the coracles and added to the unit list of that tile, which can only hold 12 units. Which means no additional can be added.

From a player's perspective, I consider it as a bug, but it is the design. A quite stupid design and gives the AI trouble to unload units. For instance if the transporter is in a city with 11 other units on the target continent.

Basically, this changes it so that the units are not inserted into the unit list of the transporter position, but only their position is set to the transporter position and then they are moved to their final destination. However, nothing has changed for helicopters if they unload their cargo at their position.


MartinGuehmann:
That should have been enough time. I haven't seen any problems with it in further testing. So it is time for merging.

PR #118, opened 2019-04-15
Improve pathing

MartinGuehmann:
This pull requests addresses some path finding problem. First, it cleans up some of the A-star files, then it gives pretty paths to everybody and improves the heuristic for future costs, which is the main goal of this pull request. It also adds some debug code for detecting endless loops, but with the changes here, the endless loop in my save game disappeared, so this issue may still show up again, and will be addressed in this pull request or a later one. This pull request is based on #115. Here is the main part of the summary of the last commit for this pull reqest:

Use the entry costs times the distance to target to estimate the future costs for a node in the A-star path finding algorithm. The original heuristic was the minimum road costs within a four times four square. This biased ship routes along costs with roads, railroads, or maglevs, even so the direct route over the ocean would be shorter.

The other problem with this heuristic is that it is not necessarily in the same scale as the distance, the trade A-star uses trade values that may be just half the value and the city A-star uses the build costs of the tile improvements or some measure that depends on it.

However, the new heuristics is still not perfect since it still underestimates or overestimates the future costs if some mountains are on the path, so it still may give sub-optimal paths. But that is a property of the heuristic and fine as long as we are not too far away from the optimum.


MartinGuehmann:
And this can also go in.

PR #119, opened 2019-04-22
Fix the AI 3

MartinGuehmann:
Another pull request for fixing the AI. This pull request includes #118.

Right now it includes rather minor thinks like not confusing Gaia computers with Gaia satellites. This stops the AI from building Gaia computers even so it already has enough so that it can now build the satellites.

Then a little cargo heli unload fix. It makes the AI unloading cargo in front of a city if the city is full even so the cargo is supposed to go to the city.

The AI now builds diplomats. However, the diplomats still don't establish embassies.
The AI now builds missionaries. However, they are still restricted to theocracy, even so the televangelists appear rather after theocracy.

Finally, we have a fix of two debug asserts and of a memory leak, plus some further cleanup.


LynxAbraxas:
Many thanks @MartinGuehmann for all this work, will test it.

> However, they are still restricted to theocracy, even so the televangelists appear rather after theocracy.

They appear later in the tech tree but if a civ sticks to theocracy for longer, building televangelists does work (needed #114 but otherwise tested from a human player view).
In regard to #129, how about making some AI civs prefer some govs over others making them differ to other AI civs?

MartinGuehmann:
> In regard to #129, how about making some AI civs prefer some govs over others making them differ to other AI civs?

They actually do, however they research all the stuff in the same order, except for a minor variation. Even so all the different personalities have their own advance list, which are just copies of each other and somebody forgot to modify.

Actually, all the different governments have ranks, 1, 2, 3, and 4 matching the city limits. So in principle, it should be possible that you can research all the governments of the same rank without having any of them. I don't know whether the tech tree is designed like this, but in fact it should be designed like this, so that you really have a choice.

Right now the AI will research the governments and will use the one of the highest rank available, even so it has not researched its preferred government. And once it got its preferred government it will switch, even so it has been for quite long in a non-preferred government.

LynxAbraxas:
> Even so all the different personalities have their own advance list, which are just copies of each other and somebody forgot to modify.

Would some random shuffeling suffice or do we need to take advance tree connections into account? I.e. only shuffle them if they are accessable at that stage.

MartinGuehmann:
> > Even so all the different personalities have their own advance list, which are just copies of each other and somebody forgot to modify.
> 
> Would some random shuffeling suffice or do we need to take advance tree connections into account? I.e. only shuffle them if they are accessable at that stage.

This is actually a bit off-topic so I put it to issue #130.

MartinGuehmann:
So I merged in the current master, would prefer to keep it like this and add anything more to a new PR.

LynxAbraxas:
Have played tested this a bit. So far no unexpected effects but also did not come across any of the cool new improvements so far. I'd say better in than without ;-)

LynxAbraxas:
I can confirm that diplomats are no built by the AI (and expelled by other AIs).

MartinGuehmann:
> I can confirm that diplomats are no built by the AI (and expelled by other AIs).

I hope you mean "are built".

MartinGuehmann:
> Have played tested this a bit. So far no unexpected effects but also did not come across any of the cool new improvements so far. I'd say better in than without ;-)

You also have to consider, that you are only supposed to see 18 of the 36 commits. Some of them are file cleanups (white space, unused functions), then optimizations for speed and memory, for instance the AI shouldn't consider goals if it has no units for them. Fixing memory leaks and crashes, those you don't really sea since unless that are super common crashes, which should have been caught by Activision, anyway. And then we have the Gaia Controller fixes, which should give the AI a chance to win the game by science, at least if it can cover 60% of the map with 60 cities.

MartinGuehmann:
> I'd say better in than without ;-)

Well, then let's merge.

LynxAbraxas:
> I hope you mean "are built".

Yes, I meant "are now built".

>  at least if it can cover 60% of the map with 60 cities.

Hm, why those values? Does that apply to a human player as well?

MartinGuehmann:
> > at least if it can cover 60% of the map with 60 cities.
> 
> Hm, why those values? Does that apply to a human player as well?

60 cities is the limit for the best governments. I allowed the AI to go a few cities over the limit, but then it would have to heavily invest in happiness, which currently it can't. Each city over the limit reduces the happiness in all cities by 1% or whatever the unit is. So eventually, you cripple your civilization with too many cities.

60% is the value your obelisks have to cover, depending on the number of Gaia satellites you have, each obelisks covers a radius of 5 to 8 tiles. On the large map of my test game the AI may cover 40% to 50%. Actually, the map has space for 230 to 250 cities, and the AIs now also fill that space. So the problem here is you cannot win the game by conquest, unless you conquer all the AI cities within a turn, but you cannot win the game by science, either. Here you must own a huge part of the worlds.

And of course these limits also apply to human players as, well. Human players might just more flexible here.

LynxAbraxas:
> AIs now also fill that space.

That's really cool. It is well noticeable that the AI has become much stronger since the original release. 

> So the problem here is you cannot win the game by conquest, unless you conquer all the AI cities within a turn

Why only in a single turn?

> you cannot win the game by science, either

Why? Why does it depend on the relative size of ones territory?

MartinGuehmann:
> > So the problem here is you cannot win the game by conquest, unless you conquer all the AI cities within a turn
> 
> Why only in a single turn?

My map has space for let's say 220 cities. The space is filled, also with undersea cities. And so you have to conquer them all and you cannot destroy them. Otherwise the AI would resettle the space.

You can have at max 60 cities in Virtual Democracy or Ecotopia; any other government allows less. Now we have 220 cities. If you have all of them, then you have 160 cities more then allowed. Each city over the limit costs you one point of happiness. That are in total 160 points. A city is content at 73 points. Let's the a city has 73 points without the limit. With the limit and 160 cities more than allowed, its happiness is then -87, in that case the city would revolt and split off. Of course you can increase the happiness in the city with buildings and wonders but that is limited and you get only 2 points per entertainer, so if you convert all citizens to entertainers in a size 60 city then you would get 120 extra points of happiness, that is still not enough, since your city would then have a happiness of 33, which still makes the city revolt.

> > you cannot win the game by science, either
> 
> Why? Why does it depend on the relative size of ones territory?

The science victory is about building the Gaia controller and for that you must build obelisks. The obelisks cover a certain radius. And the coverage must be 60% of the map so that you can win. And you can only build the obelisks in your own territory.

LynxAbraxas:
I see, many thanks for that explanation. I was not aware of that, never having finished a game so far. I guess with all that extra challenges, it's good that I read the modern age just slightly after 0 AD, so I have some time left to tackle all those coming problems.

PR #123, opened 2019-04-29
expecting leader names < 100 chars during GL-CI tests

LynxAbraxas:
to speed up test in case of missed sikuli events. 
This reduces the time of the name-game test by a factor of 10 in case a sikuli event gets lost.

### DIFF
--- a/tests/name-game.sikuli/name-game.py
+++ b/tests/name-game.sikuli/name-game.py
@@ -20,7 +20,7 @@ def main():
     hover(getLastMatch().getTarget().offset(0, 100)) # move mouse out of the text field for later matching
     Settings.TypeDelay = 0.1; # ctp2-SDL needs some more time
     count= 0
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
Year, these little things cause so much. The leader in the ProfileDB has a max size of 512.

LynxAbraxas:
I'd leave it at 100 (in case some backspace events are lost) in this case because we know it will be 'Julius Ceasar' in the GL-CI, tested on my GL-CI and works as expected, so ready to go in from my side.

MartinGuehmann:
> I'd leave it at 100 (in case some backspace events are lost) in this case because we know it will be 'Julius Ceasar' in the GL-CI, tested on my GL-CI and works as expected, so ready to go in from my side.

I did not suggest to go for 512, just noted that the 1024 were too much, anyway. What else did we have, 512 bytes for a city name. Well. Anyway, it can go in when Travis is done with #124.

MartinGuehmann:
Then it goes in.

PR #124, opened 2019-04-29
get '/' (for e.g. /reloadslic) with Shift+7

LynxAbraxas:
essential for testing if there is no number-pad emulation on your laptop and you don't have a spare keyboard around...


MartinGuehmann:
Actually, the problem is that the keyboard is not fully British if your keyboard is German, but the main fix would be done with SDL2. So this can now go in.

LynxAbraxas:
Many thanks for merging @MartinGuehmann. Yes SDL2 would be great for other things as well but is not as simple as this little change that made my day;-)

PR #126, opened 2019-04-30
Message for gold from trade through city

LynxAbraxas:
This PR re-enables the former bonus for trade routes through a city (which was disabled for unobvious reasons in https://github.com/civctp2/civctp2/commit/174224ad9acee585a40fb28d8f9bd46a1f5e4c07) and informs the recipent about this fact. The bonus can be seen like gold from tax on trade transit or from local trade arising from the transit.

The nice thing about this concept is that it can make it more profitable to treat a transit trade route well instead of pirating it because the bonus is added for every transit sity so if more than (currently) 3 transit cities are on a route the overall bonus of the transit is higher than the pirating value.  In the case of the screen shot below, the trade route passes through 4 cities (Comisense, Esfahan, Washington and Chicago):
ss_2019-04-30_13:08:23
yielding in total `4 * 22 = 88` instead of just `59` in case of successful piracy (which increases the probability that the owner cancels the route):

ss_2019-04-30_13:21:07

With https://github.com/civctp2/civctp2/pull/126/commits/f4e72d2675be3acb9290a7144488081d4c40377a no transit gold will be given if the route is pirated, either by oneself or others, motivating to look for pirates even on other civs routes. Therefore it might be a good to have another message in case of piracy telling who pirates that foreign route, which is currently not possible with the bug behind #75. If that is solved and the message works as epxected the message of https://github.com/civctp2/civctp2/pull/126/commits/a22052722a3eef1925f038b9484bb3ad9cf7b1b9 could actually be removed to reduce the message flood, since a player will get to know about this game concept as soon as he or other pirate such a transit route through one of his cities (thought this would be kind of quite an indirect info).

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -11281,10 +11281,15 @@ void CityData::GiveTradeRouteGold()
 					Unit fromCity = route.GetSource();
 					Unit toCity = route.GetDestination();
 
+					ROUTE_TYPE type;
+					sint32 good;
+					route.GetSourceResource(type, good);
+					
 					SlicObject * so = new SlicObject("359TradePassing");
 					so->AddRecipient(GetOwner());
 					so->AddGold(tgold) ;
 					so->AddCity(m_home_city); // m_home_city should equal fromCity
+					so->AddGood(good);
 					so->AddCity(fromCity);
 					so->AddCity(toCity);
 					so->AddCivilisation(fromCity.GetOwner());
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -11285,6 +11285,7 @@ void CityData::GiveTradeRouteGold()
 				    route.GetSourceResource(type, good);
 					
 				    if(!route->IsBeingPirated()){
+					/*
 					g_player[m_owner]->AddGold(tgold);
 					
 					SlicObject * so = new SlicObject("359TradePassing");
@@ -11296,6 +11297,7 @@ void CityData::GiveTradeRouteGold()
 					so->AddCity(toCity);
 					so->AddCivilisation(fromCity.GetOwner());
 					g_slicEngine->Execute(so);
+					*/
 					}
 				    else { // message that tgold is not earned from transit due to pirating
 					SlicObject * so = new SlicObject("358TradePassingPirated");
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -11275,7 +11275,8 @@ void CityData::GiveTradeRouteGold()
 				if((route.GetSource().GetOwner() != m_owner)
 				&&(route.GetDestination().GetOwner() != m_owner)
 				){
-					sint32 tgold = static_cast<sint32>(route->GetValue() * g_theConstDB->Get(0)->GetCityOnTradeRouteCoeff());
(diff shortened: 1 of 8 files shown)


LynxAbraxas:
Eventhough this works already, better wait for #75 to be solved, such that https://github.com/civctp2/civctp2/pull/126/commits/f4e72d2675be3acb9290a7144488081d4c40377a can be apropriately tested, since without pirating and getting transit bonus is unfair and unrealistic.

LynxAbraxas:
From https://github.com/civctp2/civctp2/issues/75#issuecomment-488140779
> Which one do you mean?

The `CITY_ON_TRADE_ROUTE_BONUS`
https://github.com/civctp2/civctp2/pull/126/commits/4fea6b9325a407e822036b03439074f3a84b4986#diff-1234622e6bc50faa7bdfefaec536ae91R393

> In fact, you enabled three things. 

Yes, good point, I will separate them.

> And the most plausible reason that they have been disabled is that those features haven't been finished.

True, but I checked at least  `CITY_ON_TRADE_ROUTE_BONUS`, which was there since the very first commit and was removed with https://github.com/civctp2/civctp2/commit/174224ad9acee585a40fb28d8f9bd46a1f5e4c07 which according to its commit message focused on " generating too many hills".
Apart from that I tested the changes and they worked as expected, partially proven by the screen shots.

MartinGuehmann:
> True, but I checked at least `CITY_ON_TRADE_ROUTE_BONUS`, which was there since the very first commit and was removed with [174224a](https://github.com/civctp2/civctp2/commit/174224ad9acee585a40fb28d8f9bd46a1f5e4c07) which according to its commit message focused on " generating too many hills".
> Apart from that I tested the changes and they worked as expected, partially proven by the screen shots.

That's indeed the one that should be enabled. Yeah, these SVN commits contain many unrelated things, I think we didn't like it spinning revision numbers without a reason.

LynxAbraxas:
With #127, this now work as expected, I wonder though, if this spams the human player with too many massages. @MartinGuehmann what do you think? Should I revert https://github.com/civctp2/civctp2/pull/126/commits/025931e54d1aa2f96d7feff117233d4783732875 and just keep the message in case of pirating?

MartinGuehmann:
> With #127, this now work as expected, I wonder though, if this spams the human player with too many massages. @MartinGuehmann what do you think? Should I revert [025931e](https://github.com/civctp2/civctp2/commit/025931e54d1aa2f96d7feff117233d4783732875) and just keep the message in case of pirating?

As far as I can see it is indeed too spammy, these I would prefer to have the the info in the trade manager.

LynxAbraxas:
For both or only in the non-pirating case?
In any way, how would you suggest to put that info into the trade manager?

MartinGuehmann:
> For both or only in the non-pirating case?
> In any way, how would you suggest to put that info into the trade manager?

The trade adviser gives you details: Caravans, Caravans being used, Income, Number of trade routes.

Here we could add trade though cities, income lost to piracy.

If we get pirated or if we are pirated then this information should go to the messages we get, anyway.

LynxAbraxas:
> Here we could add trade though cities, income lost to piracy.

I like that idea, however I would still go for a message in case a transit is pirated by a 3rd party because in the trade manager we could not give the info on who pirates, which I find quite interesting since (depending on the income lost) it tempts the player to consider to take counter actions against that civ. I see no other easy way to reveal the 3rd party otherwise.

LynxAbraxas:
Rebased on current master, basically ready to be merged. I'd create a new PR for the changes for the trade advisor.

MartinGuehmann:
> Rebased on current master, basically ready to be merged. I'd create a new PR for the changes for the trade advisor.

Good to know, it was a bit unclear whether you still wanted to add something, so then let't merge.

PR #125, opened 2019-04-29
WIP: message for the pirate victim

LynxAbraxas:
Follow up on #76 based on @MartinGuehmann comment https://github.com/civctp2/civctp2/pull/76#issuecomment-482749127, the message for the victim of piracy should not be based on the success of the attacker but instead on the fact that the trade gold is not added in case of piracy (especially by multiple civs on the same route).

Issue #116, opened 2019-04-14
diplomacy window never shows the anti-piracy option


LynxAbraxas:
It seems that currently the diplomacy window never shows the anti-piracy option like (from the procedures described in https://github.com/civctp2/civctp2/issues/75#issuecomment-483026718):

![ss_2019-04-11_18:50:47](https://user-images.githubusercontent.com/21012234/56098124-80abf280-5efd-11e9-945d-06efccec8254.png)

even if a trade route got pirated by a civ successfully (no concurrent pirating civs on the route) in multiple successive turns, see also #75.

LynxAbraxas:
Oddly, the AI players do get the option to ask for stopping piracy and that is passed to the player, this works already.

LynxAbraxas:
Solved with #127.
Seems the AI players were only getting the option towards the human player but not towards other AI players.

PR #127, opened 2019-04-30
Debug piracy issue

LynxAbraxas:
This PR solves the issue of #75, i.e. that the player does not get a colored square in the piracy column of the trade manager, nor the diplomatic option to ask for stopping piracy (#116) and is the base for PRs like #125 and #126.

### DIFF
--- a/ctp2_code/gs/gameobj/Army.cpp
+++ b/ctp2_code/gs/gameobj/Army.cpp
@@ -56,7 +56,7 @@ void Army::KillArmy()
 
 	CAUSE_REMOVE_ARMY cause = GetRemoveCause();
 
-	AccessData()->StopPirating();
+	AccessData()->StopPirating(); // do not execute pirating if army gets killed
 
 	Army tmp(*this);
 	tmp.SetRemoveCause(cause);
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -6201,7 +6201,9 @@ void ArmyData::AddOrders(UNIT_ORDER_TYPE order, Path *path, const MapPoint &poin
 		}
 	}
 
-	StopPirating(); // why?
+	if(point != m_pos || order != UNIT_ORDER_INTERCEPT_TRADE){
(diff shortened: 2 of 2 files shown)


LynxAbraxas:
@MartinGuehmann what would you suggest to do with `StopPirating` in `AutoAddOrders`
https://github.com/civctp2/civctp2/blob/0668dd199c2519ff8ab7d2871e7c7a89e21c8e52/ctp2_code/gs/gameobj/ArmyData.cpp#L6057
and `AutoAddOrdersWrongTurn`
https://github.com/civctp2/civctp2/blob/0668dd199c2519ff8ab7d2871e7c7a89e21c8e52/ctp2_code/gs/gameobj/ArmyData.cpp#L6095
, same code snippet as well?

LynxAbraxas:
Wonderful! Works as expected and solves #75 and #116. 
@MartinGuehmann  Can go in from my side.

MartinGuehmann:
> , same code snippet as well? Do you know when those are called?

As you have already figured out: Yes. And then we wait for Travis.

LynxAbraxas:
> And then we wait for Travis.

Of corse (it had finished with the old ones but I just noticed I still had the debug fprintf in). 
I'll start rebasing the other PRs on to this to sort things out a bit.

MartinGuehmann:
Then this can go in.

LynxAbraxas:
Great! This bug really bugged me. It seems that it also has an interesting effect on the AI players. I guess that the AI players only got to know the pirating from the human player and not from the other AIs because of this bug. This might increase diplomatic interaction between the AI players.
I also observed that the pirating animation of AI units now only happens once (while it used to happen every turn, see e.g. the test game I uploaded) because the pirating order stays over turns as long as no action is taken that "turns it off". Perhaps this even stops the AI from trying to use transporters for pirating (have to observe this for longer).

@MartinGuehmann Would you say that this bug already was in the original version of cpt2 or was it introduced when pimping up the AI, making special use of `AddOrder` then?
Do you know when `AutoAddOrders` and `AutoAddOrdersWrongTurn` are used?

Issue #131, opened 2019-05-03
Pirating state of trade routes not saved/loaded


LynxAbraxas:
Now that #127 is in, it seems that the pirating states of trade routes are not saved/loaded. E.g. when pirating in the test-game of #75, ending the turn then the trade manager shows the pirating color orange. However when saving and re-loading that game, the pirating column color is back to white. 
Same holds for the action: pirating in the test-game then save and load before ending the turn will lead to no pirating for blue in the next turn.
I'd guess that the `m_piratingArmy` from 
https://github.com/civctp2/civctp2/blob/0668dd199c2519ff8ab7d2871e7c7a89e21c8e52/ctp2_code/gs/gameobj/TradeRouteData.cpp#L618-L621
is not saved, not sure about `m_isPirating` here
https://github.com/civctp2/civctp2/blob/0668dd199c2519ff8ab7d2871e7c7a89e21c8e52/ctp2_code/gs/gameobj/ArmyData.cpp#L508
at least allown this would not be enough since `IsBeingPirated` needs  `m_piratingArmy`
https://github.com/civctp2/civctp2/blob/0668dd199c2519ff8ab7d2871e7c7a89e21c8e52/ctp2_code/gs/gameobj/TradeRouteData.cpp#L628-L631
@MartinGuehmann What do you think?

MartinGuehmann:
The problem is that m_piratingArmy is not saved.

https://github.com/civctp2/civctp2/blob/0668dd199c2519ff8ab7d2871e7c7a89e21c8e52/ctp2_code/gs/gameobj/TradeRouteData.cpp#L376-L461

You can reuse the slot of m_path_selection_state, it also occupies 32 bits like Army.

So you have to replace the lines

archive.PutSINT32(m_path_selection_state);
m_path_selection_state = archive.GetSINT32();

by

m_piratingArmy.Serialize(archive);

each.

Possible values for m_path_selection_state are 0, 1, and 2. Actually, it is also somewhere in the code set, but it is not used. So it shouldn't be a problem to reuse the spot. 0, 1, and 2 are not valid IDs for armies, IsValid should care about it.

LynxAbraxas:
Solved with #132, many thanks @MartinGuehmann.

PR #132, opened 2019-05-04
save and load pirating army:

LynxAbraxas:
idea from @MartinGuehmann in #131

### DIFF
--- a/ctp2_code/gs/gameobj/TradeRouteData.cpp
+++ b/ctp2_code/gs/gameobj/TradeRouteData.cpp
@@ -387,7 +387,7 @@ TradeRouteData::Serialize(CivArchive &archive)
 		archive.PutUINT32(m_color);
 		archive.PutUINT32(m_outline);
 		archive.PutSINT32(m_selectedIndex);
-		archive.PutSINT32(m_path_selection_state);
+		m_piratingArmy.Serialize(archive);
 		archive.PutSINT32(m_valid);
 		archive << m_payingFor;
 		archive << m_gold_in_return;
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Amazing, just works! Many thanks @MartinGuehmann!

Any idea what `m_path_selection_state` is/was meant for?

MartinGuehmann:
> Amazing, just works! Many thanks @MartinGuehmann!
> 
> Any idea what `m_path_selection_state` is/was meant for?

To indicate the state of a selected path of a trade route. The code has some has a getter and setter for it but only the setter is used. So the information is set but never used. And I would say that we don't have any use for that, either.

LynxAbraxas:
It now becomes noticable, that apparently the count of how many times a trade route was pirated is not saved/loaded either. I.e. by saving and loading the AI keeps up the trade route whereas without saving+loading it cancels the route after it has been pirated 3 times. So far I was not able to find the setting for this 3 times count. 
@MartinGuehmann any idea where that might come form?

MartinGuehmann:
That is hold in Diplomat::m_piracyHistory, but it is not saved, anyway that is information that could also be stored in TradeRouteData. As far as I can see the current approach just wastes memory.

So what we need in TradeRouteData is a member that is not really be used and thus can be reused and that is saved. At least we have to save the year and that is a sint32. We could use the spot of  m_recip for the turn actually it is not used and can thus be removed. The number of accumulated events can be a sint16 or a sint8. For that we could split the spot of m_crossesWater it is a BOOL so occupies 32 bits and is also saved as 32 bit integer.

LynxAbraxas:
Not able to fully follow you here. Could you create a commit on top that does what you suggest?

MartinGuehmann:
> Not able to fully follow you here. Could you create a commit on top that does what you suggest?

That's a bit more complicated, so I guess I merge this first.

MartinGuehmann:
> Not able to fully follow you here. Could you create a commit on top that does what you suggest?

This is now #133. Please test, whether everything works as intended.

PR #133, opened 2019-05-05
Save and load piracy history

MartinGuehmann:
Save and load the piracy history so that the AI would cancel its trades routes when they have been pirated too often even after reloads.

This pull request fixes the issue found in https://github.com/civctp2/civctp2/pull/132#issuecomment-489365578.

Note that so far I only have tested whether, it can load old save games with trade routes and can save and reload those again.

### DIFF
--- a/ctp2_code/ai/diplomacy/Diplomat.h
+++ b/ctp2_code/ai/diplomacy/Diplomat.h
@@ -75,40 +75,6 @@ class Diplomat;
 class Diplomat
 {
 public:
-	struct PiracyHistory
-	{
-		PiracyHistory()
-		:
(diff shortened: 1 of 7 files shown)


LynxAbraxas:
Just great! Works as expected with the pirating test game of #75, ordering to pirate once then saving and loading before ending the turn. Next turn the trade manager reports player 2 as pirating as do the messages. After doing so 3 times, player 2 cancels the trade route as expected.
Many thanks @MartinGuehmann, I wouldn't have dared to touch this.

MartinGuehmann:
So then we can merge this.

PR #134, opened 2019-05-05
name of good in pirating messages

LynxAbraxas:
eases identification, since the good is visible:


LynxAbraxas:
Just an extension to #76 and #71, works as expected.

MartinGuehmann:
> Just an extension to #76 and #71, works as expected.

Well then it can go in.

PR #136, opened 2019-05-07
More power graph options

MartinGuehmann:
Adds more options to the power graph, such as trade, gold, number of cities, population, land area, sea area, and units.

The category military was changed, now it shows the attack and defense strength of military units. Before, it was the unit build cost divided by two.  A military unit is defined as a unit with attack > 0. Now the category unit shows the build costs of all units.

The category gold was added and fixed. However, at some time there seems that too much trade is generated. And land and sea area, still do not work correctly. Sea area is right always zero and land area seems to be total area.


MartinGuehmann:
There seem to be some problems with what is shown, sometimes there seems to be too much trade shown and it showed some population decrease without explanation.

However, I consider these things unrelated and should be addressed in another pull request.

LynxAbraxas:
> sometimes there seems to be too much trade shown

In what way "too much"? 
I'll give it a test.

MartinGuehmann:
> > sometimes there seems to be too much trade shown
> 
> In what way "too much"?
> I'll give it a test.

After, 200 turns or so, I got a peak in the power graph, it looks like some random high value. I assume it is something that is not initialized, but I haven't found the problem, yet.

And in fact, the problem is not related to this pull request, in my last AI only game I also got such a peak in the Overall graph.

LynxAbraxas:
Tested for some time, seems to work as expected. I had no unusual peaks so far. Good to go.

MartinGuehmann:
Then let's put it in.

PR #137, opened 2019-05-12
Splitter civs inherit the age of their parent civ

MartinGuehmann:
So far when a city becomes independent and forms a new civilization it inherits all the advances of its parent civilization. Since the advances define, in which age a civilization is, the age should also be inherited. Otherwise the new civilization is the first age until it discovers a new advance.
..\ctp2_code\gs\gameobj\Civilisation.cpp

### DIFF
--- a/ctp2_code/gs/gameobj/Civilisation.cpp
+++ b/ctp2_code/gs/gameobj/Civilisation.cpp
@@ -164,9 +164,10 @@ void civilisation_CreateNewPlayer(sint32 pi, sint32 old_owner)
 	    (old_owner >= 0) && g_player[old_owner]
 	   )
 	{
-	    delete g_player[pi]->m_advances;
+		delete g_player[pi]->m_advances;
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Sounds reasonable.
When does a city becomes independent and form a new civilization? Never saw that happen in the games I played.

MartinGuehmann:
> Sounds reasonable.
> When does a city becomes independent and form a new civilization? Never saw that happen in the games I played.

If you start a new game then you can set the maximum number of players for instance to 20 or 31, then the game adds whenever cities split off, if you have the maximum number of players or more in the game then those cities become Barbarian.

LynxAbraxas:
Tested for some time but not encountered the formation of a new civilization, so cannot say if the disired effect happens. In any case this did not seem to introduce any unwanted effects so I'd say it is good to go in.

MartinGuehmann:
Then this goes in.

PR #138, opened 2019-05-13
Fix a crash in the piracy at risk code introduced by #133

MartinGuehmann:
Fix a crash in the piracy at risk code introduced by #133.
..\ctp2_code\ai\diplomacy\diplomat.cpp

### DIFF
--- a/ctp2_code/ai/diplomacy/diplomat.cpp
+++ b/ctp2_code/ai/diplomacy/diplomat.cpp
@@ -4457,8 +4457,9 @@ bool Diplomat::GetTradeRoutePiracyRisk(const Unit & source_city, const Unit & de
 		{
 			TradeRoute route = city.CD()->GetTradeSourceList()->Access(r);
 
-			if ((route->GetSource() == source_city) &&
-			    (route->GetDestination() == dest_city) &&
+			if (route->GetPirate() >= 0 &&
+			    route->GetSource() == source_city &&
+			    route->GetDestination() == dest_city &&
 			    !AgreementMatrix::s_agreements.HasAgreement(route->GetPirate(),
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Cool! Under what circumstances did you hit that crash?
How did it get in #133, what was the reason to replace `piracy_iter`?

MartinGuehmann:
> Cool! Under what circumstances did you hit that crash?

Well, I merged master into #336 and just let the AIs play against each other and eventually the game crashed. So I would expect if you just test long enough you will hit hit.

> How did it get in #133, what was the reason to replace `piracy_iter`?

Originally, it iterated through the piracy history list, which only contained trade routes that where actually pirated, but that was not saved, and saving it would have me required to change the save game format, at least I thought so. And so, I moved this information to the trade route itself. Conceptionally, that is better, because that data belongs to the trade route. However, in principle it uses more memory, because it is now in the trade route, but on the other hand, I reused existing but unused variables, so that the memory need is the same. And I think memory consumption of the trade route is not that important because you don't have so many TradeRouteData objects in the game, it is much more important for the Goal class.

The difference is that I now iterate over all trade routes and thus the pirating player could now be -1 so that you try to deference g_player[-1], which then crashes.

LynxAbraxas:
> saving it would have me required to change the save game format, at least I thought so. And so, I moved this information to the trade route itself

That's a very clever idea. Many thanks for getting into this so thoroughly. I wouldn't have been able to come up with that solution.
The new dynamics this brings to the trade routes makes the game much more enjoyable.

LynxAbraxas:
Tested for a while and all seems to work as expeced. So I'd say this PR is good to go.
Thanks @MartinGuehmann for catching this.

MartinGuehmann:
Then we can put in this as, well.

Issue #141, opened 2019-05-19
City influence not expanding if nearby city is destroyed


LynxAbraxas:
If a nearby (foreign) city is destroyed a city close by does not expand its region of influence even though it would have a larger region of influence if the former city would not have existed.
This is an unexpected and unlogical behavior. Possibly related to #73?

MartinGuehmann:
No it is not related to #73. It is just that the nearby cities aren't updated. Maybe there is in some code that is supposed to do it, but it may be executed before the city is actually dead or its influence has been removed.

LynxAbraxas:
Solved with #143, which has its own issue for optimization #151.

PR #143, opened 2019-06-01
Update city influence at end of turn to account for newly available territory

LynxAbraxas:
PR concerning #141:
The idea is to call `AdjustSizeIndices`
https://github.com/civctp2/civctp2/blob/247c8dc9c9be287194605ea5ba1f555884f3e0d7/ctp2_code/gs/gameobj/CityData.cpp#L8574-L8654
at the end of every turn to enable accounting for newly available territory if an adjacent city was destroyed or has shrunk. It is not clear to me though where growing is accounted for since
https://github.com/civctp2/civctp2/blob/247c8dc9c9be287194605ea5ba1f555884f3e0d7/ctp2_code/gs/gameobj/CityData.cpp#L8624-L8631
only seems to cover shrinkage, tough it only seems to consider adding city-ownership to a tile and not its removal.
@MartinGuehmann any idea on this? Am I investigating the wrong code sections?
Should the accounting be done else where than at the end of each turn, e.g. at the beginning?

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -5155,6 +5155,9 @@ void CityData::EndTurn()
 	}
 
 	m_build_queue.EndTurn();
+
+	// let city expand if adjacent city is gone or has shrunk
+	GenerateCityInfluence(m_home_city.RetPos(), m_sizeIndex);
 }
 
 void CityData::CalcHappiness(sint32 &virtualGoldSpent, bool isFirstPass)
--- a/ctp2_code/gs/gameobj/CityInfluenceIterator.cpp
+++ b/ctp2_code/gs/gameobj/CityInfluenceIterator.cpp
@@ -129,7 +129,7 @@ bool ExpandInfluence(Unit &city, const MapPoint &centerPos, MapPoint curPos,
 	Cell *cell = g_theWorld->GetCell(curPos);
 
 	if(cell->GetScratch() != 0) return false;
-	if(cell->GetCityOwner().m_id != 0 && cell->GetCityOwner().m_id != city.m_id) return false;
+	if(cell->GetCityOwner().m_id != 0 && cell->GetCityOwner().m_id != city.m_id) return false; // prevents assignment of tiles owned by other adjacent city
 	if(UnitData::GetDistance(centerPos, curPos, 0) > rec->GetSquaredRadius()) return false;
 
 	cell->SetCityOwner(city);
@@ -283,7 +283,7 @@ void GenerateCityInfluence(const MapPoint &cpos, sint32 size)
 {
 	Assert(g_theWorld->GetCell(cpos)->HasCity());
 	Unit					city		= g_theWorld->GetCell(cpos)->GetCity();
-	sint32 const			intRadius	= RadiusFromIndex(size);
+//	sint32 const			intRadius	= RadiusFromIndex(size);
 	ClearScratch(cpos, size);
 	g_theWorld->GetCell(cpos)->SetScratch(1);
 


MartinGuehmann:
> PR concerning #141:
> The idea is to call `AdjustSizeIndices`
> https://github.com/civctp2/civctp2/blob/247c8dc9c9be287194605ea5ba1f555884f3e0d7/ctp2_code/gs/gameobj/CityData.cpp#L8574-L8654
> 
> 
> at the end of every turn to enable accounting for newly available territory if an adjacent city was destroyed or has shrunk. It is not clear to me though where growing is accounted for since

That's not really nice, because that shouldn't be done each turn, but when there is really need like when the neighbor city is destroyed. Actually, you destroy a city and your city does not grab immediately the space is odd. So you would search for any city within radius of two or three times the empire border distance and then update the cities. Of course also when a city shrinks.

https://github.com/civctp2/civctp2/blob/247c8dc9c9be287194605ea5ba1f555884f3e0d7/ctp2_code/gs/gameobj/CityData.cpp#L8624-L8631
> 
> 
> only seems to cover shrinkage, tough it only seems to consider adding city-ownership to a tile and not its removal.
> @MartinGuehmann any idea on this? Am I investigating the wrong code sections?
> Should the accounting be done else where than at the end of each turn, e.g. at the beginning?

Well, I assume that GenerateCityInfluence handles when a city grows. Actually the code first removes all the influence of the city except for the city tile itself. And then it add all the tiles again, which is a bit illogical. The other problem is that the code seems to remove city ownership from tiles that are not necessarily owned by the city.

LynxAbraxas:
> That's not really nice, because that shouldn't be done each turn

That's what I thought first as well, but how to handle the situation when an adjacent city shrinks (e.g. by bomardment or by starving)? Either all cities have to be updated when `ChangePopulation` is called or at every turn. I'm not sure if we catch all cases with adding that code to `ChangePopulation` but that might be an option.

> And then it add all the tiles again, which is a bit illogical.

Probably the easiest to implement.

> The other problem is that the code seems to remove city ownership from tiles that are not necessarily owned by the city.

Hm, in what situation would that matter or is it just unnecessary work?

MartinGuehmann:
> > That's not really nice, because that shouldn't be done each turn
> 
> That's what I thought first as well, but how to handle the situation when an adjacent city shrinks (e.g. by bomardment or by starving)? Either all cities have to be updated when `ChangePopulation` is called or at every turn. I'm not sure if we catch all cases with adding that code to `ChangePopulation` but that might be an option.

No, not all cities only those cities that are in a radius let's say of 3 times the empire border radius.

> 
> > And then it add all the tiles again, which is a bit illogical.
> 
> Probably the easiest to implement.
> 
> > The other problem is that the code seems to remove city ownership from tiles that are not necessarily owned by the city.
> 
> Hm, in what situation would that matter or is it just unnecessary work?

Well your city shrinks and since it overlaps with another city it also looses tiles. Actually, I have never seen this, or I have never cared about it.

LynxAbraxas:
> No, not all cities only those cities that are in a radius let's say of 3 times the empire border radius.

What that make a noticable difference in processing time? What about the cities of your own empire that shrink? I'd want that same effect for those since I ended up in a situation where I would like to give up a city in order for an adjacent one to grow better.

> Well your city shrinks and since it overlaps with another city it also looses tiles.

Hm, I'd say if a city shrinks it should not loose tiles in the new radius that it owned before the shrinkage.

MartinGuehmann:
> > No, not all cities only those cities that are in a radius let's say of 3 times the empire border radius.
> 
> What that make a noticable difference in processing time? What about the cities of your own empire that shrink? I'd want that same effect for those since I ended up in a situation where I would like to give up a city in order for an adjacent one to grow better.

Why would you check any city that is too far away from your shrinking city that it could share any tiles? It seems that you also get the empire border radius wrong, each city generates a border radius, start a new game and found a city and then you see what it is. These colored borders on the map. The border radius is now the distance from your city to the border. Obviously, any city that is within that radius, may need to expand its borders.

In principle, you only need to get the city owner of a tile. And if I think more about it, you only have to collect the city owners in the city radius + 1, and then update those cities. It is a bit easier than I thought. This way you need only to check four to five cities and that is much better than checking 100 or 200 cities each turn.

I wouldn't really think about processing time here, but I would rather avoid checking more cities then necessary.

> > Well your city shrinks and since it overlaps with another city it also looses tiles.
> 
> Hm, I'd say if a city shrinks it should not loose tiles in the new radius that it owned before the shrinkage.

The second "it" refers to the other city, sorry that the wording isn't clear. Let's  see whether I can word it more clearly: If a city shrinks than only this city should loose tiles, but not any other cities around. And that may happen with that code. Even so I haven't seen it.

LynxAbraxas:
> Why would you check any city that is too far away from your shrinking city that it could share any tiles?

1. it is the easiest to implement ;-)
2. checking which cities shrink and then determining their nearby cities seems to be more computationally expensive than just doing it for every city independent of whether it has an effect or not
3. I'm not sure how to ensure that I get all effected cities, just placing the code in `ChangePopulation` might not catch all cases

> I wouldn't really think about processing time here, but I would rather avoid checking more cities then necessary.

I do agree to that idea in principle but I fear I cannot offer a PR realizing that and considering point 2 above, I'd just go for the change from https://github.com/civctp2/civctp2/pull/143/commits/da628a052e564d16dc4bd707599b78736563bb3e undless you see any problems with that.

Tested https://github.com/civctp2/civctp2/pull/143/commits/da628a052e564d16dc4bd707599b78736563bb3e and the effect seems to be just as I expected:

Before the end of a turn:
ss_2019-06-02_16:23:57

After the end of the trun:
ss_2019-06-02_16:30:51

Stable city borders there after (2 turns later):
ss_2019-06-02_16:52:08

Same here:

Before the end of a turn:
ss_2019-06-02_16:24:58

After the end of the trun:
ss_2019-06-02_16:28:54

It is not clear to me (yet) why the border between the two adjacent cities (where the dates are) is not re-ordered (@MartinGuehmann any idea why?), but it is behaving as I would expect, i.e. preserving border shapes if there is no shrinkage.

**City shrunken:**

Before the end of a turn:
ss_2019-06-02_16:47:14

After the end of the trun:
ss_2019-06-02_16:50:57

**City gone:**

Before the end of a turn:
ss_2019-06-02_17:04:33

After the end of the trun:
ss_2019-06-02_17:18:56

LynxAbraxas:
> It is not clear to me (yet) why the border between the two adjacent cities (where the dates are) is not re-ordered (@MartinGuehmann any idea why?)

Since `ExpandInfluence` is used by `GenerateCityInfluence` this also seems to be covered by:
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityInfluenceIterator.cpp#L132

MartinGuehmann:
> > Why would you check any city that is too far away from your shrinking city that it could share any tiles?
> 
> 1. it is the easiest to implement ;-)
> 2. checking which cities shrink and then determining their nearby cities seems to be more computationally expensive than just doing it for every city independent of whether it has an effect or not
> 3. I'm not sure how to ensure that I get all effected cities, just placing the code in `ChangePopulation` might not catch all cases
> 
> > I wouldn't really think about processing time here, but I would rather avoid checking more cities then necessary.
> 
> I do agree to that idea in principle but I fear I cannot offer a PR realizing that and considering point 2 above, I'd just go for the change from [da628a0](https://github.com/civctp2/civctp2/commit/da628a052e564d16dc4bd707599b78736563bb3e) undless you see any problems with that.

I disagree with point 1 and 2. You need to find the cities that might have to claim tiles, which are the cities in the surrounding:

```
		RadiusIterator it(m_home_city.RetPos(), oldSizeIndex + 1); 
		for(it.Start(); !it.End(); it.Next())
		{
			Unit cityOwner = g_theWorld->GetCell(it.Pos())->GetCityOwner(); 
			if(cityOwner != m_home_city && cityOwner is not in list) 
				add cityOwner to list; 
		}

		Randomize the elements in list;
		for each element in list adjust the city influence;
```

Sure it is a mix of real c++ and pseudo-code. Note that it is a RadiusIterator and not a CityInfluenceIterator, so I assume it walks through all the tiles in the radius than through those that are only owned by the city.

This code only needs to be called by CityData::AdjustSizeIndices(), this means you already have the city that shrinks. And this code gives you the cities that may also have to adjust their influence radius. This code is only used when a city actually shrinks. That is computationally less expensive then doing it every on every city.

Then we have point 3 left: Whether it is enough to only place it in CityData::AdjustSizeIndices(). Since this covers only shrinking, but not disbanding or destroying. However, that are things to figure out.

LynxAbraxas:
> `Randomize the elements in list;`

I guess you propose randomization in order to give equal chance to each city to claim newly available tiles first, which I find appropriate.

>  This code is only used when a city actually shrinks. That is computationally less expensive then doing it every on every city.

Sure, it's just that the change in https://github.com/civctp2/civctp2/commit/da628a052e564d16dc4bd707599b78736563bb3e already does not create a noticeable time-delay, it's somewhere below 1s.

> disbanding or destroying

Those I'd find important to be covered, since I noticed the problem when I disbanded one of my cities to give room for a neighbouring one of mine.

I'm not bound to this PR and am happy to close this PR in favour of one from you implementing a similar effect, please go ahead.

LynxAbraxas:
I'm wondering what this code is meant for:
https://github.com/civctp2/civctp2/blob/247c8dc9c9be287194605ea5ba1f555884f3e0d7/ctp2_code/gs/gameobj/CityData.cpp#L8624-L8631
It seems to set the cell's city to the former, larger size but for what reason?

LynxAbraxas:
Tested this for a while and works as expected so removing the WIP, though it does not come with the discussed optimizations, this PR could be merged and a new one opened for an optimized version.
One advantage of this approach is that the civ who shrinks or destroys/disbands a city first gets the chance to claim new tiles by nearby cities of the same civ before other civs can claim newly available tiles. I would find it quite unfair if another civ would get that chance before me due to randomization of the processing of surrounding cities.

LynxAbraxas:
@MartinGuehmann Do you mind merging this? I'd say it is at least better than it was before and has no noticeable impact on the evaluation speed during game-play. I can open an issue tracking your suggestions for optimization if you want.

MartinGuehmann:
> @MartinGuehmann Do you mind merging this? I'd say it is at least better than it was before and has no noticeable impact on the evaluation speed during game-play. I can open an issue tracking your suggestions for optimization if you want.

Yeah, please open the issue. I think that I won't get to it too soon.

So then let's merge it.

PR #146, opened 2019-06-15
List gold lost to piracy in trade advisor

LynxAbraxas:
Motivated by https://github.com/civctp2/civctp2/pull/126#issuecomment-489472244 this PR offers an entry in the advisor of the trade manager that lists the total amount of gold lost due to piracy.

### DIFF
--- a/ctp2_code/ui/interface/trademanager.cpp
+++ b/ctp2_code/ui/interface/trademanager.cpp
@@ -540,10 +540,7 @@ void TradeManager::UpdateAdviceWindow()
 	for(i = 0; i < g_player[pl]->m_all_cities->Num(); i++) {
 		Unit city = g_player[pl]->m_all_cities->Access(i);
 		totalRoutes += city.CD()->GetTradeSourceList()->Num();
-		sint32 r;
-		for(r = 0; r < city.CD()->GetTradeSourceList()->Num(); r++) {
-			totalProfit += city.CD()->GetTradeSourceList()->Access(r)->GetValue();
-		}
+		totalProfit += city.CD()->GetGoldFromTradeRoutes(); // takes piracy and wonder bonus into account
 	}
 
 	child = (ctp2_Static *)aui_Ldl::GetObject(s_tradeAdviceBlock, "Profit");
--- a/ctp2_code/ui/interface/trademanager.cpp
+++ b/ctp2_code/ui/interface/trademanager.cpp
@@ -536,16 +536,23 @@ void TradeManager::UpdateAdviceWindow()
 		child->SetText(buf);
 	}
 
-	sint32 i, totalProfit = 0, totalRoutes = 0;
+	sint32 i, totalProfit = 0, totalPiracy = 0, totalRoutes = 0;
 	for(i = 0; i < g_player[pl]->m_all_cities->Num(); i++) {
 		Unit city = g_player[pl]->m_all_cities->Access(i);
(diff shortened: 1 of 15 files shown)


LynxAbraxas:
While trying to implement this, I noticed that the total trade profit listed in the advisor does neither take piracy nor wonder bonus into account, this is changed with https://github.com/civctp2/civctp2/pull/146/commits/a3463610dd0d578cc443d3eba7ef70d1509d6016.

However, the value (931 in the screen-shot) still does not equal that in the details of the empire-manager (928):

ss_2019-06-15_16:10:30

ss_2019-06-15_16:11:04

Both use the same function
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/ui/interface/CauseAndEffectTab.cpp#L890
`CalculateGoldFromResources` from:
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityData.cpp#L4433-L4461
which is called at the beginning of a turn, so I'm not sure where the difference comes from.

MartinGuehmann:
> which is called at the beginning of a turn, so I'm not sure where the difference comes from.

Actually, it is called when the CauseAndEffectTab is updated and with your change in a346361 it is also updated when the trade manager is called. As far as I can see these things should be fine.

The difference between 931 and 928 smells like a rounding problem, which is probably due to some implicit rounding, but as far as I could see it shouldn't be different.

However, I cannot see the piracy lost in the trade adviser, what you could do is to resize the window so that it aligns with main window.

LynxAbraxas:
> However, I cannot see the piracy lost in the trade adviser

The screen-shots were just for https://github.com/civctp2/civctp2/commit/a3463610dd0d578cc443d3eba7ef70d1509d6016.

The piracy loss looks like:

ss_2019-06-15_22:57:53

> what you could do is to resize the window so that it aligns with main window.
 
I tried but a mere change in the height just adds a black extension:

ss_2019-06-15_17:02:07

LynxAbraxas:
Apart from the possible rounding discrepancy, this seems to work as expected and could go in.

MartinGuehmann:
> So far I could not find where the background image comes from. If it would only be used for the trade advisor, I could extent it manually but I'm not sure if that would suite other advisors.

It is in the zfs file and called upfg400.rim. Probably, you can just move up the entries. If we want to update the image then we have to do it for all advisers, which would also be fine.

I just have to figure out were to upload all these tools, I guess I put them into a subfolder of the project like ctp2_tools. The ZFS Reader at least works with Wine, I guess even better than with Windows 10. :D

By the way, I also found the Activision's sprite make tool.

MartinGuehmann:
> Apart from the possible rounding discrepancy, this seems to work as expected and could go in.

I guess we can look at the rounding problem or whatever that actually is later. So let's merge.

LynxAbraxas:
> I just have to figure out were to upload all these tools, I guess I put them into a subfolder of the project like ctp2_tools.

Sounds reasonable.

>  I also found the Activision's sprite make tool.

That's cool. Do you have the sources as well?

MartinGuehmann:
No, I don't have the sources.

LynxAbraxas:
> No, I don't have the sources.

That's a real pity. Any way, I think they came from the ctp2 code, because one very specific error-message concerning only the sprite creation is reported on a page describing the use of the sprite make tool.

PR #147, opened 2019-06-15
List transit gold in trade advisor

LynxAbraxas:
Also motivated by https://github.com/civctp2/civctp2/pull/126#issuecomment-489472244 his PR (based on #146) offers an entry in the advisor of the trade manager that lists the total amount of gold earned due to trade transit.

It contains a correction of a bug that was due to commenting too much: https://github.com/civctp2/civctp2/pull/147/commits/bc59dc02056704a5a9cebc59a0a3ad3531594560


LynxAbraxas:
Due to the different way of assigning transit gold
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityData.cpp#L5119-L5122
the bookkeeping is problematic and probably the reason for the inappropriate value of 540263293:

ss_2019-06-15_23:20:19

@MartinGuehmann any idea what could be the problem here? 
`m_goldFromTransitRoutes` so far is not saved, in contrast to `m_goldFromTradeRoutes` and `m_goldLostToPiracy`:
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityData.cpp#L640
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityData.cpp#L802
If this is in general the way to go do you see a chance to squeeze this value into another unused varaible in the save-format somewhere?

What are the lines here 
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityData.cpp#L383
doing?

MartinGuehmann:
Does m_goldFromTransitRoutes needs to be saved?

Line 640 saves m_goldFromTradeRoutes and line 802 loads m_goldFromTradeRoutes. If you want to save m_goldFromTransitRoutes then you have to find another sint32 that is not used but saved.

Line 383 initialized m_goldFromTradeRoutes to 0, which you have to do, because in c++ variables just happen to have the value that is accidentally in memory at that place. That might be your problem. I assume that you just opened a save game and looked at that value. After an end turn that should be gone.

LynxAbraxas:
> Does m_goldFromTransitRoutes needs to be saved?

Good point! I guess not, which makes life easier;-)

> Line 383 initialized m_goldFromTradeRoutes to 0, which you have to do, because in c++ variables just happen to have the value that is accidentally in memory at that place. That might be your problem.

Many thanks @MartinGuehmann for that explanation, never saw this kind of initialization before (something new every day...), only know it in the constructor like `{m_goldFromTransitRoutes= 0;}`

LynxAbraxas:
> I assume that you just opened a save game and looked at that value. After an end turn that should be gone.

True:

ss_2019-06-16_00:34:42

MartinGuehmann:
> > Line 383 initialized m_goldFromTradeRoutes to 0, which you have to do, because in c++ variables just happen to have the value that is accidentally in memory at that place. That might be your problem.
> 
> Many thanks @MartinGuehmann for that explanation, never saw this kind of initialization before (something new every day...), only know it in the constructor like `{m_goldFromTransitRoutes= 0;}`

Fromafar did these things. If I remember correctly, it is a speed optimization. And is especially important for objects, because these would be initialized with a standard constructor first, which is not so nice if you already have some arguments to pass to another constructor. And here you would also call parent constructors.

Actually, we have quite some around here. So that you could look up more example, I think in the GUI section you have quite some.

So is the last screenshot current? If yes, could you move up the values so that the fit into the window. And could you reorder them so that the profits are together, followed by the piracy lost? The number of trade routes can go below or above.

LynxAbraxas:
> So is the last screenshot current? If yes, could you move up the values so that the fit into the window. And could you reorder them so that the profits are together

So how about:

ss_2019-06-16_17:41:02

MartinGuehmann:
> > So is the last screenshot current? If yes, could you move up the values so that the fit into the window. And could you reorder them so that the profits are together
> 
> So how about:
> 
> ss_2019-06-16_17:41:02

The order is good. However, my impression is that it is a bit too squeezed. But I don't know how much space we can use there. I thought that there still was some, but of course this way we are on the save side.

LynxAbraxas:
> But I don't know how much space we can use there.

Me neither but since the advisor text changes in length depending on the advice and the city names involved, I decided to leave "Caravans available" at the original position.
Tested for some time now, good to go in from my side.

MartinGuehmann:
> > But I don't know how much space we can use there.
> 
> Me neither but since the advisor text changes in length depending on the advice and the city names involved, I decided to leave "Caravans available" at the original position.
> Tested for some time now, good to go in from my side.

Yeah, it looks a bit awkward, but we can take it like this. In case of doubt we can still adjust it later.

LynxAbraxas:
@MartinGuehmann 
In relation to https://github.com/civctp2/civctp2/pull/147#issuecomment-502403105, I still observe the non-initialized value for every loaded save-game. I wonder why that happens despite the initialization in https://github.com/civctp2/civctp2/pull/147#issuecomment-502403105.
Any way, this might make it worth to save `m_goldFromTransitRoutes` since it can only be re-evalutated at the end of a turn. Any idea where this could be squeezed in?

Issue #140, opened 2019-05-19
List trade route civs for tile


LynxAbraxas:
At least with SDL it is not possible to see what civs have a trade route at some tiles where many trade routes run along. There is only one color even if there are trade routes from other civs and following them on the map can become very cumbersome if there are many.
It would be great to have some option to list the trade routes of a tile, like the units on a tile. However, the space where units are listed is often already over full, so perhaps the trade route listing should go in the tile-info.

LynxAbraxas:
Solved with #150.

Issue #148, opened 2019-06-15
Introduce bonus for the receiver of a trade route


LynxAbraxas:
So far only the civ that sells a good gets something (gold) but not the civ that receives the good. 
Eventhough a message is issued to the receiver if a trade route got cancelled by the sender
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/TradeRoute.cpp#L53-L64
this seems to have no negative effect to the receiver.

Realistic would be that the receiver gets the trade good and pays for that in trun. However the gold that the seller gets
https://github.com/civctp2/civctp2/blob/3a375905777f3d902fdbfbe5b2437fb573234555/ctp2_code/gs/gameobj/CityData.cpp#L4442
is not removed from the receiver.

Types of bonuses the receiver could get:

- gold bonus (which would be illogic, as the receiver should actually have to pay)
- PW/production (coal, iron, oil, glass, ...), food (wheat, rice, crap, fish, mushrooms, banana, dates, ...), happiness (tea, coffe, dye, poppies, chilli, grapes, herbs, sugar, perls, emeralds, elephants, camels), perhaps even some reduce pollution (copper, gold, uran?)
- strategic benefits, like oil to be able to built e.g. tanks

In regard to what could possibly be implemented most easily would probably be to just assign the additional food, production, commerce values of the good received or add the good to the pool of tradable goods.
This way it would be possible to built cities in the mountains or on glaciers where hardly any food is found locally but that could be supplied by trade.

Regarding strategic benefits, we would need to introduce strategic meaning to goods, possibly bound to advances revealing goods, goods being used up after some time (see #142).
All this would also demand that the AI is "taught" to make use of this as well in order to avoid that this only gives benefits to a human player, thereby indirectly weakening the AI.

Any way, some bonus for the receiver would make the trading concept of ctp2 much more interesting.

LynxAbraxas:
@MartinGuehmann any ideas on how to realize "assign the additional food, production, commerce values of the good received"? How much work do you recon that would be?

MartinGuehmann:
> @MartinGuehmann any ideas on how to realize "assign the additional food, production, commerce values of the good received"? How much work do you recon that would be?

I think that is quite easy, the question is just how to do it and what are the boni, would it be the field boni that are then transferred to the other city?

Basically you have to add the stuff when the city collects the food, gold and production.

LynxAbraxas:
[...]

PR #150, opened 2019-06-16
List trade routes in tile info window

LynxAbraxas:
PR in regard to #140, which lists the trade routes of a tile in the tile info window. This seems to be the best place because the length of the list varies and likely does not fit in any case into the status bar info line. Possibly a drop down list could be used but would need an extra click to reveal the info, other suggestions and help welcome.


LynxAbraxas:
Even though this basically seems to work, I never get more than 2 trade routes listed even if a tile (must) have more than that.

1 route listed for the tile indicated with the active tile frame (below the mouse cursor) eventhough the graphics show a split, so at least 2 routes should be there:

ss_2019-06-16_20:19:39

After the split one for each tile (correct):

ss_2019-06-16_20:20:11

ss_2019-06-16_20:19:46


2 listed but there should be at least 3:

ss_2019-06-16_20:19:19

My current guess is that there might be something wrong in
https://github.com/civctp2/civctp2/blob/06f6af9e2761a265c83dba48a84709cde44fcc81/ctp2_code/gs/world/Cell.cpp#L698-L709

Oddly, even with https://github.com/civctp2/civctp2/pull/150/commits/30ef9631f3a4d3eac3bfe6dfab64f8014716a45b I do not get a segfault or other crash so I wonder if something is wrong with
https://github.com/civctp2/civctp2/blob/06f6af9e2761a265c83dba48a84709cde44fcc81/ctp2_code/gs/gameobj/ObjPool.cpp#L109-L118
which would have a very wide impact.
@MartinGuehmann any thoughts?

MartinGuehmann:
You can check a few thinks:

For debugging you can add to the text the number of trade routes first, so that you know whether it is a problem there. However, I don't think it is there.

Then you could concatenate the names with some visible character then with a white space to see if some problem is there.

I also had some problem with the std::string, so you could a plain string for debugging, but again it doesn't seem to be the problem.

And finally you have the field that displays the country names, that also has a width of 100 pixels. Of course, that depends on the implementation, but if you put in a string that is longer than 100 pixels then it may be cut off.

LynxAbraxas:
Thanks @MartinGuehmann, the string concatenation seems to be the problem:

ss_2019-06-16_21:17:23

Now, what is going wrong with the concatenation?

MartinGuehmann:
> Thanks @MartinGuehmann, the string concatenation seems to be the problem:
> 
> ss_2019-06-16_21:17:23
> 
> Now, what could be going wrong with the concatenation?

In fact, at that point I would go to the debugger and see what is happening inside. Of course, without a debugger you could do some std::cerr, and see that the trade routes have a civ associated with. And whether, the string is concatenated.

By the way, what is up with 1977f37? I would be careful with it, since links may be handled differently on different file systems.

LynxAbraxas:
> By the way, what is up with 1977f37? I would be careful with it, since links may be handled differently on different file systems.

True, but as far as I know git takes care of that. Any other idea how to avoid such ugly duplication of exact the same code?

LynxAbraxas:
Hm, getting all on stderr:
```
ui/interface/helptile.cpp L301: Persian
ui/interface/helptile.cpp L301: Mexican
ui/interface/helptile.cpp L301: Persian
ui/interface/helptile.cpp L301: Carthaginian
```
Cannot see what could be wrong with the `.append()`. It makes the impression as if can only append two strings but no more, which should not be the case (https://stackoverflow.com/questions/662918/how-do-i-concatenate-multiple-c-strings-on-one-line).

LynxAbraxas:
Apparently the label object for displaying the text cuts off strings exceeding its width at spaces... https://github.com/civctp2/civctp2/pull/150/commits/743e3254e9de14f8a3e08c5831924aba63567e0a solves the issue:

ss_2019-06-16_23:01:28

LynxAbraxas:
@MartinGuehmann do you know of a uniq (sort) fuction in the code base that could be used to list civs only once in the string?

MartinGuehmann:
> > By the way, what is up with [1977f37](https://github.com/civctp2/civctp2/commit/1977f373613701aab565a50e36e7186630a65c67)? I would be careful with it, since links may be handled differently on different file systems.
> 
> True, but as far as I know git takes care of that. Any other idea how to avoid such ugly duplication of exact the same code?

Actually, the stuff is supposed to go into the default folder, but for some reason every language has its own set. I think there are little differences between some of the files, but that was too tiny to notice and we had not all the files of the different languages.

So the idea is that you copy a file to the language folder if you need to adjust the width of such a text field if it usually needs to bigger in that language.

I think that is the way it was intended, however I cannot undo that, because the game looks first into the language folder before it goes to default. So in principle, we would need an installer that removes the copies from the language folders, but that is too much work.

So if Git cares about it than we could just leave it like this. And actually also change also all the other files this way.

MartinGuehmann:
> @MartinGuehmann do you know of a uniq (sort) fuction in the code base that could be used to list civs only once in the string?

The easiest way, would be an array or if you want to have it a bit more comfortable a vector or a list. Then loop over it before you insert. The thing is that we don't expect here hundreds of trade routes. So that could also be fine. Another way would be using a data structure that only allows you to insert a value only once, or has an is_contained functions that works in O(1). Probably, the standard library has a data structure like this.

The other thing is, whether you want to list only the civs and not also the source city, target city, and good. That would be a bit more interesting. But for that you have to add a list box, and extend the tile window. And if you think that information is a bit too much, well you know traders, they can't keep their mouth shut. :D

LynxAbraxas:
Turns out that this is quite easy:

ss_2019-06-20_22:56:58

and even suitable for tiles where many civs have route:

ss_2019-06-20_22:56:37

still, there is a limit.

>The other thing is, whether you want to list only the civs and not also the source city, target city, and good. That would be a bit more interesting. But for that you have to add a list box, and extend the tile window. 

The info is mostly interesting for pirating and for checking what bootle-necks are most interesting and if there is a route from a civ you agreed on to not pirate trade. The good, source and target city is not much important, at least as long #148 is not realized.
Since a drop-box (needs extra click) with a list or extending the window and adding a list-box (no extra click) is quite a bit more challenging, I'd go for the current solution so far.

MartinGuehmann:
> Since a drop-box (needs extra click) with a list or extending the window and adding a list-box (no extra click) is quite a bit more challenging, I'd go for the current solution so far.

Well, I prefer the listbox, but we can add it. Could you still add an issue, so that it doesn't get lost?

PR #154, opened 2019-06-22
Debug breaking no-piracy treaty

LynxAbraxas:
PR to resolve #152.

Issue #152, opened 2019-06-20
Breaking no-piracy treaty not possible


LynxAbraxas:
It seems that currently it is not possible to break a no-piracy treaty, not even by declaring war. @MartinGuehmann could it be that this was broken by the remedy to #75: #127?

I get the message that ask me if I want to break the treaty with the Incans, but clicking yes has no effect:

![ss_2019-06-20_23:50:11](https://user-images.githubusercontent.com/21012234/59884684-5ec96300-93b9-11e9-99e3-899c7a7cbe11.png)

The order to pirate comes here in case of a no-piracy treaty
https://github.com/civctp2/civctp2/blob/93ca09d6e24d6b4a8f958da2e16b0a1fdd6af75c/ctp2_code/gs/gameobj/ArmyData.cpp#L5911-L5927
which then leads to here
https://github.com/civctp2/civctp2/blob/93ca09d6e24d6b4a8f958da2e16b0a1fdd6af75c/ctp2_data/default/gamedata/script.slc#L1465
where I wonder if the remedy in #127 now prevents that order to be executed. However, since the order is not bound to a change in position, this should not be the case.
Any idea what could be the problem?

MartinGuehmann:
> It seems that currently it is not possible to break a no-piracy treaty, not even by declaring war. @MartinGuehmann could it be that this was broken by the remedy to #75: #127?
> 
> I get the message that ask me if I want to break the treaty with the Incans, but clicking yes has no effect:
> 
> ![ss_2019-06-20_23:50:11](https://user-images.githubusercontent.com/21012234/59884684-5ec96300-93b9-11e9-99e3-899c7a7cbe11.png)

So whose route are you trying to pirate? One from the Incas? The message doesn't tell me that it is from the Incas it just says "that nation".

> The order to pirate comes here in case of a no-piracy treaty
> https://github.com/civctp2/civctp2/blob/93ca09d6e24d6b4a8f958da2e16b0a1fdd6af75c/ctp2_code/gs/gameobj/ArmyData.cpp#L5911-L5927
> 
> 
> which then leads to here

Well, important is also the line before, which should break the treaty.

https://github.com/civctp2/civctp2/blob/93ca09d6e24d6b4a8f958da2e16b0a1fdd6af75c/ctp2_data/default/gamedata/script.slc#L1464-L1465
> 
> 
> where I wonder if the remedy in #127 now prevents that order to be executed. However, since the order is not bound to a change in position, this should not be the case.
[...]

PR #156, opened 2019-06-23
Re-introduce turnBuf which was removed in ec67297c1

LynxAbraxas:
The commits of this PR would re-introduce `turnBuf` (which was removed in ec67297c1 without any comment) for asking to break a no-piracy treaty, and once were part of #154.
However the game crashes when pirating with these changes, nor does this PR seem to help solving #152.


LynxAbraxas:
Closing, due to the above mentioned reasons. PR merely created to keep track of having checked these changes and then discarding them.

PR #159, opened 2019-07-03
Change the way trade routes are drawn on the map

LynxAbraxas:
This PR consists of commits that change the way trade routes are drawn on the map. So far a trade route has a black "outline":

ss_2019-07-02_23:30:08

together with the line anti-aliasing it is hard in some cases to recognize the color of the owner.

https://github.com/civctp2/civctp2/commit/64ac72f4fe3af92b6b6fb6d74f97f11cc52c4c5b replaces the "outline" color black with the color of the player (which is already used for the main line):

ss_2019-07-02_23:27:01

https://github.com/civctp2/civctp2/commit/f7bb70f13caf392d7816566773b05a8d9352aed5 removes the outline completely, leading to thinner trade route lines, in a way more elegant but due to the anti-aliasing also less well visible:

ss_2019-07-02_23:22:46

Currently, I'm not sure which version is best.

### DIFF
--- a/ctp2_code/gfx/layers/citylayer.cpp
+++ b/ctp2_code/gfx/layers/citylayer.cpp
@@ -143,6 +143,23 @@ void DrawTradeRouteSegment(aui_Surface *surf, MapPoint &pos, WORLD_DIRECTION dir
 	if (x1 >= 0 && y1 >= 0 && x2 >= 0 && y2 >= 0 &&
 		x1 < surf->Width() && y1 < surf->Height() && x2 < surf->Width() && y2 < surf->Height()) {
 
+		primitives_DrawDashedLine16(surf,x1,y1,x2,y2,route, k_TRADE_DASH_LEN);
+	}
+
+	if (x1==x2) {
+		x1++;
+		x2++;
+	} else {
+		y1++;
+		y2++;
+	}
+
+	tempRect.right++;
+	tempRect.bottom++;
+
+	if (x1 >= 0 && y1 >= 0 && x2 >= 0 && y2 >= 0 &&
+		x1 < surf->Width() && y1 < surf->Height() && x2 < surf->Width() && y2 < surf->Height()) {
+
 		primitives_DrawDashedLine16(surf,x1,y1,x2,y2,outline, k_TRADE_DASH_LEN);
 	}
 
--- a/ctp2_code/gfx/layers/citylayer.cpp
+++ b/ctp2_code/gfx/layers/citylayer.cpp
@@ -140,11 +140,13 @@ void DrawTradeRouteSegment(aui_Surface *surf, MapPoint &pos, WORLD_DIRECTION dir
 	tempRect.right++;
 	tempRect.bottom++;
 
+	/* do not draw an "outline" for a trade route, this looks iritating and makes it difficult to recognize the main color
 	if (x1 >= 0 && y1 >= 0 && x2 >= 0 && y2 >= 0 &&
 		x1 < surf->Width() && y1 < surf->Height() && x2 < surf->Width() && y2 < surf->Height()) {
 
 		primitives_DrawDashedAALine16(surf,x1,y1,x2,y2,outline, k_TRADE_DASH_LEN);
 	}
+	*/
 
 	g_tiledMap->AddDirtyRectToMix(tempRect);
 
--- a/ctp2_code/gfx/layers/citylayer.cpp
+++ b/ctp2_code/gfx/layers/citylayer.cpp
@@ -140,13 +140,11 @@ void DrawTradeRouteSegment(aui_Surface *surf, MapPoint &pos, WORLD_DIRECTION dir
 	tempRect.right++;
 	tempRect.bottom++;
 
-	/* do not draw an "outline" for a trade route, this looks iritating and makes it difficult to recognize the main color
 	if (x1 >= 0 && y1 >= 0 && x2 >= 0 && y2 >= 0 &&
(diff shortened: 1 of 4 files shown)


LynxAbraxas:
Oddly, even without the drawing of the outline, routes going north-west/south-east (see e.g. the one leaving Pachacamac) still have some additional color (dark blue) around them, kind of like a shadow.

MartinGuehmann:
If you wanna change it then I prefer a thicker line. The point of the black outline is to make the trade route more visible. In principle, the extra black line shouldn't make the player color less visible. So the problem could be that the black line removes some of the player color.

LynxAbraxas:
I prefer the thicker line too. Just checking how things look without anti-aliasing.
Any idea what this code might have been meant for (was disabled already in the root commit):
https://github.com/civctp2/civctp2/blob/42c109ad024828b8b57ca5f98135f27e46640c14/ctp2_code/gs/gameobj/TradePool.cpp#L97-L117

LynxAbraxas:
> perhaps an error in the anti-aliasing code.

This seems odd:
https://github.com/civctp2/civctp2/blob/42c109ad024828b8b57ca5f98135f27e46640c14/ctp2_code/ui/aui_utils/primitives.cpp#L3184
Does that replace the apparent function by a constant value of 4? Is that the range of the anti-aliasing?

MartinGuehmann:
> I prefer the thicker line too. Just checking how things look without anti-aliasing.
> Any idea what this code might have been meant for (was disabled already in the root commit):
> https://github.com/civctp2/civctp2/blob/42c109ad024828b8b57ca5f98135f27e46640c14/ctp2_code/gs/gameobj/TradePool.cpp#L97-L117

That seems to be some code that highlights a trade route if it has been selected. I think this never made it even into CTP1.

LynxAbraxas:
> That seems to be some code that highlights a trade route if it has been selected.

Yeah, that's what I thought too, just cannot find a way to select a trade route, neither by clicking nor with the trade manager, although `SetPathSelectionState` is used in some places.

MartinGuehmann:
> > perhaps an error in the anti-aliasing code.
> 
> This seems odd:
> https://github.com/civctp2/civctp2/blob/42c109ad024828b8b57ca5f98135f27e46640c14/ctp2_code/ui/aui_utils/primitives.cpp#L3184
> 
> 
> Does that replace the apparent function by a constant value of 4? Is that the range of the anti-aliasing?

I would call it a function style makro. Basically, COVERAGE(dist) would be replaced by the precompiler by some code, which follows. In this case it is 4. Maybe here some test code that was supposed to be replaced by something more sophisticated.

So in the end the program is compiled as you would have 4 at the place of COVERAGE(dist) or COVERAGE(jjshh@#dä), or whatever in the brackets. Actually, the Assert makro is a better example for it.

MartinGuehmann:
> > That seems to be some code that highlights a trade route if it has been selected.
> 
> Yeah, that's what I thought too, just cannot find a way to select a trade route, neither by clicking nor with the trade manager, although `SetPathSelectionState` is used in some places.

Well, there is nothing like this in place, you can still select the goods on the map.

LynxAbraxas:
> you can still select the goods on the map.

True, does that have a use? If you select the elephant good, you even get a sound.

MartinGuehmann:
> > you can still select the goods on the map.
> 
> True, does that have a use? If you select the elephant good, you even get a sound.

Not that I know. Actually, they get deselected shortly afterwards.

LynxAbraxas:
No AA, no dashes, https://github.com/civctp2/civctp2/pull/159/commits/b6843a9ffae8902f5e666bb1612c56a695ff92e3:

ss_2019-07-03_22:34:55

No AA, dashed, no outline, https://github.com/civctp2/civctp2/pull/159/commits/d0e4578767c1885c1e2601244ac999d8c6a49962:

ss_2019-07-03_23:36:17

No AA, dashed, outline color same as line (i.e. thick lines), https://github.com/civctp2/civctp2/pull/159/commits/833724f763796026df1e0b772b2dd9e944c29ed3

ss_2019-07-04_00:24:42

No AA, dashed, outline color black, i.e. like originally just no anit-aliasing, https://github.com/civctp2/civctp2/pull/159/commits/5f6bfef4c4176ab37aceac7c2af2f0cec

ss_2019-07-04_00:47:13

Oddly, it seem that the outline does not get blended in fog-of-war (in contrast to the main line, no clue why) which makes the routes of https://github.com/civctp2/civctp2/pull/159/commits/833724f763796026df1e0b772b2dd9e944c29ed3 brighter in FOW but those of https://github.com/civctp2/civctp2/pull/159/commits/5f6bfef4c4176ab37aceac7c2af2f0cec are better visible on varying land tiles, especially along rail-roads.

LynxAbraxas:
OK, since the origina AA made the lines even thicker than 2 pixel, I created another commit that draws the route with a line of 2 pixel width in the civ color and a 1 pixel width outline in black, https://github.com/civctp2/civctp2/pull/159/commits/a68e1465c392c54faaf6344bec617ac87028a02c

ss_2019-07-04_18:59:32

LynxAbraxas:
Final tweaks, shorter dash length and a small trick to have the dashes distributed more evenly:
ss_2019-07-04_19:55:27

MartinGuehmann:
> Final tweaks, shorter dash length and a small trick to have the dashes distributed more evenly:
> ss_2019-07-04_19:55:27

So, I guess you are done then?

LynxAbraxas:
If you agree, then yes it can go in (sorry, forgot the WIP at the beginning).

MartinGuehmann:
> If you agree, then yes it can go in (sorry, forgot the WIP at the beginning).

Sure, it can go in. Then let's do it.

PR #135, opened 2019-05-06
Show the prerequisites for the current research goal in the tech tree of the Great Library

MartinGuehmann:
Show the prerequisites for the current research goal in the tech tree of the Great Library: This way it is easier to see what needs to be researched for the current research goal.

This PR should still add a label to the tech tree giving the current research and how many prerequisites it has has and how many of those are known.

PR #160, opened 2019-07-04
Fix a crash after loading when opening the Great Library and tech goal was set

MartinGuehmann:
This fixes a bug introduced in #135

..\ctp2_code\ui\aui_ctp2\chart.cpp
..\ctp2_code\ui\interface\sci_advancescreen.cpp
..\ctp2_code\ui\interface\sci_advancescreen.h


LynxAbraxas:
Preparing for a test, will report back.

LynxAbraxas:
Going directly into the GL works again! Many thanks @MartinGuehmann for fixing this.

MartinGuehmann:
So then this can go in.

Issue #162, opened 2019-07-06
Message for the owner in case an agreement or treaty is broken


LynxAbraxas:
While breaking an agreement or treaty yields a message to the violator, no message is sent to the owner. In case of no-piracy agreements (see e.g. #161) the owner of the agreement can only notice that is was broken by checking if an agrement-icon was removed in the diplomacy manager. Remebering the diplomacy state is quite hard for a human, so a message in case an agreement or treaty was broken would be very helpful.

It seems this was the case for CTP1
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/gs/gameobj/AgreementData.cpp#L1175-L1205
using the non-`_By_You` messages like
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_data/default/gamedata/script.slc#L1643-L1646
https://github.com/civctp2/civctp2/blob/504e67b40fda7ef96112d1ac9fa77b3466044d33/ctp2_data/english/gamedata/info_str.txt#L81

It would be nice to re-activate these messages (creating new ones for agreements that were not in CTP1). This could be acomplished by adding the `so2` from
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/gs/gameobj/AgreementData.cpp#L1178-L1198
to
https://github.com/civctp2/civctp2/blob/504e67b40fda7ef96112d1ac9fa77b3466044d33/ctp2_code/ai/diplomacy/diplomat.cpp#L1125-L1132

This would need messages for all proposal types handled by
https://github.com/civctp2/civctp2/blob/504e67b40fda7ef96112d1ac9fa77b3466044d33/ctp2_code/ai/diplomacy/diplomat.cpp#L1000-L1117

However, such an effort would only make much sense if the AI actually can break agreements and treaties. @MartinGuehmann do you know if the AI can do so in general? And in what cases not?

LynxAbraxas:
I have the impression that the missing CTP1 message for `004NoPiracyBrokenByYou` (`NO_PIRACY_BROKEN_BY_YOU`) corresponding to
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_data/default/gamedata/script.slc#L1643-L1646
[...]

PR #161, opened 2019-07-06
Ask if player wants to break no-piracy agreement

LynxAbraxas:
In contrast to #154, this PR solves #152 such that the player is asked if a no-piracy agreement should be broken (#157). 
This PR makes use of the CTP1 question-message but uses CTP2 code for breaking the agreement (https://github.com/civctp2/civctp2/commit/5f4529c3da7587642c7d9184bcd0c55269e17cf8).
In addition, this PR solves the bug, that the agreement was interpretet the wrong way round (https://github.com/civctp2/civctp2/commit/98e52512c3e1bbb4fa91ad27965a566aea55b02b), i.e. not as an offer but as a request.

So it is no possible in the example below, where the Incans agreed upon a no-piracy request by the Persians, that the Persians pirate an Incan route
ss_2019-07-06_14:14:15

since there is no agreement involved:
ss_2019-07-06_14:13:22
and this changes nothing in the agreements between the two.

Becoming the Incans, who have agreed upon no-piracy with the Persians
ss_2019-07-06_14:28:43

and pirating a Persian route, pops up the question message
ss_2019-07-06_14:30:23

clicking "yes" then executes pirating and leads to the CTP2 trust-loss message and the agreement is cancelled
ss_2019-07-06_14:35:28

Issue #157, opened 2019-06-26
Question message if action would result in the violation of an agreement


LynxAbraxas:
In the scope of #154, it was noticed that there is CTP1 code that shows the player a question message if an action would result in the violation of a standing agreement, like no-piracy or leave-our-lands. This seems not to be implemented for CTP2 agreements but would be quite helpful, especially between longer gaming breaks.

Ideally such a question message should e.g. be sent to a player if a unit/army move would violate a withdraw troops agreement. It seems the CTP2 way is to expect that a player remembers this (or was not implemented due to lack of time/programmers). We could possibly implement this for CTP2 with the CTP1 messages, i.e. moving
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/gs/gameobj/AgreementData.cpp#L1099-L1206
to e.g.
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/ai/diplomacy/agreementmatrix.cpp#L335-L368
which so far only handles `PROPOSAL_TREATY*` but not `PROPOSAL_REQUEST*` or `PROPOSAL_OFFER*` of
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/ai/diplomacy/diplomattypes.h#L93-L151

LynxAbraxas:
Another way could be to replace
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/gs/slic/slicfunc.cpp#L4542-L4556
with a call to `StopPiracyRegardEvent`
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/ai/diplomacy/regardevent.cpp#L246-L264
@MartinGuehmann do you know how to issue events like `StopPiracyRegardEvent`?
The only place I could find it so far is when set as a call-back:
https://github.com/civctp2/civctp2/blob/fd65451326ffd76e46e0c2ed5015fb7876d5af0e/ctp2_code/ai/diplomacy/regardevent.cpp#L1186-L1188

LynxAbraxas:
Another thought: 
@MartinGuehmann do you know if there is a way to return from
[...]

PR #164, opened 2019-07-06
Ask if player wants to break a standing trade pact

LynxAbraxas:
In regard to #157, based on #161 (which should be merged first), with this PR a question-message (with a general (so far unused) text from CTP1) is shown that asks if the player wants to break a standing trade pact by pirating a route belonging to the pact partner
ss_2019-07-06_18:32:58

Clicking "yes" executes the pirating which breaks the pact
ss_2019-07-06_18:33:43

A message for the partner about the fact that the pact got broken would be in the scope of #162.

### DIFF
--- a/ctp2_code/ai/diplomacy/diplomat.cpp
+++ b/ctp2_code/ai/diplomacy/diplomat.cpp
@@ -1445,9 +1445,9 @@ void Diplomat::EnactStopPiracy(const PLAYER_INDEX victimId, const PLAYER_INDEX p
 			if (army.IsValid() && (army.GetOwner() == pirateId))
 			{
 				g_gevManager->AddEvent
-				    (GEV_INSERT_Tail, GEV_SetPiratingArmy,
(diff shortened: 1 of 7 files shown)


MartinGuehmann:
I guess that is now also complete.

LynxAbraxas:
Yes, tested for a while and seems to work all right.

MartinGuehmann:
> Yes, tested for a while and seems to work all right.

OK, then let's merge.

LynxAbraxas:
Hm, just observed that the AI seems to have some trouble with this. If it has agreed to a no-piracy agreement, it keeps trying to pirate the route, it gets an invalid-action sign but keeps trying turn after turn.
@MartinGuehmann any idea how to counter act that AI behaviour?

MartinGuehmann:
I guess this has to be handled by Goal::IsInvalidByDiplomacy(), you should be able to find all the pieces of code you need in functions that are called by Goal::IsTotallyComplete().

Issue #158, opened 2019-07-02
City-Icon for conversion, franchies etc should update even for cities in fog-of-war


LynxAbraxas:
Currently, at least with SDL, the icon of a city indicating conversion or franchise does not update if the city is in fog-of-war. However, since in any case some kind of income occurs per round, such an icon should vanish (for the receiving civ) if the conversion or franchise is removed even if that city is in fog-of-war as the income does not occur any more.

MartinGuehmann:
Actually, if you want to update the icon, even in fog of war then you should also send a message to the franchise owner if the franchise is cleared.

LynxAbraxas:
I though of trying to do less message spamming and just removing the icon. Any idea for where to start?

LynxAbraxas:
This can be closed since #166 is merged and #171 is on the way.

PR #166, opened 2019-07-07
Converted and franchise city icons updated independent of FOW

LynxAbraxas:
because they indicate a state of extra income and a stop of income would reveal this.
PR in relation to #158.

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -300,6 +300,7 @@
 #include "UnitPool.h"                   // g_theUnitPool
 #include "UnitRecord.h"
 #include "unitutil.h"
+#include "UnseenCell.h"
 #include "WonderRecord.h"
 #include "WonderTracker.h"
 #include "wonderutil.h"
@@ -5977,6 +5978,16 @@ void CityData::RemoveFranchise()
 {
 	if(m_franchise_owner < 0) // according to CityData::Unconvert
 		return;
+
+	//// let the owner of the franchise know about its removal, see https://github.com/civctp2/civctp2/pull/166
+	MapPoint pos; // needed to get un-seen cell at this city pos
+	UnseenCellCarton ucell;
+
+	m_home_city.GetPos(pos);
+	if(g_player[m_franchise_owner]->GetLastSeen(pos, ucell)) // get un-seen cell of franchise owner
+	    {
+	    ucell.m_unseenCell->SetIsFranchised(false); // remove franchise flag
+	    }
 
 	m_franchise_owner = -1;
 	m_franchiseTurnsRemaining = -1;
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -4912,8 +4912,7 @@ void CityData::DoTurnCounters()
 	}
 	else if(m_franchiseTurnsRemaining == 0) // only zero is meant to reset a Franchise (owner), positive numbers define turns until Franchise is removed, -1 is the default for infinit Franchise, see https://github.com/civctp2/civctp2/pull/167
 	{
-		m_franchise_owner = -1; // owner == -1 means no Franchise
-		m_franchiseTurnsRemaining = -1;
+		RemoveFranchise();
 	}
 }
 
@@ -5974,6 +5973,15 @@ void CityData::MakeFranchise(sint32 player)
 	m_franchiseTurnsRemaining = -1;
 }
 
+void CityData::RemoveFranchise()
+{
+	if(m_franchise_owner < 0) // according to CityData::Unconvert
+		return;
+
+	m_franchise_owner = -1;
+	m_franchiseTurnsRemaining = -1;
+}
+
 sint32 CityData::GetFranchiseTurnsRemaining() const
 {
 	return m_franchiseTurnsRemaining;
@@ -5984,8 +5992,7 @@ void CityData::SetFranchiseTurnsRemaining(sint32 turns)
 	m_franchiseTurnsRemaining = turns;
 	if(turns < 1)
 	{
-		m_franchiseTurnsRemaining = -1;
-		m_franchise_owner = -1;
+		RemoveFranchise();
 	}
 }
 
@@ -6598,8 +6605,7 @@ void CityData::ResetCityOwner(sint32 owner)
 	NewGovernment(g_player[m_owner]->m_government_type);
 
 	m_walls_nullified = false;
-	m_franchiseTurnsRemaining = 0;
-	m_franchise_owner = -1;
+	RemoveFranchise(); // Franchise removed in contrast to conversion, see below
 	m_watchfulTurns = 0;
 	m_bioInfectionTurns = 0;
 	m_bioInfectedBy = -1;
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -6357,6 +6357,16 @@ void CityData::Unconvert(bool makeUnhappy)
 	if(m_convertedTo < 0)
 		return;
 
+	//// let the owner of the conversion know about its removal, see https://github.com/civctp2/civctp2/pull/166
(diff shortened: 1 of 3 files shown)


LynxAbraxas:
Despite the issue of https://github.com/civctp2/civctp2/pull/103#issuecomment-509030084, this seems to work fine for conversion icons:

Without this PR
ss_2019-07-07_23:26:52

with this PR
ss_2019-07-07_23:25:04

With this PR, before Bonanza is reformed
ss_2019-07-07_23:10:19

after reformation despite being in FOW
ss_2019-07-07_23:11:09

LynxAbraxas:
@MartinGuehmann are there any other kind of incomes from other actions like conversions and franchise that should be included in the PR?

MartinGuehmann:
> @MartinGuehmann are there any other kind of incomes from other actions like conversions and franchise that should be included in the PR?

Not that I know.

However, should we tell everybody that the franchise is gone? I mean this is only something the city owner and the franchise owner can notice.

And by the way, the city can be destroyed, so the place wouldn't contain a city anymore.

LynxAbraxas:
> However, should we tell everybody that the franchise is gone? I mean this is only something the city owner and the franchise owner can notice.

Thanks @MartinGuehmann, that's a very good point. I like that idea, but how to realize that in the code? Do you know of any other graphics that is treated differently concerning attacker and victim in contrast to 3rd?

> And by the way, the city can be destroyed, so the place wouldn't contain a city anymore.

Yes, another good point I did not think of directly, but that might be sufficiently covered by 
https://github.com/civctp2/civctp2/pull/166/commits/9af9d9cde95511814b324a3b4d42a12b781b8c15#diff-7d7cb1aaf6318518fcf8bfee0f699dcaR3440
or can you foresee any problem with that the way it is handled so far?

Did you have any success in ridding a city of a franchise with a Lawyer?

MartinGuehmann:
> > However, should we tell everybody that the franchise is gone? I mean this is only something the city owner and the franchise owner can notice.
> 
> Thanks @MartinGuehmann, that's a very good point. I like that idea, but how to realize that in the code?

Well, when the franchise is removed, you have to access the unseen cell of the franchise owner and then remove the franchise. That shouldn't be much.

> Do you know of any other graphics that is treated differently concerning attacker and victim in contrast to 3rd?

Not that I know of. I guess we'll figure out. :D

> > And by the way, the city can be destroyed, so the place wouldn't contain a city anymore.
> 
> Yes, another good point I did not think of directly, but that might be sufficiently covered by
> [9af9d9c#diff-7d7cb1aaf6318518fcf8bfee0f699dcaR3440](https://github.com/civctp2/civctp2/commit/9af9d9cde95511814b324a3b4d42a12b781b8c15#diff-7d7cb1aaf6318518fcf8bfee0f699dcaR3440)
> or can you foresee any problem with that the way it is handled so far?
> 
> Did you have any success in ridding a city of a franchise with a Lawyer?

Well, I have other things to do, so I haven't tried, sorry.

LynxAbraxas:
So with #167, it is now possible to remove a Franchise, which then also leads to the intended effect that the Franchise icon is not shown any more. This also works with #168.

So what is left to do here is to limit the removal of the icons in FOW only to the owner of the Conversion or Franchise.

> Yes, another good point I did not think of directly, but that might be sufficiently covered by
9af9d9c#diff-7d7cb1aaf6318518fcf8bfee0f699dcaR3440
or can you foresee any problem with that the way it is handled so far?

@MartinGuehmann do you foresee any problems there or is 9af9d9c#diff-7d7cb1aaf6318518fcf8bfee0f699dcaR3440 sufficient?

ctp2_code/gs/gameobj/CityData.cpp
```diff
@@ -6598,8 +6605,7 @@ void CityData::ResetCityOwner(sint32 owner)
 	NewGovernment(g_player[m_owner]->m_government_type);
 
 	m_walls_nullified = false;
-	m_franchiseTurnsRemaining = 0;
-	m_franchise_owner = -1;
+	RemoveFranchise(); // Franchise removed in contrast to conversion, see below
```
LynxAbraxas: @MartinGuehmann do you see a problem here effectively replacing `m_franchiseTurnsRemaining = 0;` with `m_franchiseTurnsRemaining = -1;`?

ctp2_code/gs/gameobj/CityData.cpp
```diff
@@ -5974,6 +5973,15 @@ void CityData::MakeFranchise(sint32 player)
 	m_franchiseTurnsRemaining = -1;
 }
 
+void CityData::RemoveFranchise()
+{
+	if(m_franchise_owner < 0) // according to CityData::Unconvert
```
LynxAbraxas: @MartinGuehmann  is it critical here to return if `m_franchise_owner < 0`?

LynxAbraxas:
> > > However, should we tell everybody that the franchise is gone? I mean this is only something the city owner and the franchise owner can notice.
> > 
> > 
> > Thanks @MartinGuehmann, that's a very good point. I like that idea, but how to realize that in the code?
> 
> Well, when the franchise is removed, you have to access the unseen cell of the franchise owner and then remove the franchise. That shouldn't be much.

@MartinGuehmann  Is https://github.com/civctp2/civctp2/pull/166/commits/4ff90b493edb9ca6c1d06c3a095b6cdf5d16c90d doing things according to your suggestion?
https://github.com/civctp2/civctp2/pull/166/commits/aafa19dac98547db77862695bba58571b26361e4 does the same for conversions.
`RemoveFranchise` from https://github.com/civctp2/civctp2/pull/166/commits/6283ed8472666e696946dcdaecf460f7a3dfce9e is similar to
https://github.com/civctp2/civctp2/blob/aafa19dac98547db77862695bba58571b26361e4/ctp2_code/gs/gameobj/CityData.cpp#L6353
but its use does not exactly fit the former code, see in-line comments.

ctp2_code/gs/gameobj/CityData.cpp
MartinGuehmann: Well, I would assume with our current changes it should also be -1. Simply, this resets the city to default values.

But actually what is RemoveFranchise() doing? Does it set m_franchiseTurnsRemaining to -1?

LynxAbraxas: Yes, see https://github.com/civctp2/civctp2/pull/166/commits/6283ed8472666e696946dcdaecf460f7a3dfce9e#diff-647de8d8715871ce5f3623bf0a4f0158R5976

ctp2_code/gs/gameobj/CityData.cpp
MartinGuehmann: That depends, if whatever follows would be a lot of code then yes. However, here you just want to reset the franchise and if you want to do it cleanly then both members have to set to -1.

I think in the end it is a speed issue. So if the code is used very often, then you would optimize it in such a way, but then the function should be inline. However, I do not expect that this code is heavily used and so it can be just executed.

LynxAbraxas:
Tested, an seems to work as expected for removal of Conversions and Franchises.
Before the removal of a Conversion (of the Persians) of Bonanza
ss_2019-07-20_21:09:31
seen by the Italians
ss_2019-07-20_21:07:20
after Reformation seen by the Persians
ss_2019-07-20_21:13:07
but not by the Italians
ss_2019-07-20_21:14:32

LynxAbraxas:
Works similarly if a Franchise is removed. However this I could only test using the cheat-mode because the AI does not seem to remove Franchises so far (see https://github.com/civctp2/civctp2/pull/167#issuecomment-512394317).
Despite of that it is ready to be merged.

LynxAbraxas:
Game now crashes if the removal happens for a city not in FOW.

MartinGuehmann:
> Game now crashes if the removal happens for a city not in FOW.

That's because there is no unseen cell for cities not in FOW. So you have to check this and if there is no unseen cell you don't have to do anything.

LynxAbraxas:
Did so, currently testing this.

LynxAbraxas:
OK, now works as expected inside or outside FOW.

MartinGuehmann:
Looks good now, and thus can go in.

PR #167, opened 2019-07-09
Franchise not properly removed

LynxAbraxas:
Apparently (https://github.com/civctp2/civctp2/pull/103#issuecomment-509030084, https://github.com/civctp2/civctp2/pull/166#issuecomment-509384392) franchise are neither properly installed nor removed.
https://github.com/civctp2/civctp2/commit/f5a260df6b3561fe7f7246b0ad476e34cfd589b7 resolves proper removal.

### DIFF
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -2384,7 +2384,7 @@ ORDER_RESULT ArmyData::SueFranchise(const MapPoint &point)
 		return ORDER_RESULT_ILLEGAL;
 	}
 
-	if(cell->GetCity().GetFranchiseTurnsRemaining() <= 0) {
+	if(cell->GetCity().GetFranchiseTurnsRemaining() == 0) {
 
 		return ORDER_RESULT_ILLEGAL;
 	}
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -2389,6 +2389,11 @@ ORDER_RESULT ArmyData::SueFranchise(const MapPoint &point)
 		return ORDER_RESULT_ILLEGAL;
 	}
 
+	if(cell->GetCity().GetFranchiseTurnsRemaining() > 0 && cell->GetCity().GetFranchiseTurnsRemaining() <= g_theConstDB->Get(0)->GetTurnsToSueFranchise()) {
+
+		return ORDER_RESULT_ILLEGAL;
+	}
+
 	AddSpecialActionUsed(u);
 
 	g_gevManager->AddEvent(GEV_INSERT_AfterCurrent, GEV_RemoveFranchise,
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -4910,9 +4910,9 @@ void CityData::DoTurnCounters()
 	{
 		m_franchiseTurnsRemaining--;
 	}
-	else if(m_franchiseTurnsRemaining <= 0)
+	else if(m_franchiseTurnsRemaining == 0) // only zero is meant to reset a Franchise (owner), positive numbers define turns until Franchise is removed, -1 is the default for infinit Franchise, see https://github.com/civctp2/civctp2/pull/167
 	{
-		m_franchise_owner = -1;
+		m_franchise_owner = -1; // owner == -1 means no Franchise
 		m_franchiseTurnsRemaining = -1;
 	}
 }
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -4910,7 +4910,7 @@ void CityData::DoTurnCounters()
 	{
 		m_franchiseTurnsRemaining--;
 	}
-	else if(m_franchiseTurnsRemaining == 0)
+	else if(m_franchiseTurnsRemaining <= 0)
 	{
 		m_franchise_owner = -1;
 		m_franchiseTurnsRemaining = -1;
--- a/ctp2_code/gs/gameobj/armyevent.cpp
+++ b/ctp2_code/gs/gameobj/armyevent.cpp
@@ -1430,7 +1430,7 @@ STDEHANDLER(RemoveFranchiseEvent)
 
 	if(!args->GetArmy(0, a)) return GEV_HD_Continue;
 	if(!args->GetUnit(0, lawyer)) return GEV_HD_Continue;
-	if(!args->GetUnit(0, city)) return GEV_HD_Continue;
+	if(!args->GetCity(0, city)) return GEV_HD_Continue;
 
 	city.SetFranchiseTurnsRemaining(g_theConstDB->Get(0)->GetTurnsToSueFranchise());
 	return GEV_HD_Continue;


LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/167/commits/0f12a2ac2d7c0647445639627c11c3015c3debd3 it is now possible to do a Sue-Franchise-Attack with a Lawer
ss_2019-07-10_00:18:19
but so far the franchise is not removed, neither directly nor in the next turn.

LynxAbraxas:
Apparently, when sueing a Franchise, the game gets up to here:
https://github.com/civctp2/civctp2/blob/f726b014a742f640530a451f1c138b5e771a0bbc/ctp2_code/gs/gameobj/armyevent.cpp#L1433
but
https://github.com/civctp2/civctp2/blob/f726b014a742f640530a451f1c138b5e771a0bbc/ctp2_code/gs/gameobj/armyevent.cpp#L1435
does not get into
https://github.com/civctp2/civctp2/blob/f5a260df6b3561fe7f7246b0ad476e34cfd589b7/ctp2_code/gs/gameobj/CityData.cpp#L5982-L5990
which should stop the Franchise because currently there is not delay:
https://github.com/civctp2/civctp2/blob/0f12a2ac2d7c0647445639627c11c3015c3debd3/ctp2_data/default/gamedata/Const.txt#L195

@MartinGuehmann any idea what could be wrong? https://github.com/civctp2/civctp2/pull/167/commits/f726b014a742f640530a451f1c138b5e771a0bbc does not solve this, eventhough I though that this would be the bug since a city seems to be handled like a unit.

LynxAbraxas:
Nor does https://github.com/civctp2/civctp2/pull/167/commits/5e8c019f0bd3c71a19d40c43e0befe27377d7302 help, compiles, does not crash but now the game not even ends up in the `RemoveFranchiseEvent`.

MartinGuehmann:
> Nor does [5e8c019](https://github.com/civctp2/civctp2/commit/5e8c019f0bd3c71a19d40c43e0befe27377d7302) help, compiles, does not crash but now the game not even ends up in the `RemoveFranchiseEvent`.

Of course, it does not even go to the franchise event. Before the events are executed, they are checked whether you passed all the needed arguments. If they are missing the event is discarded. So, since you have removed the arguments, your event goes to the bin.

MartinGuehmann:
And another thing to add, the expected list of arguments for an event is defined elsewhere.

MartinGuehmann:
If you add a new field to const.txt such as TurnsFranchised, then give it a default value so that nobody has to update a mod.

MartinGuehmann:
Actually, I think it should be "!args->GetCity(0, city)", otherwise the index 0 doesn't make sense in the original.

And please, keep the number of arguments, you can call these events also via slic code, and if we want to support old mods then we should not change the number of arguments. Even so, I don't know any mod that is calling the RemoveFranchiseEvent. In fact based on a certain lawyer or army you could stop the RemoveFranchiseEvent. However, we can still add optional arguments.

LynxAbraxas:
Thanks for the feedback.

> If you add a new field to const.txt such as TurnsFranchised, then give it a default value so that nobody has to update a mod.

Sure, but where? I took `TurnsToSueFranchise` as an example and it seems only to get a value from `Const.txt `.

> the expected list of arguments for an event is defined elsewhere.

Any idea where? Could not find that so far. Anyway, I've no problem with the original amount of arguments, it was for testing if they matter and discussing things.
Still the main question remains: Why does
https://github.com/civctp2/civctp2/blob/f726b014a742f640530a451f1c138b5e771a0bbc/ctp2_code/gs/gameobj/armyevent.cpp#L1435
not lead to a call to
https://github.com/civctp2/civctp2/blob/f5a260df6b3561fe7f7246b0ad476e34cfd589b7/ctp2_code/gs/gameobj/CityData.cpp#L5982-L5990
which successfully ends a Franchise in case the `TurnsFranchised` are spent.

LynxAbraxas:
@MartinGuehmann What default value do you think would be appropriate for `TurnsFranchised`?

MartinGuehmann:
> @MartinGuehmann What default value do you think would be appropriate for `TurnsFranchised`?

Actually, I would go for -1, and also for the original code. I think the idea is that the franchise is permanent, but that should be exposed for modders.

The other problem with the countdowns is that it is not shown on the map, so as a player I don't know when it is over.

LynxAbraxas:
> and also for the original code

Why? https://github.com/civctp2/civctp2/pull/167/commits/134044c9d6f16bc72f317e0735480d268d1cda74 was definitly a bug and 
https://github.com/civctp2/civctp2/pull/167/commits/f5a260df6b3561fe7f7246b0ad476e34cfd589b7 probably was not noticed but without https://github.com/civctp2/civctp2/pull/167/commits/0f12a2ac2d7c0647445639627c11c3015c3debd3
the original count-down here
https://github.com/civctp2/civctp2/blob/f5a260df6b3561fe7f7246b0ad476e34cfd589b7/ctp2_code/gs/gameobj/CityData.cpp#L4909-L4917
makes no sense.
With a default of `-1` we would get the original behaviour without the bugs. However, regarding the problem that the AI has hard times countering the Franchises so far (as described in #103 and I have so far never observe that the AI uses a Lawer to rid one of its cities from a Franchise) a value > 0 seems to be a simple way to counter this AI weakness a bit.

MartinGuehmann:
The original code had no countdown, because m_franchiseTurnsRemaining was -1. However, it was only cleared when it reached zero. The lawyer was supposed to set it to something bigger than zero, which however never happened.

LynxAbraxas:
Wasn't `m_franchiseTurnsRemaining= -1` because that is its initialization value?
I see, so
https://github.com/civctp2/civctp2/blob/f5a260df6b3561fe7f7246b0ad476e34cfd589b7/ctp2_code/gs/gameobj/CityData.cpp#L4909-L4917
was only there for the delay to remove a Franchise?

MartinGuehmann:
> Wasn't `m_franchiseTurnsRemaining= -1` because that is its initialization value?
> I see, so
> https://github.com/civctp2/civctp2/blob/f5a260df6b3561fe7f7246b0ad476e34cfd589b7/ctp2_code/gs/gameobj/CityData.cpp#L4909-L4917
> 
> 
> was only there for the delay to remove a Franchise?

I think so, except that it wasn't implemented correctly. Since the remove counter is not set. But of course, it is a good thing to have the option for modders to limit the turns of franchises.

I think we should first get it working as it was originally indented and then we can have the limited number of turns. It is already almost a mod anyway.

LynxAbraxas:
OK, how about these commits. I decided to revert the initial commit instead of just skipping it to use the chance to document some of the understanding gained here. https://github.com/civctp2/civctp2/pull/167/commits/be207ac79f9237560226ccd5628be6a3b2835bb3 is then the actual removal of the bug.

> and then we can have the limited number of turns. It is already almost a mod anyway.

Trying to realize a modular way in #168.

LynxAbraxas:
Hm, seems not to work as expected with the original value of `-1`: With that Lawers cannot sue a Franchise, not sure why though.

LynxAbraxas:
Seems to work with https://github.com/civctp2/civctp2/pull/167/commits/a755b7ad2647ee09d2b0f68539b045c7857c6d97, but that makes me wonder if a value of `-1` was truely meant to indicate infinit number of turns. What do you think @MartinGuehmann ?

MartinGuehmann:
> Seems to work with [a755b7a](https://github.com/civctp2/civctp2/commit/a755b7ad2647ee09d2b0f68539b045c7857c6d97), but that makes me wonder if a value of `-1` was truely meant to indicate infinit number of turns. What do you think @MartinGuehmann ?

Well, we have the invalid player number that is -1 and the invalid database index for buildings etc., which is -1 as well. So from this, it is logical for me that -1 is an infinitive number of turns. Maybe there are also some other database flags entries treated like this.

In principle,  a755b7a would have fixed alone the bug this pull request is about. As far as I can see franchising was quite bugged. In the end, it is about what we want and not what Activision originally wanted. So we can see the code and interpret it in a way it makes sense to us.

LynxAbraxas:
OK, so if you regard this PR now as complete, it can go in since it seems to work the way one would expect from the original code and #168 (not ready yet) would bring the option to modify this on a user basis.

MartinGuehmann:
> Do you see any chance to realize teaching that the AI?

Sure, that can be done.

> OK, so if you regard this PR now as complete, it can go in since it seems to work the way one would expect from the original code and #168 (not ready yet) would bring the option to modify this on a user basis.

There is still an issue, if sue the franchise and the removal time is bigger then zero then you can sue it again. Of course, you can also address it in another pull request.

LynxAbraxas:
> There is still an issue, if sue the franchise and the removal time is bigger then zero then you can sue it again. Of course, you can also address it in another pull request.

Does https://github.com/civctp2/civctp2/pull/167/commits/139e7688567ee60db171ea74a441d8910578b742 solve that sufficiently?

MartinGuehmann:
> > There is still an issue, if sue the franchise and the removal time is bigger then zero then you can sue it again. Of course, you can also address it in another pull request.
> 
> Does [139e768](https://github.com/civctp2/civctp2/commit/139e7688567ee60db171ea74a441d8910578b742) solve that sufficiently?

Looks good. So it can go in.

LynxAbraxas:
>>    Do you see any chance to realize teaching that the AI?

> Sure, that can be done.

@MartinGuehmann, possibly something wrong around here:

https://github.com/civctp2/civctp2/blob/4bb84c20542f4d9d85199c445cffb75b96cd627b/ctp2_code/ai/strategy/goals/Goal.cpp#L2878-L2882

PR #168, opened 2019-07-15
Limit turns a Franchise stays effective

LynxAbraxas:
This PR should be based on #167. It allows the user to provide a modified `Const.txt` in user space, e.g. ~/.civctp2/ (where also `userprofile.txt` and `userkeymap.txt` are placed).
The default  `Const.txt` is then extended to provide a variable for how many turns a Franchise stays active `TurnsFranchised`, initilized with `-1` (infinit) as in the original code.
With this PR a user can copy the default `Const.txt` to user space and modify values, e.g.  `TurnsFranchised` to limit the turns a Franchise is active.
This can make the game more insteresting, since that keeps your Corporate Branches busy. Without this (and without #103) a single Corporate Branch was sufficient to Franchise all foreign cities, without the AI being able to counter either the unit nor the installed Franchises.
With #167, it is now possible to remove Franchises with a Lawer (as it was inteded) but that option is only available to AIs which can already built Lawers, other AIs have no means of getting rid of a Franchise. With this PR the Franchise can at least be made to end after some turns, which also motivates to built more than one Corporate Branch and to re-visit Franchised cities (see also #166).

### DIFF
--- a/ctp2_code/ctp/civapp.cpp
+++ b/ctp2_code/ctp/civapp.cpp
@@ -758,9 +758,10 @@ bool CivApp::InitializeAppDB(void)
 	g_theProgressWindow->StartCountingTo( 40 );
 
 	MBCHAR	fullpath[_MAX_PATH];
-	sprintf(fullpath, "%s%s%s", C3DIR_DIRECT, FILE_SEP, g_constdb_filename);
+	g_civPaths->FindFile(C3DIR_DIRECT, g_constdb_filename, fullpath);
 
-	if (CI_FileExists(fullpath) && g_theConstDB->Parse(C3DIR_DIRECT, g_constdb_filename))
+	if (CI_FileExists(fullpath) &&
+	    g_theConstDB->Parse(C3DIR_DIRECT, g_constdb_filename))
 	{
 		fprintf(stderr, "%s L%d: Using Const.txt from user space!\n", __FILE__, __LINE__);
 	}
--- a/ctp2_code/ctp/civapp.cpp
+++ b/ctp2_code/ctp/civapp.cpp
@@ -757,7 +757,14 @@ bool CivApp::InitializeAppDB(void)
 
 	g_theProgressWindow->StartCountingTo( 40 );
 
-	if (!g_theConstDB->Parse(C3DIR_GAMEDATA, g_constdb_filename))
+	MBCHAR	fullpath[_MAX_PATH];
+	sprintf(fullpath, "%s%s%s", C3DIR_DIRECT, FILE_SEP, g_constdb_filename);
+
+	if (CI_FileExists(fullpath) && g_theConstDB->Parse(C3DIR_DIRECT, g_constdb_filename))
+	{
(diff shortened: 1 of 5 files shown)


LynxAbraxas:
NB: The current code does not seem to include any similar game logic concerning Conversions. It seems Reformation was never intended to be "delayed" like suing a Franchise.

MartinGuehmann:
> NB: The current code does not seem to include any similar game logic concerning Conversions. It seems Reformation was never intended to be "delayed" like suing a Franchise.

Indeed, it doesn't. But since a converted city can be reformed by any unit, conversions should also be cleared after a delay. This means we have to check for some space in the CityData object so that we can squeeze something into the save game format.

Anyway, I think the AI has no trouble with it.

LynxAbraxas:
Works as expected now, but should not be merged before #167.

LynxAbraxas:
Rebased onto current master, works as expected. 
@MartinGuehmann Can be merged if you see no problems.

MartinGuehmann:
> Rebased onto current master, works as expected.
> @MartinGuehmann Can be merged if you see no problems.

The only thing I have some hassle with is having a copy of const.txt in the user directory. I rather consider it as a database file and not a user setting file.

However, we can keep it. Well it is supposed that you cannot join a multiplayer game if your database does not match.

LynxAbraxas:
> However, we can keep it. Well it is supposed that you cannot join a multiplayer game if your database does not match.

True, did not think of that. Any suggestion how to solve this in a better way? Do you know of a place where this could be defined in user-space but also be compatible with multiplayer? How are the other game-rules a user can modify (like upgrading units) synced in case of a multiplayer game?

MartinGuehmann:
> > However, we can keep it. Well it is supposed that you cannot join a multiplayer game if your database does not match.
> 
> True, did not think of that. Any suggestion how to solve this in a better way? Do you know of a place where this could be defined in user-space but also be compatible with multiplayer?

That's difficult to answer, user space may even mean here the directory of the executable. In principle the thing is compatible with multiplayer, you just have to make sure that everybody had the same const.txt, whether it comes from the ctp2data subfolders or from another place does not matter.

How are the other game-rules a user can modify (like upgrading units) synced in case of a multiplayer game?

Well, the databases have serialize methods, even so they are not used for save games, they are used for checking the databases, basically you convert them to binary and then compare. I do not remember whether there is a hash value involved, but yeah that is rather a technical detail.

Then we have the game rules of the userprofile.txt, those are either on or of. Upgrading is something like this. I also assume that it is checked somehow and maybe even disabled, but here I would have to look into the code. In principle, it could be for upgrading that you could play it with being enabled on one computer and on the other computer with being disabled. I don't assume that E has addressed all the details about it.

LynxAbraxas:
Hm, so shall we stick to the current way and record an issue concerning multiplayer (which seems to be hardly tested any way)?

MartinGuehmann:
> Hm, so shall we stick to the current way and record an issue concerning multiplayer (which seems to be hardly tested any way)?

Yes, so this can then go in.

PR #171, opened 2019-07-20
Draw conversion/franchise loss value on corresponding city-icon

LynxAbraxas:
This PR offers the drawing of the franchise loss value on franchise icon like:
ss_2019-07-20_23:23:27

### DIFF
--- a/ctp2_code/gfx/tilesys/tiledmap.h
+++ b/ctp2_code/gfx/tilesys/tiledmap.h
@@ -436,8 +436,8 @@ public:
 	void		DrawCityIcons(aui_Surface *surf, MapPoint const &pos, sint32 owner, bool fog, RECT &popRect,
 								BOOL isBioInfected, BOOL isNanoInfected, BOOL isConverted,
 								BOOL isFranchised, BOOL isInjoined, BOOL wasHappinessAttacked,
-								sint32 bioInfectedOnwer, sint32 nanoInfectedOwner, sint32 convertedOwner,
-								sint32 franchiseOwner, sint32 injoinedOwner, sint32 happinessAttackOwner,
+								sint32 bioInfectedOnwer, sint32 nanoInfectedOwner, sint32 convertedOwner, sint32 convertedLoss,
+								sint32 franchiseOwner, sint32 franchiseLoss, sint32 injoinedOwner, sint32 happinessAttackOwner,
 								uint32 slaveBits, BOOL isRioting, BOOL hasAirport, BOOL hasSleepingUnits,
 								BOOL isWatchful, BOOL isCapitol, BOOL isProdIcon, sint32 citySize,
 								BOOL isPollutionRisk);
--- a/ctp2_code/gfx/tilesys/tiledraw.cpp
+++ b/ctp2_code/gfx/tilesys/tiledraw.cpp
@@ -3446,8 +3446,8 @@ void TiledMap::DrawCityNames(aui_Surface * surf, sint32 layer)
 					    franchiseLoss    = unit.GetData()->GetCityData()->GetProductionLostToFranchise();
 					    }
 					else{// in case city was destroyed since last visit
-					    convertedLoss    = ucell.m_unseenCell->GetConvertedGold();
-					    franchiseLoss    = ucell.m_unseenCell->GetProductionLostToFranchise();
+					    convertedLoss    = 0; // no info on loss if city was destroyed
+					    franchiseLoss    = 0; // no info on loss if city was destroyed
 					    }
 					isInjoined           = ucell.m_unseenCell->IsInjoined();
 					wasHappinessAttacked = ucell.m_unseenCell->WasHappinessAttacked();
@@ -4221,7 +4221,7 @@ void TiledMap::DrawCityIcons(aui_Surface *surf, MapPoint const & pos, sint32 own
 		DrawColorizedOverlay(cityIcon, surf, iconRect.left, iconRect.top, color);
 
 		sint32 myOwner = g_selected_item->GetVisiblePlayer();
-		if(myOwner >= 0 && (myOwner == owner || myOwner == convertedOwner))
+		if(convertedLoss && myOwner >= 0 && (myOwner == owner || myOwner == convertedOwner))
 		    {
 		    sint32 width, height;
 		    RECT rect;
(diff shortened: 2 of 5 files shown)


LynxAbraxas:
ToDo: 
- Make value independent of FOW, e.g. value should not be 0 but 17 for "Ad Dawhah" of the Arabians above
- Do not draw value for 3rd parties, e.g. the Arabians should not see a value for "Nueva Guinea" of the Nicaraguans:
ss_2019-07-20_23:23:58

LynxAbraxas:
> - Make value independent of FOW, e.g. value should not be 0 but 17 for "Ad Dawhah" of the Arabians above

https://github.com/civctp2/civctp2/pull/171/commits/9454740da4ebd3bcf71375f77c9eec2887b19efb solves this:

ss_2019-07-21_00:25:06

The un-seen cell value is only needed in case a city got destroyed:

ss_2019-07-21_00:26:31

However, currently the `m_franchiseLoss` of `UnseenCell` is not saved nor restored.
@MartinGuehmann do you know of a slot where an sint32 could be saved in?

LynxAbraxas:
> - Do not draw value for 3rd parties, e.g. the Arabians should not see a value for "Nueva Guinea" of the Nicaraguans:

https://github.com/civctp2/civctp2/pull/171/commits/66ef8b7ca3849bbca83b2d3a521a77677f34efa8 solves this:

ss_2019-07-21_10:05:58

All also now working for conversion loss, as Persian:

ss_2019-07-21_10:54:27

as Italian:

ss_2019-07-21_10:54:54

LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/171/commits/8643646d937e975b46655c968fd06940908cad98 the loss number is not 0 directly after conversion/franchise as it was the case before.

LynxAbraxas:
> However, currently the `m_franchiseLoss` of UnseenCell is not saved nor restored.

@MartinGuehmann do you see any chance to store 2 sint32 (`m_franchiseLoss` and `m_convertedLoss`) in the current save-game format?

MartinGuehmann:
> > However, currently the `m_franchiseLoss` of UnseenCell is not saved nor restored.
> 
> @MartinGuehmann do you see any chance to store 2 sint32 (`m_franchiseLoss` and `m_convertedLoss`) in the current save-game format?
> 
> Otherwise the PR would be ready.

First, why would you like to save and load both members? And why do you have added them to the UnseenCell object in the first place. As far as I understood it you only show these values to the first and second party, but never to a third party. The first party is the city owner who has no FOW on the city. And the second party is the franchise owner who knows the value despite any FOW. Anybody only knows about the franchise and that should it be.

Then we should think about, whether it is a good idea to expand the UnseenCell object, anyway. You added two sint32 members. That are 8 bytes more and you inserted them in such away that you waste another 4 bytes. That has something to do with data alignment. That's actual why I added this PrintSizeOfClass() function. So we have additional 12 bytes.

On a gigantic map we have 70 * 140 = 9800 tiles. Then the question is how much we go up with the number of players.We can have up to 32 players, probably the Barbarians do not count. Then we have 31. Each discovered tile, which is not seen is under fog of war. For simplicity, I assume that each tile is visible at least for one player, so we have at maximum for 30 players 9800 tiles unseen. So that are then 9800*30*12 = 3528000 bytes. For simplicity 1 kbyte = 1000 bytes, we get some 3,5 Mbyte.

Actually, that is not so much extra space then I thought but it is something and you have quite many of those UnseenCell objects in the game. Actually, this is more important for debug version, since the debug version needs much longer to allocate all that memory, since it is also logging the memory allocation.

Anyway, this is a place were we should not add memory needlessly.

LynxAbraxas:
I agree to your thoughts to not add memory needlessly. My question was whether there is a chance to store those in the current save-game format, i.e. in unused variables that are stored so far any way. Let's see if changes to UnseenCell can be avoided.

LynxAbraxas:
So let's take the simple approach to not show conversion/franchise info to the attacker in case a city was destroyed, the same as in the case a conversion/franchise is removed.

With just https://github.com/civctp2/civctp2/pull/171/commits/00868329cc5da721cb2db6a8c82600742629006f the lack of the loss value indirectly gives away that the city was destroyed

ss_2019-07-22_11:26:09

while with https://github.com/civctp2/civctp2/pull/171/commits/e88344dfdf55097a38331bd6129e5099373fc48b this is not the case:

ss_2019-07-22_11:45:50

LynxAbraxas:
@MartinGuehmann what do you think? Can it go in as is now?

MartinGuehmann:
> I agree to your thoughts to not add memory needlessly. My question was whether there is a chance to store those in the current save-game format, i.e. in unused variables that are stored so far any way. Let's see if changes to UnseenCell can be avoided.

To answer that question: Yes. Except that we don't have unused variables we could use, but we have Dynamic array that was unused in the original game, I already used it to squeeze additional information into the save game.

> @MartinGuehmann what do you think? Can it go in as is now?

I am not sure, yet. It depends if you wanna perfection. For now the last image gives away that your neighbor does not own anymore the city of Ad Dawhah. That's because of the borders, however the image does not tell me whether it is because the city revolted and formed a new civilization, which the player on screen hasn't met, yet. Or whether it is, because the city does not exist anymore.

Actually, when the city changes hands you now are informed that the franchise has been remove. I guess if the city is destroyed then you won't get a message.

However, I think that we can give this away. If you have a franchise in that city then you should notice unless the war did not left any survivors who could spread the message. I tend to leave this issue as it is.

Another question is what happens if you have a three digit loss, does it still fit onto the icon?

LynxAbraxas:
> However, I think that we can give this away. If you have a franchise in that city then you should notice unless the war did not left any survivors who could spread the message. I tend to leave this issue as it is.

That's what I wanted to indicate with "quite realistic". The drawing of the destroyed city in the screen-shot above might not be as in a normal game since I destroyed that city in cheat-mode using the eraser. I would actually expect that the city sprite would be visible still, but if that is not working I'd say that is another issue.

> Another question is what happens if you have a three digit loss, does it still fit onto the icon?

Tested that, but already had a hard time to get a city pimped-up even in cheat -mode to have a franchise loss with 3 digits. With a 1 at the beginning it kind of fits:

ss_2019-07-21_10:13:01

It's not perfect but apparently unlikely. I had also though of rotating the number by 90° but when investigating 
https://github.com/civctp2/civctp2/blob/2c9b619e1bb4fe4e3fab1b86dddf521482e930d7/ctp2_code/ui/aui_common/aui_bitmapfont.cpp#L758-L984
figured that is not available and would need a lot of changes/additions. There is the option to wrap text but I think that would look far too ugly in this case.
Anyway, the indicator for the turns to the next population increase also cannot do more than 2 digits.

Unless you have another idea how to handle this better, I'd say this can go in.

MartinGuehmann:
> Anyway, the indicator for the turns to the next population increase also cannot do more than 2 digits.

I think, this then goes to infinity or ---. So that should be fine.

> Unless you have another idea how to handle this better, I'd say this can go in.

I think, we just put it in and if the three digits turn out to be a problem then we'll fix it later.

LynxAbraxas:
NB: With a TV the conversion loss can easily need 3 digits, which still are displayed all right:

ss_2019-09-23_22:36:10

PR #172, opened 2019-07-21
Violation event messages for victim

LynxAbraxas:
PR for #162, that intends to re-implement the CTP1 messages for the victim of an agreement violation for CTP2 diplomacy code.
Despite being helpful for humans (as described in #162), this would help to check if the AI is capable of breaking standing treaties/agreements. At least it seems that the AI cannot break a no-piracy agreement like a human can with #161, see https://github.com/civctp2/civctp2/pull/164#issuecomment-510256835.


LynxAbraxas:
With #175, works as expected when a no-piracy agreement is broken:

ss_2019-07-22_12:42:36

This is the save-game with a standing no-piracy agreement where the Minoans intend to pirate in the next turn (as in the screen-shot above): [break-no-piracy.zip](https://github.com/civctp2/civctp2/files/3417963/break-no-piracy.zip) (don't forget to do a `/reloadslic` before ending the turn).

LynxAbraxas:
While testing #173, it seems the AI does not break a standing withdraw-troops agreement, at least I never observed so nor was able to force it to. However, I was also not able to find any code similar to #175 that prevents the AI from breaking a withdraw-troops agreement.
@MartinGuehmann any idea if that is the case or how to force the AI to break a withdraw-troops agreement?

MartinGuehmann:
> While testing #173, it seems the AI does not break a standing withdraw-troops agreement, at least I never observed so nor was able to force it to. However, I was also not able to find any code similar to #175 that prevents the AI from breaking a withdraw-troops agreement.
> @MartinGuehmann any idea if that is the case or how to force the AI to break a withdraw-troops agreement?

The code is in diplomat.cpp:

https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/ai/diplomacy/diplomat.cpp#L4570-L4602

I am bit unsure what now cold war regard means.

This is then copied in robotastar2.cpp to m_incursionPermission and then used here:

https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/robot/pathing/robotastar2.cpp#L181-L209

Basically, the AI would only walk into your land if it has to. However, if it moves into your land there is also some danger, so it would avoid your dangerous armies, too. Unless of course it wants to go into your territory.

LynxAbraxas:
Many thanks @MartinGuehmann  for those pointers.
It turned out that it was quite easy to get the violation message as a victim in my current test-game:

ss_2019-08-04_02:02:31

So that works as expected but I am a bit disappointed that the AI did violate that agreement so ruthlessly. I would not have expected that with the code quoted above.

MartinGuehmann:
> Many thanks @MartinGuehmann for those pointers.
> It turned out that it was quite easy to get the violation message as a victim in my current test-game:
> 
> ss_2019-08-04_02:02:31
> 
> So that works as expected but I am a bit disappointed that the AI did violate that agreement so ruthlessly. I would not have expected that with the code quoted above.

So what did violate the withdraw troops agreement? What it an AI transporter? Actually, those may just cross the line. So you should still look at the places where DefensivePathCallback is called from and the alternatives for transport and airlift. My impression is that the AI now breaks this agreement without intention.

LynxAbraxas:
>  My impression is that the AI now breaks this agreement without intention.

Yes, that very well fits my impression, too. I think it was an AI transporter, at least a transport moved into my territory, not sure if another unit actually cased the violation before.

LynxAbraxas:
Since it is kind of related, any idea/comment concerning https://github.com/civctp2/civctp2/pull/175#issuecomment-515758932?

LynxAbraxas:
> My impression is that the AI now breaks this agreement without intention.

Possibly that happens if the army already had a move-to order set before a withdraw troops agreement was made.

MartinGuehmann:
> > My impression is that the AI now breaks this agreement without intention.
> 
> Possibly that happens if the army already had a move-to order set before a withdraw troops agreement was made.

Not really, because these things are recalculated each turn. To fix this you have to go to robotastar2.cpp. And change RobotAstar2::TransportPathCallback and RobotAstar2::AirliftPathCallback also call RobotAstar2::DefensivePathCallback.

Actually, the question is whether transporters are supposed to break withdraw troop agreements you could also transport a diplomat or a settler or other civilists. So should a diplomat break a withdraw agreement whether he comes by food or by ship?

LynxAbraxas:
Concerning the breaking of a cease-fire treaty: Player::BreakCeaseFire is used in ArmyData::Move, which seems to be a left over from CTP1.   However, a replacement for WillViolateCeaseFire and possibly for BreakCeaseFire in CTP2 code (similar to DoLeaveOurLandsCheck, #173) might be appropriate.

LynxAbraxas:
There seems to be no Eco-Pact in CTP2, i.e. no correspondence to `005EcopactBroken`.

LynxAbraxas:
With e26eda2 I get the cease-fire-broken message (2730dd6) as a victim
ss_2019-08-29_20:00:28

LynxAbraxas:
It seems a peace treaty is not added to an existing cease-fire treaty but removes/replaces this in the agreement-matrix. Therefore breaking a peace treaty does not lead to a cease-fire-broken message. I'm not sure if this has other possibly unwanted side-effects.

MartinGuehmann:
> @MartinGuehmann Do you think [e26eda2](https://github.com/civctp2/civctp2/commit/e26eda20f14833cc28687676f23a482e32c3eefd) is fine for the current game logic?

I think there is a problem: Before you go to war, you have no treaty at all. But after a war you have a cease fire, which unlike in CTP1, does not expire. So if you turn it into a treaty, the state is now different then before. But that is a more conceptual problem, with the thing itself.

LynxAbraxas:
> So if you turn it into a treaty, the state is now different then before. 

Not sure I get what you mean here.

LynxAbraxas:
@MartinGuehmann how about this PR? Should I just revert [e26eda2](https://github.com/civctp2/civctp2/commit/e26eda20f14833cc28687676f23a482e32c3eefd) for this long standing PR to go in?

MartinGuehmann:
> @MartinGuehmann how about this PR? Should I just revert [e26eda2](https://github.com/civctp2/civctp2/commit/e26eda20f14833cc28687676f23a482e32c3eefd) for this long standing PR to go in?

I think, I had there some conceptual problem there. Actually the thing is only called from Diplomat::Execute_Proposal so then it should be fine doing it. There seems to be a difference between cancel and break n agreement, but that is another PR.

So that can go in. Just resolve the conflicts, and also update the info_str.txt files of the new languages. Make sure that all files have the same number of lines. That makes it a bit easier to compare files across the different languages.

LynxAbraxas:
> So that can go in. Just resolve the conflicts, and also update the info_str.txt files of the new languages. Make sure that all files have the same number of lines. That makes it a bit easier to compare files across the different languages.

OK, rebased onto current master and resolved conflicts, and all have the same number of lines:
```
wc -l ctp2_data/*/gamedata/info_str.txt 
   640 ctp2_data/chinese/gamedata/info_str.txt
   640 ctp2_data/czech/gamedata/info_str.txt
   640 ctp2_data/english/gamedata/info_str.txt
   640 ctp2_data/german/gamedata/info_str.txt
   640 ctp2_data/italian/gamedata/info_str.txt
   640 ctp2_data/polish/gamedata/info_str.txt
   640 ctp2_data/spanish/gamedata/info_str.txt
```

PR #174, opened 2019-07-22
Save m_goldFromTransitRoutes instead of m_foodVatPollution

LynxAbraxas:
Concerning https://github.com/civctp2/civctp2/pull/147#issuecomment-509025872 this PR removes the unused `m_foodVatPollution` and saves `m_goldFromTransitRoutes` instead such that the total gold from transit routes is even uptodate when loading a game.

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -402,7 +402,6 @@ CityData::CityData(PLAYER_INDEX owner, Unit hc, const MapPoint &center_point)
 	m_total_pollution                   (0),
 	m_cityPopulationPollution           (0),
 	m_cityIndustrialPollution           (0),
-	m_foodVatPollution                  (0),
 	m_cityPollutionCleaner              (0),
 	m_contribute_materials              (true),
 	m_contribute_military               (true),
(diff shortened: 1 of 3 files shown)


LynxAbraxas:
@MartinGuehmann  what do you think of this PR? Does it handle the saving appropiratly?
Do you know what `m_foodVatPollution` was meant/used for?

MartinGuehmann:
> @MartinGuehmann what do you think of this PR? Does it handle the saving appropiratly?

The saving is handled, appropriately. The question is just whether you want to save it or whether you can regenerate it from state. That means from the trade routes, once everything has been loaded. If you look at the indicators for how many turns to the next pop next to the city name then you see they are also at zero. However this changes when the resources are recalculated.

> Do you know what `m_foodVatPollution` was meant/used for?

That's a bit difficult, I guess it has something to do with the Cornucopic Vat. Here is what the gameplay says:

Cornucopic Vats synthesize a cheap, abundant, soy-based food substance that satisfies all nutritional requirements.  Although it may not appease the taste buds of everyone, it is certainly a step towards feeding all citizens.

The German version calls it Nahrungsreplikator, something you may know from Star Trek. The building's entry has the flag FoodVat. Maybe that is indented to add some pollution, too.

LynxAbraxas:
> The question is just whether you want to save it or whether you can regenerate it from state.

That's what I tried, but could not find a way because the transit accounting happens at the beginning of a turn (which makes sense because pirating is evaluated then as well). Unless you see a way of solving that differently, I would go for the saving as in this PR.

> Maybe that is indented to add some pollution, too.

Thanks for that explanation, never got that far so far. I guess since that apparently was not implemented, we could just go with the current solution.

MartinGuehmann:
> > The question is just whether you want to save it or whether you can regenerate it from state.
> 
> That's what I tried, but could not find a way because the transit accounting happens at the beginning of a turn (which makes sense because pirating is evaluated then as well). Unless you see a way of solving that differently, I would go for the saving as in this PR.

OK, it is at least difficult to regenerate it from state, so that it makes sense to save it.

> > Maybe that is indented to add some pollution, too.
> 
> Thanks for that explanation, never got that far so far. I guess since that apparently was not implemented, we could just go with the current solution.

Well, I haven't checked what FoodVat does, but obviously it doesn't use m_foodVatPollution in CityData.

So then, this can go in.

Issue #169, opened 2019-07-20
AI apparently neither doing Franchise attacks nor Suing Franchises in its cities


LynxAbraxas:
Apparently the AI neither does Franchise attacks nor Suing Franchises in its cities, see also https://github.com/civctp2/civctp2/pull/167#issuecomment-512394317.

LynxAbraxas:
Possibly something wrong around here:
https://github.com/civctp2/civctp2/blob/4bb84c20542f4d9d85199c445cffb75b96cd627b/ctp2_code/ai/strategy/goals/Goal.cpp#L2878-L2882

LynxAbraxas:
@MartinGuehmann any idea how to make this possible?

LynxAbraxas:
Concerning the establishing of Franchises:
@MartinGuehmann shouldn't this test
https://github.com/civctp2/civctp2/blob/2c9b619e1bb4fe4e3fab1b86dddf521482e930d7/ctp2_code/ai/strategy/goals/Goal.cpp#L2890-L2894
be based on the franchise owner being `>= 0` instead of `FranchiseTurnsRemaining`?

LynxAbraxas:
Concerning suing of Franchises:
This seems fine
https://github.com/civctp2/civctp2/blob/2c9b619e1bb4fe4e3fab1b86dddf521482e930d7/ctp2_code/ai/strategy/goals/Goal.cpp#L2878-L2882
but might be without effect due to
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/Unit.cpp#L2422-L2423
which probably is due to the fact that a `CanSueFranchise` apparently is not implemented.

LynxAbraxas:
Before 76f7820eb764077640cdf2eeaed6d9dd32894ba3 there was
https://github.com/civctp2/civctp2/blob/1fa7e961cb47c3012c64f6ec443e3eb779aebaf6/ctp2_code/robotcom/Agent/armyagent.cpp#L2316-L2336

LynxAbraxas:
@MartinGuehmann these lines
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/Unit.cpp#L2422-L2423
seem to appear in e500b666413db49f981a19bbd541e06a86396b8a, any chance you remember where those originated from?

[...]

PR #176, opened 2019-07-22
Debug Franchise and SueFranchise AI orders

LynxAbraxas:
PR to solve issue #169.

Apparently, letting the AI establish Franchises fails due to https://github.com/civctp2/civctp2/issues/169#issuecomment-513934362
and letting the AI sue Franchised cities of its own fails due to https://github.com/civctp2/civctp2/issues/169#issuecomment-513942751.

### DIFF
--- a/ctp2_code/ai/strategy/goals/Goal.cpp
+++ b/ctp2_code/ai/strategy/goals/Goal.cpp
@@ -2889,7 +2889,7 @@ bool Goal::IsTargetImmune() const
 
 	if(order_record->GetUnitPretest_CanCreateFranchise())
 	{
-		if(m_target_city.GetCityData()->GetFranchiseTurnsRemaining() <= 0)
+		if(m_target_city.GetCityData()->GetFranchiseOwner() == m_playerId)
 			return true;
 	}
 
--- a/ctp2_code/gs/events/GameEventDescription.h
+++ b/ctp2_code/gs/events/GameEventDescription.h
@@ -84,7 +84,7 @@ GameEventDescription g_eventDescriptions[] =
 	EVENT(BombardOrder,               "Army bombards", "%a%l")
 	EVENT(FranchiseOrder,             "Army franchises", "%a%l")
 	EVENT(SueOrder,                   "Army sues", "%a%l")
-	EVENT(SueFranchiseOrder,          "Army sues franchise", "%a")
+	EVENT(SueFranchiseOrder,          "Army sues franchise", "%a%l")
 	EVENT(ExpelOrder,                 "Army expels", "%a%l")
 	EVENT(EstablishEmbassyOrder,      "Army establishes embassy", "%a%l")
 	EVENT(AdvertiseOrder,             "Army advertises", "%a%l")
--- a/ctp2_code/gs/events/GameEventDescription.h
+++ b/ctp2_code/gs/events/GameEventDescription.h
@@ -84,7 +84,7 @@ GameEventDescription g_eventDescriptions[] =
 	EVENT(BombardOrder,               "Army bombards", "%a%l")
 	EVENT(FranchiseOrder,             "Army franchises", "%a%l")
 	EVENT(SueOrder,                   "Army sues", "%a%l")
-	EVENT(SueFranchiseOrder,          "Army sues franchise", "%a%l")
+	EVENT(SueFranchiseOrder,          "Army sues franchise", "%a")
 	EVENT(ExpelOrder,                 "Army expels", "%a%l")
 	EVENT(EstablishEmbassyOrder,      "Army establishes embassy", "%a%l")
 	EVENT(AdvertiseOrder,             "Army advertises", "%a%l")
--- a/ctp2_code/gs/gameobj/Army.cpp
+++ b/ctp2_code/gs/gameobj/Army.cpp
@@ -352,19 +352,9 @@ bool Army::CanSue(sint32 &uindex)
 	return AccessData()->CanSue(uindex);
 }
 
-bool Army::CanSueFranchise(sint32 &uindex)
-{
-	return AccessData()->CanSueFranchise(uindex);
-}
-
 bool Army::CanSue()
 {
 	return AccessData()->CanSue();
-}
-
-bool Army::CanSueFranchise()
-{
-	return AccessData()->CanSueFranchise();
 }
 
 bool Army::CanBeSued()
--- a/ctp2_code/gs/gameobj/Army.cpp
+++ b/ctp2_code/gs/gameobj/Army.cpp
@@ -352,9 +352,19 @@ bool Army::CanSue(sint32 &uindex)
 	return AccessData()->CanSue(uindex);
 }
 
+bool Army::CanSueFranchise(sint32 &uindex)
+{
+	return AccessData()->CanSueFranchise(uindex);
+}
+
 bool Army::CanSue()
 {
 	return AccessData()->CanSue();
+}
+
+bool Army::CanSueFranchise()
+{
+	return AccessData()->CanSueFranchise();
 }
 
 bool Army::CanBeSued()
--- a/ctp2_code/gs/gameobj/Army.h
+++ b/ctp2_code/gs/gameobj/Army.h
@@ -157,7 +157,9 @@ public:
 	bool CanFranchise(double &chance, sint32 &uindex);
 
 	bool CanSue(sint32 &uindex);
+	bool CanSueFranchise(sint32 &uindex);
 	bool CanSue();
+	bool CanSueFranchise();
 	bool CanBeSued();
 	bool CanCauseUnhappiness(double &chance, sint32 &timer, sint32 &amount,
 							 sint32 &uindex);
--- a/ctp2_code/gs/gameobj/Army.h
+++ b/ctp2_code/gs/gameobj/Army.h
@@ -157,9 +157,7 @@ public:
 	bool CanFranchise(double &chance, sint32 &uindex);
 
 	bool CanSue(sint32 &uindex);
-	bool CanSueFranchise(sint32 &uindex);
 	bool CanSue();
-	bool CanSueFranchise();
 	bool CanBeSued();
 	bool CanCauseUnhappiness(double &chance, sint32 &timer, sint32 &amount,
 							 sint32 &uindex);
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -9508,7 +9508,6 @@ bool ArmyData::ExecuteSpecialOrder(Order *order, bool &keepGoing)
 			break;
 		case UNIT_ORDER_SUE_FRANCHISE:
 			result = SueFranchise(order->m_point);
-			useDefaultSuccessSound = true;
 			break;
 		case UNIT_ORDER_EXPEL:
 			result = Expel(order->m_point);
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -2234,11 +2234,6 @@ bool ArmyData::CanSue(sint32 &index) const
 	return false;
 }
 
-
-bool ArmyData::CanSueFranchise(sint32 &index) const
-{
-	return CanSue(index);
-}
 //----------------------------------------------------------------------------
 //
 // Name       : ArmyData::CanSue
@@ -2263,11 +2258,6 @@ bool ArmyData::CanSue() const
 		}
 	}
 	return false;
-}
-
-bool ArmyData::CanSueFranchise() const
-{
-	return CanSue();
 }
 
 //----------------------------------------------------------------------------
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -2312,6 +2312,7 @@ ORDER_RESULT ArmyData::Sue(const MapPoint &point)
 	}
 	Cell *cell = g_theWorld->GetCell(point);
 	if(!cell->UnitArmy() || !cell->UnitArmy()->CanBeSued() || cell->UnitArmy()->GetOwner() == m_owner) {
+	    /* SueFranchise now handled by SueFranchiseOrder directly
 		if(cell->GetCity().m_id != 0 &&
 		   cell->GetCity().GetOwner() == m_owner) {
 			Assert(m_orders->GetHead() && m_orders->GetHead()->m_order == UNIT_ORDER_SUE);
@@ -2320,7 +2321,7 @@ ORDER_RESULT ArmyData::Sue(const MapPoint &point)
 			}
 			return SueFranchise(point);
 		}
-
+	    */
 		return ORDER_RESULT_ILLEGAL;
 	}
 
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -2234,6 +2234,11 @@ bool ArmyData::CanSue(sint32 &index) const
 	return false;
 }
 
+
+bool ArmyData::CanSueFranchise(sint32 &index) const
+{
+	return CanSue(index);
+}
 //----------------------------------------------------------------------------
 //
 // Name       : ArmyData::CanSue
@@ -2258,6 +2263,11 @@ bool ArmyData::CanSue() const
 		}
 	}
 	return false;
+}
+
+bool ArmyData::CanSueFranchise() const
+{
+	return CanSue();
 }
 
 //----------------------------------------------------------------------------
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -9508,6 +9508,7 @@ bool ArmyData::ExecuteSpecialOrder(Order *order, bool &keepGoing)
 			break;
 		case UNIT_ORDER_SUE_FRANCHISE:
 			result = SueFranchise(order->m_point);
+			useDefaultSuccessSound = true;
 			break;
 		case UNIT_ORDER_EXPEL:
 			result = Expel(order->m_point);
--- a/ctp2_code/gs/gameobj/ArmyData.h
+++ b/ctp2_code/gs/gameobj/ArmyData.h
@@ -216,9 +216,7 @@ public:
     ORDER_RESULT Franchise(const MapPoint &point);
 
     bool CanSue(sint32 &uindex) const;
-    bool CanSueFranchise(sint32 &uindex) const;
     bool CanSue() const;
-    bool CanSueFranchise() const;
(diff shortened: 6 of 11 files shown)


LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/176/commits/978ace19879b65a01cf619a2e609d81ac0e6e50e the AI establishes Franchises.
Giving the Arabians lots of Corporate Branches (in cheat-mode) 

ss_2019-07-23_00:33:15

then becoming another civ again (Persian) and ending the turn leads to some Franchises of the Arabians

ss_2019-07-23_00:41:25

and Advertisement (only visible before the new turn begins)

ss_2019-07-23_00:34:57

LynxAbraxas:
@MartinGuehmann what maps this
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_data/default/gamedata/Orders.txt#L803-L822
to
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L2365-L2414

LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/176/commits/6d30143c233424d5c85a5e04c3bde40c3d2b031a I get a new order icon for the Lawer to sue Franchises
ss_2019-07-23_17:50:56
However, using that order on a city has no effect while the Sue-order still removes the Franchise all right.
@MartinGuehmann any idea what is still wrong? I guess some checks in Sue have to be moved to SueFranchise.

LynxAbraxas:
Here is some mapping done
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/Order.cpp#L258-L279
not sure though if the problem lies there.

MartinGuehmann:
> @MartinGuehmann what maps this
> https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_data/default/gamedata/Orders.txt#L803-L822
> 
> 
> to
> https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L2365-L2414
> 
> 
> Is it just stripping the `Order` from the `EventName`?

No, probably ArmyData::SueFranchise is just called from the SueFranchiseOrder event, directly or inderiectly. EventName is a slic event, which also can be could from slic code.

MartinGuehmann:
> With [6d30143](https://github.com/civctp2/civctp2/commit/6d30143c233424d5c85a5e04c3bde40c3d2b031a) I get a new order icon for the Lawer to sue Franchises
> ss_2019-07-23_17:50:56

Yeah, that's expected

> However, using that order on a city has no effect while the Sue-order still removes the Franchise all right.
> @MartinGuehmann any idea what is still wrong? I guess some checks in Sue have to be moved to SueFranchise.

Maybe, but you could also check whether the SueFranchiseOrder event gets the correct arguments or expects the correct arguments. The problem could be in a chain, but it starts with SueFranchiseOrder event.

LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/176/commits/d80088bc986fcd643f36f7dc6b1c646eb550d02c SueFranchise order gets up to here
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L10823-L10828
but not to here (as is the case for Sue)
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L9422

LynxAbraxas:
> Maybe, but you could also check whether the SueFranchiseOrder event gets the correct arguments or expects the correct arguments.

Many thanks @MartinGuehmann for that hint, still took me some hours to get it right eventhough the changes in the end were simple: 
With https://github.com/civctp2/civctp2/pull/176/commits/455543d3e2e0a958c4fbdfad3965f27d847b73ee not only the SueFranchise button/order works as expected, also the AI makes use of it to remove Franchise in its own city:

ss_2019-07-24_13:54:29

LynxAbraxas:
Also works with https://github.com/civctp2/civctp2/pull/176/commits/a99b0306fb2cb9b880e874f609d38cdfecb1340f but something is wrong with what special order was used. The order-buttons after removing a Franchise for a Lawer inside a city

ss_2019-07-24_14:45:05

and outside a city

ss_2019-07-24_14:45:19

are not consistent. This also seems to irritate the AI as with https://github.com/civctp2/civctp2/pull/176/commits/a99b0306fb2cb9b880e874f609d38cdfecb1340f it tries to remove further Franchises after having removed one. I guess there might be something wrong here
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L10465-L10551
probably something depending on `range == 0`.

In addition the sound in case the order cannot be performed is this kind of cashier bell (which I oberseved for other orders before) which is kind of irritating since it lets one associate something successful.

LynxAbraxas:
The odd sound does not appear with https://github.com/civctp2/civctp2/pull/176/commits/3025c62bd5ff6b12eae16160d938b33fe743911a, so I guess it is connected to `range == 0` logic.

MartinGuehmann:
> probably something depending on `range == 0`. The behavior is similar in case of using ReformCity but in that case the unit only has one special action (as far as I know).

Or it has something to do with CanPerformSpecialAction().

LynxAbraxas:
> Or it has something to do with CanPerformSpecialAction().

Does not look like it. `CanPerformSpecialAction()` does not depend on the position, as is the case here (see screen-shots above).

MartinGuehmann:
> > Or it has something to do with CanPerformSpecialAction().
> 
> Does not look like it. `CanPerformSpecialAction()` does not depend on the position, as is the case here (see screen-shots above).

Then we have TargetValidForOrder, but it doesn't seem to be it either. Unless there is also a problem with the clerics. However, I would expect that to be the place to check whether you can execute, orders like reform and sue franchises.

LynxAbraxas:
Hm, the only thing related there seems to be
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L10654
but it made me wonder if perhaps we need more pre-test specs for the Sue and SueFranchise order.

MartinGuehmann:
> Hm, the only thing related there seems to be
> https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L10654
> 
> 
> but it made me wonder if perhaps we need more pre-test specs for the Sue and SueFranchise order, must admit I feel a bit lost there.

We may need more pre-test. So can you franchise a city over and over again as a human player? And can you sue a franchise over and over again as a human player, even so the city has no franchise anymore. And can you convert cities over and over again, even so the city is already converted? And can you reform a city over and over again, even so it is not converted anymore?

LynxAbraxas:
> We may need more pre-test. So can you franchise a city over and over again as a human player? And can you sue a franchise over and over again as a human player, even so the city has no franchise anymore. And can you convert cities over and over again, even so the city is already converted? And can you reform a city over and over again, even so it is not converted anymore?

As far as I could test this, the logic to disable the action if it was used or if the target is already effected works as expected. Most of that logic is in the function that would execute the order but not tested before. I think that is the source for the button enable/disable problem and for the AI trying because it does not know from the tests before.
In my last tests I saved the game modified in cheat-mode and then tested after a reload. The buttons for a Lawyer are then always enabled, independent if the action was already used by that unit but at least also independent from the Lawer position. So it might be some odd behavior originating from using the cheat-mode. @MartinGuehmann is the use of the cheat-mode saved in some way or can I expect that a reloaded game does not know about its former use?

LynxAbraxas:
The no sufficently specific pre-tests seem to be another problem, not neccessarily bound to franchising and suing franchises, so I opened #179 to track that.

LynxAbraxas:
Tested this for a few more turns and seems to work as expected now: AI establishes Franchises and sues Franchises in its own city. Only observed though that the AI established one Franchise and sued one Franchise per turn even if it has mulitple Corporate Branches and Lawyers.
@MartinGuehmann Is the limit of one special order meant to be per army or per turn? I've alos never observed that the AI reformed more than one city per turn. However I can establish multiple Franchise and sue multiple Franchises per turn if I have multiple Armies for that.

LynxAbraxas:
@MartinGuehmann is the amount of special orders depending on other factors like wonders, government, civ size?
I can sue 3 (possilby all) Franchises as Persian with 3 Lawyers in the very same turn

ss_2019-07-25_11:44:57

ss_2019-07-25_11:43:40

but I can only sue 2 Franchises as Carthagian per turn

ss_2019-07-25_11:47:08

LynxAbraxas:
Just noticed that it was the lack of sufficent gold to execute the action. This case would deserve an info message. With enough gold, I can remove all Franchises, each one with its own Layer

ss_2019-07-25_12:28:45

However, the Sue or SueFranchise button is not disabled after the order was performed (as I would expect):

ss_2019-07-25_12:29:44

So I would conclude that the pre-test takes the available amount of gold into account, but somehow does not work appropriatly when there is enough gold (buttons allways enabled even after use) and if gold is lacking it depends on the position of the Lawer, only being correct if the Lawyer is inside a city.

LynxAbraxas:
> Just noticed that it was the lack of sufficent gold to execute the action. This case would deserve an info message. 

Possibly the cashier sound is meant to indicate that there is insufficient gold to execute the order...

LynxAbraxas:
> So I would conclude that the pre-test takes the available amount of gold into account, but somehow does not work appropriatly when there is enough gold (buttons allways enabled even after use) and if gold is lacking it depends on the position of the Lawer, only being correct if the Lawyer is inside a city.

The enabling of buttons is done here
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/ui/interface/UnitControlPanel.cpp#L865-L884
so I guess the test
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/ui/interface/UnitControlPanel.cpp#L865
reports something else before it would result in `ORDER_TEST_LACKS_GOLD`.

MartinGuehmann:
> Tested this for a few more turns and seems to work as expected now: AI establishes Franchises and sues Franchises in its own city. Only observed though that the AI established one Franchise and sued one Franchise per turn even if it has mulitple Corporate Branches and Lawyers.
> @MartinGuehmann Is the limit of one special order meant to be per army or per turn? I've alos never observed that the AI reformed more than one city per turn. However I can establish multiple Franchise and sue multiple Franchises per turn if I have multiple Armies for that.

Well, that's determined by the GoalElements in strategies.txt:

```
GoalElement { Goal GOAL_FRANCHISING         Priority   450000  MaxEval   5  MaxExec   3 }
GoalElement { Goal GOAL_SUE_FRANCHISE       Priority   450000  MaxEval   1  MaxExec 0.5 PerCity}
```

 So the default strategy the AI evaluates the best 5 franchise goals and executes the top 3. Of course, suing is done like defense, so one goal per city. Probably you could bump the MaxExec to 1. Actually, the values should match to the number of Lawyers and Corporate Branches it has.

MartinGuehmann:
> @MartinGuehmann is the use of the cheat-mode saved in some way or can I expect that a reloaded game does not know about its former use?

Let's put it this way, the game should remember whether you have used the cheat-mode. But I think it just doesn't.

LynxAbraxas:
> > @MartinGuehmann Is the limit of one special order meant to be per army or per turn? I've alos never observed that the AI reformed more than one city per turn. However I can establish multiple Franchise and sue multiple Franchises per turn if I have multiple Armies for that.
> 
> Well, that's determined by the GoalElements in strategies.txt:
> 
> ```
> GoalElement { Goal GOAL_FRANCHISING         Priority   450000  MaxEval   5  MaxExec   3 }
> GoalElement { Goal GOAL_SUE_FRANCHISE       Priority   450000  MaxEval   1  MaxExec 0.5 PerCity}
> ```
> 
> So the default strategy the AI evaluates the best 5 franchise goals and executes the top 3. Of course, suing is done like defense, so one goal per city. Probably you could bump the MaxExec to 1.

Thanks for that pointer. So the current 0.5 means the AI only decides to sue a franchise with a 50% chance? I.e. if it has one city with a Franchise and a Lawyer around it might decide to not sue the Franchise but possibly in the next turn (with a likelyhood of 1-0.5*0.5)? And with 1 it would do so right away?

> Actually, the values should match to the number of Lawyers and Corporate Branches it has.

Not sure what you mean by that.

LynxAbraxas:
> ss_2019-07-24_14:45:05

So I think in the case above (Lawyer inside the city, not enough gold for another special action) the button for
- Sue is enabled because the current pos is invalid and
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/ui/interface/UnitControlPanel.cpp#L865
leading to
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L10465-L10551
evalutated at the army's current postition and thereby returning `ORDER_TEST_NEEDS_TARGET`
- Sue Franchise is disabled because the test returns `ORDER_TEST_LACKS_GOLD`
- why injunction is disabled is not absolutely clear to but probalbly `ORDER_TEST_INVALID_TARGET` is returned because the current pos is tested and despite being a city it is invalide because it needs to belong to another civ

> ss_2019-07-24_14:45:19

In the case the Lawyer is outside the city:
- Injunction is enabled because the return is `ORDER_TEST_NEEDS_TARGET` which is retruned before `ORDER_TEST_LACKS_GOLD`
- Sue is disabled because  `ORDER_TEST_LACKS_GOLD`
- Sue Franchise is enabled because `ORDER_TEST_INVALID_TARGET` is returned

So the simplest solution is to move the amy in such a case, which might not be the most obvious solution. I was considering moving the test for `can_afford` to the beginning, but am worried that this has some unforseen side effects concerning the rest of the logic. The fact that the return variable is called `best_result` might indicate that some more complex logic was once planned instead of directly returning from the if statements.
For this PR I would leave it as is since it is not directly connected to AI Franchising and suing Franchises, which seems to works as expected. Since #179 is also not specific for this PR, I would say this can go in as is.
What do you think @MartinGuehmann?
Should we re-enable `useDefaultSuccessSound = true` for suing a franchise?

PS: What I do wonder is why the Sue button worked for Suing Franchises for a human player before with
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_data/default/gamedata/Orders.txt#L789

MartinGuehmann:
> Thanks for that pointer. So the current 0.5 means the AI only decides to sue a franchise with a 50% chance? I.e. if it has one city with a Franchise and a Lawyer around it might decide to not sue the Franchise but possibly in the next turn (with a likelyhood of 1-0.5*0.5)? And with 1 it would do so right away?

Well, I guess so but that is something I prefer to look at the code for.

> 
> > Actually, the values should match to the number of Lawyers and Corporate Branches it has.
> 
> Not sure what you mean by that.

The AI only builds a certain number of Corporate Branches, it only builds a certain number of Lawyers. So it does not make any sense to execute more goals of that type then the AI would ever build units for.

Then a special unit may be able to execute more than one type of special action, so you would like the values below the number of the units, so that the AI can use both special actions.

LynxAbraxas:
> > @MartinGuehmann is the use of the cheat-mode saved in some way or can I expect that a reloaded game does not know about its former use?
> 
> Let's put it this way, the game should remember whether you have used the cheat-mode. But I think it just doesn't.

OK, so I can use it for setting up some test cases and expect that the game evolves as if it had ended up in that situation naturally. I did notice though that in cheat-mode some things are not completely right, e.g. some armies are not displayed at the position where their field of view is drawn. It works find though if saving that state and re-loading it to test the evolution of the game from there on.

LynxAbraxas:
So if there is nothing missing in abc3337ca630c and a99b0306fb2 nor removing 8464392637cdec, this could go in.

MartinGuehmann:
> So the simplest solution is to move the amy in such a case, which might not be the most obvious solution. I was considering moving the test for `can_afford` to the beginning, but am worried that this has some unforseen side effects concerning the rest of the logic. The fact that the return variable is called `best_result` might indicate that some more complex logic was once planned instead of directly returning from the if statements.

I wouldn't worry so much about the name "best_result". They hadn't really a convention on how naming things. So the "best_result" could be just out of habit. And actually this variable is needed, at least for the internal logic of the function. There is a difference between the stack contains no unit that can execute the special action and it only contains units that can perform the special action but have already used it and the stack contains at least one unit that can execute the special action right now. The question is whether that is still relevant outside of the function.

Anyway, I wouldn't worry too much about what was planned. Either we see a reason to add something or not. If not and the design is more complicated then necessary then it should be simplified.

> What do you think @MartinGuehmann?
> Should we re-enable `useDefaultSuccessSound = true` for suing a franchise?

Yes, once you have sued a franchise you should her that you have done it. For most of the others the special effects and sounds are handled elsewhere, but not for suing a franchise.
 
> PS: What I do wonder is why the Sue button worked for Suing Franchises for a human player before with
> https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_data/default/gamedata/Orders.txt#L789

That has to do with the code you outcommented in ArmyData::Sue(const MapPoint &point), it just expects a MapPoint, it doesn't matter what is there. If you looked carefully, you could have noticed that path to that position became red, indicating an invalid action.

MartinGuehmann:
> So if there is nothing missing in [abc3337](https://github.com/civctp2/civctp2/commit/abc3337ca630caf1aa08a5a5492a6ef5c84ffa41) and [a99b030](https://github.com/civctp2/civctp2/commit/a99b0306fb2cb9b880e874f609d38cdfecb1340f) nor removing [8464392](https://github.com/civctp2/civctp2/commit/8464392637cdece50dd7658b7ae13098e33bba05), this could go in.

Well, just as I said, just enable the default sound. (I just repeat here, since it is not buried under the other stuff.)

LynxAbraxas:
> Well, just as I said, just enable the default sound. (I just repeat here, since it is not buried under the other stuff.)

Done, finally something simple again... ;-)

LynxAbraxas:
> If you looked carefully, you could have noticed that path to that position became red, indicating an invalid action.

I see, so it was executed despite the pre-test failing.

MartinGuehmann:
> > Well, just as I said, just enable the default sound. (I just repeat here, since it is not buried under the other stuff.)
> 
> Done, finally something simple again... ;-)

And passed the tests, so it can go in.

PR #177, opened 2019-07-24
make Diplomats un-expelleable, which strengthens AI

LynxAbraxas:
This PR makes Diplomats un-expelleable, which seems to strengthen the AI due to diplomatic effects. For example the AI code seems to take into account the info gained from an embassy, resulting in e.g. more advance exchange, which over all helps each AI civ to get better, which in turn makes the game more challenging and therefore interesting. While I'm not fully sure about my preference concerning #103 (which inspired this PR) it seems quite appropriat to make Diplomats immune against expelling, since they are very vulnerable and no stealth unit and it is can be problematic to accompany them with a military unit. While I find it odd that a Lawer or Coorporate Branch can block a road or railway for a 12 unit army, this does not seems so unrealistic for a Diplomat. Independen of that, immune Diplomats seem to hep the AI to be a more challenging opponent.

### DIFF
--- a/ctp2_data/default/gamedata/Units.txt
+++ b/ctp2_data/default/gamedata/Units.txt
@@ -1279,7 +1279,6 @@ UNIT_DIPLOMAT {
    MaxFuel 0
    IgnoreZOC
    NoZoc
-   CanBeExpelled
    CantCaptureCity
    EstablishEmbassy
    ThrowParty


MartinGuehmann:
So then if we keep Diplomats expellable then Lawyers and Corporate Branches must be expellable, too. Otherwise it wouldn't make any sense, because both are civilists. Well, at least officially. Diplomats have immunity, but in reality you can kick them out.

My problem is that I have no strong opinion about Lawyers and Corporate Branches, either. But logically, why should you be able to kick out a spy, but you cannot kick out a Lawyer that doesn't really make any sense.

Sure as alternative you can sue Lawyers and Corporate Branches this way they won't return.

LynxAbraxas:
> So then if we keep Diplomats expellable then Lawyers and Corporate Branches must be expellable, too. 

When relating to reality, yes, but not necessarily for the game logic.

> But logically, why should you be able to kick out a spy, but you cannot kick out a Lawyer that doesn't really make any sense.
> Sure as alternative you can sue Lawyers and Corporate Branches this way they won't return.

That's why I think Lawyers and Coroporate Branches should be expellable, at least for civs that cannot built Lawyers yet.
However, I find there is a difference in the game between a Diplomat and other special units like Lawyers, Spys and Coroporate Branches: A Diplomat does no harm and is not invisible to normal units, which leads to many expells while a Diplomat succeeding strengthens the AI and therefore makes the game as is more intersting.
So another option to disable expelling for Diplomats would be to make the invisible for normal units.

MartinGuehmann:
> So another option to disable expelling for Diplomats would be to make the invisible for normal units.

Reality and game logic collide here. In Civ2 I am not sure whether you could see Diplomats for sure in Civ1. In Civ2 they can also make cities revolt and join your civ. So Civ2 Diplomats were also spies and in fact were also replaced in the later game by a spy. So these Diplomats should be invisible. In CTP2 diplomats can not make cities revolt and join your civ, they cannot steal technology, but what they can do is exploring your territory.

So from the game logic I tend to like them invisible to normal units, but then you can still kick them out if you have other units around. The alternative would be to make them suable, so you could still get rid of them.

Or we just keep them unexpellable, probably none of the alternatives are without downsides. Be it game logic or reality.

LynxAbraxas:
> The alternative would be to make them suable, so you could still get rid of them.

That, I think, would not help much as Diplomats turn up far earlier than Lawyers, so that would not help in the earlier stages of  a game. I'd go for either be invisible to normal units or un-expellable.

MartinGuehmann:
Well, then have them unexpellable and merge this.

LynxAbraxas:
I think then #103 should go in as well. Handling that more specifically, i.e. depending on wether a civ can built Lawyer or not will not be so simple.

MartinGuehmann:
> I think then #103 should go in as well. Handling that more specifically, i.e. depending on wether a civ can built Lawyer or not will not be so simple.

In principle, it devalues suing. But eventually we have to decide and not all ideas of Activision are necessarily the best. So then it goes in.

LynxAbraxas:
> In principle, it devalues suing.

Well, suing destroys, expelling just sents the unit home. I'll try to think of a way to realize that expelling is only possible for civs that cannot yet built Lawyers, that way we would have both concepts without weakening slow AI civs badly.

PR #178, opened 2019-07-24
Message adjustments

LynxAbraxas:
This PR adjust and removes some messages, reasons given in each commit message.


LynxAbraxas:
I've never seen this city-report or alike
https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L2196
It would correspond to the disabled injunction message (https://github.com/civctp2/civctp2/pull/178/commits/5f24cb87030c016027302c86c5a9cd6bd6e7965c) and I think it is obsolete since a successful Franchise results in a Franchise city icon.

MartinGuehmann:
> I've never seen this city-report or alike
> https://github.com/civctp2/civctp2/blob/82a15c92ee2b85adf01e56f89e334a15e8926dbb/ctp2_code/gs/gameobj/ArmyData.cpp#L2196
> 
> 
> It would correspond to the disabled injunction message ([5f24cb8](https://github.com/civctp2/civctp2/commit/5f24cb87030c016027302c86c5a9cd6bd6e7965c)) and I think it is obsolete since a successful Franchise results in a Franchise city icon.

I guess the reason why you have never seen it is that the AI didn't/doesn't do it.

LynxAbraxas:
> I guess the reason why you have never seen it is that the AI didn't/doesn't do it.

Ah, I missed the fact that is for the victim (not the agressor), now pops up as expected:

ss_2019-07-25_09:58:41

ss_2019-07-25_10:51:43

LynxAbraxas:
Should the message that an embassy was installed be removed as well?
ss_2020-01-20_22:18:14

LynxAbraxas:
There haven't been any new commits here for a long time and so far I do not have any other message adjustment on my list. @MartinGuehmann ready for merging?

MartinGuehmann:
Good, then we can merge.

Issue #53, opened 2018-12-13
Turn numbers along assigned path not visible with SDL


LynxAbraxas:
When a unit is given a target with the mouse pointer the path to take is shown also with number fields but the turn numbers are missing (possibly an SDL problem, master under linux).

MartinGuehmann:
So this is now solved and can be closed.

PR #107, opened 2019-02-25
Debug version with GL-CI

LynxAbraxas:
This PR offers changes to compile a debug version with GL-CI. 
This PR could either be merged into a separate branch from master or adjusted such that the debug build and test jobs are added to the normal jobs. This however would about double the use of GL-CI minutes of shared runners (but not much the overall test time because the jobs can run in parallel independently). 
If a debug build and tests are not needed for every commit to master, a separate branch might be more apropriate. However, then merging in master for testing will be neede to trigger the GL-CI pipeline.
The current artifacts contain the `logs/`https://github.com/civctp2/civctp2/commit/254704d1dbbbc655697df1331bad0a6d2ee7713b but not `CTP_LEAKS_99999.TXT`, `CTP_LEAKS_ALT_99999.TXT` (good for detecting memory leaks: https://github.com/civctp2/civctp2/pull/97#issuecomment-465246916).

PR #182, opened 2019-07-26
Debug path turn numbers for SDL

LynxAbraxas:
In relation to #53, with this PR path turn numbers are also drawn when using SDL (Linux).

### DIFF
--- a/ctp2_code/gfx/tilesys/TileHighlight.cpp
+++ b/ctp2_code/gfx/tilesys/TileHighlight.cpp
@@ -53,6 +53,7 @@
 #include "controlpanelwindow.h"
 #include "OrderRecord.h"
 #include "aui_surface.h"
+#include "aui_bitmapfont.h"
 #include "maputils.h"
 #include "primitives.h"
 #include "tiledmap.h"
@@ -804,7 +805,8 @@ void TiledMap::DrawLegalMove
 						sint32 const	textX	= x - (width >> 1);
 						sint32 const	textY	= y - (height >> 1);
 
-						primitives_DrawText(pSurface, textX, textY, turnNumber, 0, 1);
+						RECT clipRect = primitives_GetScreenAdjustedRectCopy(pSurface, turnRect);
+						m_font->DrawString(pSurface, &turnRect, &clipRect, turnNumber, 0, g_colorSet->GetColorRef(COLOR_BLACK), 0);
 					}
 				}
 			}
@@ -856,7 +858,8 @@ void TiledMap::DrawLegalMove
 					sint32 const	textX	= x - (width >> 1);
 					sint32 const	textY	= y - (height >> 1);
 
-					primitives_DrawText(pSurface, textX, textY, turnNumber, 0, 1);
+					RECT clipRect = primitives_GetScreenAdjustedRectCopy(pSurface, turnRect);
+					m_font->DrawString(pSurface, &turnRect, &clipRect, turnNumber, 0, g_colorSet->GetColorRef(COLOR_BLACK), 0);
 				}
 			}
 		}
@@ -1113,7 +1116,8 @@ void TiledMap::DrawUnfinishedMove(aui_Surface * pSurface)
 						sint32 textX = x - (width>>1);
 						sint32 textY = y - (height>>1);
 
-						primitives_DrawText(pSurface, textX, textY, turnNumber, color, 1);
+						RECT clipRect = primitives_GetScreenAdjustedRectCopy(pSurface, turnRect);
+						m_font->DrawString(pSurface, &turnRect, &clipRect, turnNumber, 0, color, 0);
 					}
 				}
 			}
--- a/ctp2_code/gfx/tilesys/TileHighlight.cpp
+++ b/ctp2_code/gfx/tilesys/TileHighlight.cpp
@@ -800,13 +800,15 @@ void TiledMap::DrawLegalMove
 						primitives_PaintRect16(pSurface, &turnRect, g_colorSet->GetColor(actual_line_color));
 						primitives_FrameRect16(pSurface, &turnRect, 0);
 
-						sint32 const	width	= textutils_GetWidth(pSurface, turnNumber);
-						sint32 const	height	= textutils_GetHeight(pSurface, turnNumber);
-						sint32 const	textX	= x - (width >> 1);
-						sint32 const	textY	= y - (height >> 1);
+						sint32 const width = m_font->GetStringWidth(turnNumber);
+						sint32 const height = m_font->GetMaxHeight();
+						RECT rect = {0, 0, width, height};
+						OffsetRect(&rect, // position rect relative to turnRect
+						    turnRect.left + (turnRect.right - turnRect.left) / 2 - width  / 2,
+						    turnRect.top  + (turnRect.bottom - turnRect.top) / 2 - height / 2);
 
-						RECT clipRect = primitives_GetScreenAdjustedRectCopy(pSurface, turnRect);
-						m_font->DrawString(pSurface, &turnRect, &clipRect, turnNumber, 0, g_colorSet->GetColorRef(COLOR_BLACK), 0);
+						RECT clipRect = primitives_GetScreenAdjustedRectCopy(pSurface, rect);
+						m_font->DrawString(pSurface, &rect, &clipRect, turnNumber, 0, g_colorSet->GetColorRef(COLOR_BLACK), 0);
 					}
 				}
 			}
@@ -853,10 +855,12 @@ void TiledMap::DrawLegalMove
 					primitives_PaintRect16(pSurface, &turnRect, g_colorSet->GetColor(actual_line_color));
 					primitives_FrameRect16(pSurface, &turnRect, 0);
 
-					sint32 const	width	= textutils_GetWidth(pSurface, turnNumber);
-					sint32 const	height	= textutils_GetHeight(pSurface, turnNumber);
-					sint32 const	textX	= x - (width >> 1);
-					sint32 const	textY	= y - (height >> 1);
+					sint32 const width = m_font->GetStringWidth(turnNumber);
+					sint32 const height = m_font->GetMaxHeight();
+					RECT rect = {0, 0, width, height};
+					OffsetRect(&rect, // position rect relative to turnRect
+					    turnRect.left + (turnRect.right - turnRect.left) / 2 - width  / 2,
+					    turnRect.top  + (turnRect.bottom - turnRect.top) / 2 - height / 2);
 
 					RECT clipRect = primitives_GetScreenAdjustedRectCopy(pSurface, turnRect);
 					m_font->DrawString(pSurface, &turnRect, &clipRect, turnNumber, 0, g_colorSet->GetColorRef(COLOR_BLACK), 0);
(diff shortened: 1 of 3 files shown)


MartinGuehmann:
I guess you have to include aui_bitmapfont.h in TileHighlight.cpp to make it compile.

LynxAbraxas:
Let's see, I'm not sure because there is already
https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/gfx/tilesys/tiledmap.h#L99

LynxAbraxas:
It seems `TileHighlight.cpp` is basically just an extension of `tiledmap.cpp`.

MartinGuehmann:
> Let's see, I'm not sure because there is already
> https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/gfx/tilesys/tiledmap.h#L99

Well, that's a forward declaration, it tells the compiler that this exists, but it is implemented in another translation unit. Basically, this way you can use it to declare things, without needing to include it in header files, which would blow up the translation unit and eventually also the final executable.

LynxAbraxas:
True, got confused by the various header files I looked at.

Do we need SDL implementations for these as well
https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/ui/aui_utils/primitives.h#L75-L99
at least some appear in the current code base but possibly in unused parts. Except for the path numbers I have not missed any text so far.

MartinGuehmann:
> True, got confused by the various header files I looked at.
> 
> Do we need SDL implementations for these as well
> https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/ui/aui_utils/primitives.h#L75-L99
> 
> 
> at least some appear in the current code base but possibly in unused parts. Except for the path numbers I have not missed any text so far.

Well, that depends if they are used, actually the debug version is supposed to print quite some text, while loading and loading a save game for the first time in the current session. So there might be something missing. Otherwise, if the stuff is indeed not used then we should get rid of it.

From a shot look the code seems also be copy'n'paste, which isn't really a good design.

LynxAbraxas:
Looks much better now.
Before

ss_2019-07-27_23:42:51

Now

ss_2019-07-27_23:37:15

LynxAbraxas:
> True, got confused by the various header files I looked at.
> 
> Do we need SDL implementations for these as well
> 
> https://github.com/civctp2/civctp2/blob/20d97537dfca686a8bdad7d2c6e1c69a19f2a3a8/ctp2_code/ui/aui_utils/primitives.h#L75-L99
> 
> at least some appear in the current code base but possibly in unused parts. Except for the path numbers I have not missed any text so far.

Some basic implementations for these are now included.
As far as I could test this, this seems to work, so if the implementations are acceptable, good to go in.

LynxAbraxas:
To avoid dublication of code in this case where there is no class, I used to have
```
#ifndef __AUI_USE_DIRECTX__
inline AUI_ERRCODE getBitmapFont(aui_BitmapFont *m_font)
    {
    AUI_ERRCODE      errcode     = AUI_ERRCODE_OK;
    aui_StringTable	*stringTable = new aui_StringTable(&errcode, "TiledMapFontStringTable");

    if (AUI_NEWOK(stringTable, errcode))
	{
	const MBCHAR *    fontNameString  = stringTable->GetString(0);
	const MBCHAR *    fontSizeString  = stringTable->GetString(1);
	
	m_font = g_c3ui->LoadBitmapFont(fontNameString);
	Assert(m_font);
	m_font->SetPointSize(atoi(fontSizeString));
	
	// const MBCHAR *    fString         = stringTable->GetString(2);
	// strncpy(m_fortifyString, fString, 3);
	}
    
    delete stringTable;

    return errcode;
    }
#endif
```
and then used that like

```
#else	
	aui_BitmapFont *m_font;
	getBitmapFont(m_font);
	    
	RECT clipRect = primitives_GetScreenAdjustedRectCopy(pDirectSurface, *destRect);
	m_font->DrawString(pDirectSurface, destRect, &clipRect, pString, k_AUI_BITMAPFONT_DRAWFLAG_VERTCENTER, textColor, 0); // no bg correspondence, not handling dropColor
#endif // __AUI_USE_DIRECTX__
```

which segfaulted. It is unclear to me why. Is what `LoadBitmapFont` allocates lost when returning from `getBitmapFont` for some reason?

MartinGuehmann:
The thing may segfault if AUI_NEWOK returns false, then m_font wouldn't be initialized. Anyway why is the thing called m_font even so it is not a member variable.

I would also make getBitmapFont return the aui_BitmapFont pointer. I don't know why the error code is needed, especially if it is always AUI_ERRCODE_OK. Anyway, initialize the aui_BitmapFont pointer with NULL if AUI_NEWOK returns false.

LynxAbraxas:
Some of the newly implemented SDL functions are also used e.g. in the save window for annotating the info power graph with the number of turns:

ss_2019-07-28_17:54:09

LynxAbraxas:
OK with https://github.com/civctp2/civctp2/pull/182/commits/c01c6685e17d64d080b0c7ccef0bad6fba72fea9 no segfaults any more but also no font in e.g.  the save window (as in the screen shot above with bff7eaa).
@MartinGuehmann  any idea what could be wrong?

MartinGuehmann:
> OK with [c01c668](https://github.com/civctp2/civctp2/commit/c01c6685e17d64d080b0c7ccef0bad6fba72fea9) no segfaults any more but also no font in e.g. the save window (as in the screen shot above with [bff7eaa](https://github.com/civctp2/civctp2/commit/bff7eaaedfc5d9223305f3b70c688b6a0eae6491)).
> @MartinGuehmann any idea what could be wrong?

Well, in the screenshot I only see that the text is not at the right place for everything else you have font. For the misplaced turn annotation, I would go to the powergraph directly. In fact, over the course of testing, the pwergraph dimensions differ a bit. Even so, the code was never changed on purpose.

LynxAbraxas:
> Well, in the screenshot I only see that the text is not at the right place for everything else you have font. 

The problem is that the font is only there with bff7eaa but its gone with c01c668.

MartinGuehmann:
> > Well, in the screenshot I only see that the text is not at the right place for everything else you have font.
> 
> The problem is that the font is only there with [bff7eaa](https://github.com/civctp2/civctp2/commit/bff7eaaedfc5d9223305f3b70c688b6a0eae6491) but its gone with [c01c668](https://github.com/civctp2/civctp2/commit/c01c6685e17d64d080b0c7ccef0bad6fba72fea9).

Well, I would like to have the function like this anyway:

inline aui_BitmapFont * getBitmapFont()

This way it is more easier to understand. And I think it is also your problem: You pass the pointer by value and not by address. So whatever you do inside with it doesn't influence it outside.

Actually, I would have to look up the syntax for passing the pointer by address. So better don't do it if you absolutely must.

LynxAbraxas:
> You pass the pointer by value and not by address. 

Many thanks for that hint. True, so it would have worked with `getBitmapFont(aui_BitmapFont **font)` and `getBitmapFont(&font)` but changed according to your recommendation now, which works well now even with https://github.com/civctp2/civctp2/pull/182/commits/2770814e6f55a2f000917eae3b37efe2f32e0fc6. So good to go in.

MartinGuehmann:
> > You pass the pointer by value and not by address.
> 
> Many thanks for that hint. True, so it would have worked with `getBitmapFont(aui_BitmapFont **font)` and `getBitmapFont(&font)` but changed according to your recommendation now, which works well now even with [2770814](https://github.com/civctp2/civctp2/commit/2770814e6f55a2f000917eae3b37efe2f32e0fc6). So good to go in.

Looks fine and thus can go in.

LynxAbraxas:
Thanks. 
BTW some tests failed in the past on GL (e.g. https://gitlab.com/civctp2/civctp2/pipelines/73307870) which is probably due to the descussed problem that some sikuli events do not get to the application if the load on the CPU is high. You could give it a try and just hit the "retry" button in that case. The corresponding commits passed all tests on my GL.

MartinGuehmann:
So retried it and it worked. There also seems to be some more, but most of them concerning building a city. It would be good if I would get an email if one test fails. That would draw my attention.

LynxAbraxas:
> There also seems to be some more, but most of them concerning building a city.

Probably for the same reason and not at all related to what was actually tested.

> It would be good if I would get an email if one test fails. That would draw my attention.

Hm, I get an email if a pipeline fails. Perhaps you need to specify what email adress should be used in that case.

MartinGuehmann:
> Hm, I get an email if a pipeline fails. Perhaps you need to specify what email adress should be used in that case.

It's just the question where. Maybe in the GitLab profile? I created it via GitHub that could be a problem.

MartinGuehmann:
OK, set my global notification level on GitLab to Custom, where failed pipeline is ticked.

LynxAbraxas:
Hm, some of the changes of the PR seem to cause a segfault in the debug version:

```
Thread 1 "ctp2" received signal SIGSEGV, Segmentation fault.
0x0000555555fb4770 in std::vector<StringRecord*, std::allocator<StringRecord*> >::operator[] (this=0x18, __n=219) at /usr/include/c++/5/bits/stl_vector.h:795
795	      { return *(this->_M_impl._M_start + __n); }
#0  0x0000555555fb4770 in std::vector<StringRecord*, std::allocator<StringRecord*> >::operator[] (this=0x18, __n=219)
    at /usr/include/c++/5/bits/stl_vector.h:795
#1  0x0000555555fb356b in StringDB::GetHead (this=0x0, 
    id=0x55555b82481c "str_tbl_ldl_arialbd_ttf") at StrDB.cpp:140
#2  0x0000555555fb3939 in StringDB::GetText (this=0x0, 
    get_id=0x55555b82481c "str_tbl_ldl_arialbd_ttf", new_text=0x7fffffffe200)
    at StrDB.cpp:278
#3  0x0000555555fb3db9 in StringDB::GetNameStr (this=0x0, 
    s=0x55555b82481c "str_tbl_ldl_arialbd_ttf") at StrDB.cpp:476
#4  0x00005555558347e4 in aui_StringTable::aui_StringTable (
    this=0x55555b855e3c, retval=0x7fffffffe694, 
    ldlBlock=0x555556337f68 "TiledMapFontStringTable")
    at ui/aui_common/aui_stringtable.cpp:89
#5  0x00005555558582d2 in getBitmapFont () at ui/aui_utils/primitives.h:76
#6  0x0000555555852474 in primitives_DrawText (pDirectSurface=0x55555b83dccc, 
    x=325, y=18, pString=0x555556b933c0 <g_splash_buf> " 519154.70 secs  ", 
    color=16317688, bg=true) at ui/aui_utils/primitives.cpp:1152
#7  0x000055555599124d in Splash::AddTextNL (this=0x55555743027c, 
    text=0x555556b933c0 <g_splash_buf> " 519154.70 secs  ")
    at ui/interface/splash.cpp:68
#8  0x0000555555774dd0 in ui_Initialize () at ctp/civ3_main.cpp:431
#9  0x000055555577d6d8 in CivApp::InitializeApp (this=0x555556cac61c, 
    hInstance=0x0, iCmdShow=2) at ctp/civapp.cpp:1285
#10 0x000055555577709d in CivMain (iCmdShow=2, pSzCmdLine=0x7fffffffecd8)
    at ctp/civ3_main.cpp:1682
#11 0x0000555555776c6d in main (argc=2, argv=0x7fffffffecd8)
    at ctp/civ3_main.cpp:1441
```

@MartinGuehmann do you have an idea what might be the problem? I'm stuck on getting the debug-version up and running again for over 7 nights now...

LynxAbraxas:
The problem occurs with the debug strings draw for the splash called from `ui_Initialize();`
https://github.com/civctp2/civctp2/blob/55b706d1eed95c20ca71b2b2e54952d7812d4d68/ctp2_code/ctp/civapp.cpp#L1285
and I think the source for the problem is that the `StringDB` is not yet initialized then. However moving its initialization from
https://github.com/civctp2/civctp2/blob/55b706d1eed95c20ca71b2b2e54952d7812d4d68/ctp2_code/ctp/civapp.cpp#L1316
before
https://github.com/civctp2/civctp2/blob/55b706d1eed95c20ca71b2b2e54952d7812d4d68/ctp2_code/ctp/civapp.cpp#L1285
is likely going to cause other problems.

LynxAbraxas:
Fixing things for the debug version in #107, discussion to be continued there.

PR #183, opened 2019-07-28
Unify the LDL files into one directory

MartinGuehmann:
This pull request removes the duplicated LDL files from the various language folders and puts them all under the default data subfolder. This we only one file has to be updated when modifying LDL files. And this also restores the Windows version on all languages.


LynxAbraxas:
Looks good to me, tested and seems to cause no changes even for Linux/SDL version.
Many thanks for that clean-up @MartinGuehmann!

MartinGuehmann:
> Looks good to me, tested and seems to cause no changes even for Linux/SDL version.
> Many thanks for that clean-up @MartinGuehmann!

Well, then let's merge.

Issue #129, opened 2019-05-01
Clerics and televangelists also for other governments


LynxAbraxas:
Issue for discussing what was started in https://github.com/civctp2/civctp2/issues/112#issuecomment-488147133.

LynxAbraxas:
One option to allow clerics and televangelists to other govs while keeping theocracy as an interesting gov would be to make clerics and televangelists cheeper to built in theocracy or limiting their numbers or their success in other govs. @MartinGuehmann how much effort would it be to implement such?
A pro for clerics and televangelists to be available to other govs would be that it would surely make the game more interesting.

MartinGuehmann:
> One option to allow clerics and televangelists to other govs while keeping theocracy as an interesting gov would be to make clerics and televangelists cheeper to built in theocracy or limiting their numbers or their success in other govs. @MartinGuehmann how much effort would it be to implement such?

I don't think it is too much, E put in this Government modified stuff. The only thing is to implement it properly. It requires quite some repetitive code changes. The alternative is to add a government specific discount.

> A pro for clerics and televangelists to be available to other govs would be that it would surely make the game more interesting.

Sure it would make the game more interesting. And of course to make Theocracy more interesting Fascism should indeed come later. So we can already agree on this.

LynxAbraxas:
> The alternative is to add a government specific discount.

How about adding a gov discount and an increased chance of success of their actions?

> Fascism should indeed come later. So we can already agree on this.

I would welcome such a change.

LynxAbraxas:
Possibly part of #186: https://github.com/civctp2/civctp2/pull/186/files#diff-7e6597dac40622ec01209766e48d9c92

LynxAbraxas:
> Sure it would make the game more interesting. And of course to make Theocracy more interesting Fascism should indeed come later. So we can already agree on this.

Hm, could that be realized by just re-ordering here:
https://github.com/civctp2/civctp2/blob/94e4ff7070cf1503909558151467dc993957bb31/ctp2_data/default/aidata/strategies.txt#L126-L129

MartinGuehmann:
> > Sure it would make the game more interesting. And of course to make Theocracy more interesting Fascism should indeed come later. So we can already agree on this.
> 
> Hm, could that be realized by just re-ordering here:
> https://github.com/civctp2/civctp2/blob/94e4ff7070cf1503909558151467dc993957bb31/ctp2_data/default/aidata/strategies.txt#L126-L129

[...]

PR #186, opened 2019-08-11
Fix the AI4

MartinGuehmann:
Add more AI improvements:

The AI now improve sea tiles if there is already a tunnel on.
The AI only connects cities within a radius of 3 times the empire radius around cities, to reduce the run time with many cities.
The AI now assigns a rally agent, even if all possible rally agents are in full cities. In this case the AI can rally for city, too.

Some additional code simplification, removing a debug assert for a valid situation, fixed a crash. And more debug reports for debugging slider optimization and its run time.


ctp2_data/default/gamedata/Units.txt
```diff
@@ -879,7 +879,7 @@ UNIT_CLERIC {
    EnableAdvance ADVANCE_THEOLOGY
    ObsoleteAdvance ADVANCE_MASS_MEDIA
    UpgradeTo UNIT_TELEVANGELIST
-   GovernmentType GOVERNMENT_THEOCRACY
+#   GovernmentType GOVERNMENT_THEOCRACY
```
LynxAbraxas: This change seems to be introduced with 6a57d7a5208ee32c8337bace478a217094448c7d but does not originate from either parent of the merge. Unless unavoidable, this is very bad! There seem to be other changes so I created a plain merge (https://github.com/LynxAbraxas/civctp2/commit/6c51019793152e4c18b1dfdf9c2495dc6611ea6a) and a commit with the extra changes (https://github.com/LynxAbraxas/civctp2/commit/523f586adda0e96a0e5c6ed213c6b835e9a93b0e). Since even these do not seem related I'd ask you to make the effort of splitting the changes into their own meaningful commits. Otherwise the changes are hardly trackable nor understandable. For example I'd make the changes concerning Clerics and Televangelists in all govs a commit on its own, even if it is just 2 lines commented. This makes a big change and it is very helpful for future development if this is in its own commit with a good commit message, that contains a link to the discussion we had on this on GH, e.g. #129. 
Though I am trying hard to do good commit messages, I know they are not always as meaningful to others, for which @DEVoytas  pointed me to (https://github.com/civctp2/civctp2/pull/14#commitcomment-31035813):
https://chris.beams.io/posts/git-commit/
https://github.com/erlang/otp/wiki/writing-good-commit-messages :)

MartinGuehmann: Thanks for pointing that out and for the clean merge commit.

Indeed 6a57d7a is quite messy, I have to see whether TortoiseGit, has a clear interface for it. And whether the mistake comes from there. Basically, the interface should show me somewhere the changes that do not come from the parents. So that means I have to clean up, here.

Basically, these changes are supposed for debugging or experimental. The extra changes of [LynxAbraxas@523f586] are not supposed to be in. So far I was aware of ctpai.cpp and civ3_main.cpp, these changes for debugging, and if it were those alone, I could still think about to leave them alone, but the other stuff is not supposed to be in.

Actually, units.txt is not even a commit of its own but also a pull request, since it changes the game play.

directoractions.h is just something outcommented, I had in the working copy. It is supposed to fix a memory leak, but it makes the game crash.

Then we have strategies.txt, that are 2 or 3 commits, one of them goes into its own PR, since I still don't know whether that is the best to handle it, and it also has a big effect.

In govern.txt, I increased the city limit to see whether the AI can conquer the whole world, that still doesn't work. It still has some trouble with underwater cities. So that is something for debugging, but not for this PR.

Same with units.txt, whether the AI can use Clerics properly, is within the scope of this PR, but not whether it can build it.

LynxAbraxas: > Thanks for pointing that out and for the clean merge commit.

Just noticed that I also screwed up the merge commit message:
```
Fixed conflicts in:
    ../ctp2_code/gs/gameobj/CityData.cpp
    with
    git checkout 6a57d7a5208ee32c8337bace478a217094448c7d -- ctp2_code/gs/gameobj/CityData.cpp
```
Should be in the message for https://github.com/LynxAbraxas/civctp2/commit/6c51019793152e4c18b1dfdf9c2495dc6611ea6a

> I have to see whether TortoiseGit, has a clear interface for it. And whether the mistake comes from there. Basically, the interface should show me somewhere the changes that do not come from the parents.

Never used TortoiseGit so far, so can't help you there, though simple `gitk` from the git-package isn't very fancy, it does show this quite nicely:

6a57d7a:
ss_2019-08-30_14:24:51

in contrast for https://github.com/LynxAbraxas/civctp2/commit/6c51019793152e4c18b1dfdf9c2495dc6611ea6a `CityData.cpp` is the only one with changes from both parents, i.e. had a merge conflict:
ss_2019-08-30_14:26:02

> Actually, units.txt is not even a commit of its own but also a pull request, since it changes the game play.

Yes, I would welcome a PR on its own here, too.

> Same with units.txt, whether the AI can use Clerics properly, is within the scope of this PR, but not whether it can build it.

Not sure if I really get here what you mean. I'll test this PR with this to see if conversions by the AI happen or if there are any bugs like there were for franchise and sue franchise.  Before this PR is merged I'd say the changes to units.txt should be reverted (which is just a `git revert SHA` if the change were in its own commit), and a new PR just concerning clerics/televangelists should be used to discuss how to best realize this.

MartinGuehmann: > Just noticed that I also screwed up the merge commit message:
> 
> ```
> Fixed conflicts in:
>     ../ctp2_code/gs/gameobj/CityData.cpp
>     with
>     git checkout 6a57d7a5208ee32c8337bace478a217094448c7d -- ctp2_code/gs/gameobj/CityData.cpp
> ```

That should have gone into the merge commit, at least the message says merge commit. Since I have to tear apart this commit, I won't use that commit and message anyway.

MartinGuehmann: Indeed TortoiseGit is not so nice a bout merge commits, it shows me the files by parent, but actually the changes that come from no parent are shown as they come from both parents. Even so they are supposed to be in an extra section.

And also when I committed the merge commit, I could decide what file goes in, but I supposed that it would not put in the files that come from no parent.

MartinGuehmann:
Rewrote history, now it should only contain the stuff that is supposed to be in.

LynxAbraxas:
Great, will take a look, but probably not before the 21st of Sep.

LynxAbraxas:
Looks goot me now, though I understand only some of the changes.
Play-tested this for some time now and had no unexpected effects. 
I do notice that the AI quite often starts war by killing a corporate branch that stands in its way (similar to what I once reported abut clerics). I also observed that the AI expells a corporate branch so I'm not sure if meeting the corporate branch actually is the little bit over the top to decide to start a war or if that is a bug.

MartinGuehmann:
> Looks goot me now, though I understand only some of the changes.

What are the ones you don't understand?

> Play-tested this for some time now and had no unexpected effects.
> I do notice that the AI quite often starts war by killing a corporate branch that stands in its way (similar to what I once reported abut clerics). I also observed that the AI expells a corporate branch so I'm not sure if meeting the corporate branch actually is the little bit over the top to decide to start a war or if that is a bug.

Actually, these AIs desire war with the other civs so they will just run over these special units. Actually, it could made so that only the militarists would run over the these units while they others have to declare war explicitly, before.

LynxAbraxas:
> What are the ones you don't understand?

No offense, I mean I understand them only in general, not in full detail.

> Actually, it could made so that only the militarists would run over the these units while they others have to declare war explicitly, before.

That would be really interesting, either fixed to specific nations or randomly determined at the start of a new game.

LynxAbraxas:
Perhaps something fitting to this PR: 
So far I never observed that an AI lawer ever sued a corporate branch or other lawer.
Neither does the AI seem to pillage improvements of a civ it is at war with.

MartinGuehmann:
> > What are the ones you don't understand?
> 
> No offense, I mean I understand them only in general, not in full detail.

Well, I just mean that you can ask.

> > Actually, it could made so that only the militarists would run over the these units while they others have to declare war explicitly, before.
> 
> That would be really interesting, either fixed to specific nations or randomly determined at the start of a new game.

Well, that would be actually fixed to the leader.

MartinGuehmann:
> Perhaps something fitting to this PR:

Sure that fits.

> So far I never observed that an AI lawer ever sued a corporate branch or other lawer.

Indeed that is not in. The lawyers rather prefer to stop production in cities.

> Neither does the AI seem to pillage improvements of a civ it is at war with.

Actually, the AIs do maybe not so often as you would like, they are more busy with conquering and in that case you shouldn't destroy the improvements of your future cities.

I think it happened that some AI land units pillaged an undersea tunnel, not a good idea if you are using it.

LynxAbraxas:
> Indeed that is not in. The lawyers rather prefer to stop production in cities.

Is that just a much higher priority or is the AI missing the rules for suing other corporate branches or other lawer completely?

> Actually, the AIs do maybe not so often as you would like, they are more busy with conquering and in that case you shouldn't destroy the improvements of your future cities.

True, I guess the AI code needs some estimation of how likely the AI is able to conquer a city and if that is quite unlikely pillaging should be used, but that is probably some bigger effort.

> I think it happened that some AI land units pillaged an undersea tunnel, not a good idea if you are using it.

True, not good at all. It might be possible to train a neural network to play the game. If that would work out, tuning the AI logic wouldn't be necessary any more but probably playing the game over the network would need some debugging.

LynxAbraxas:
I also noticed that the AI does not seem to repair its units, even if they are on the verge to die. Does the AI have a rule for that that just might need a higher priority?

LynxAbraxas:
With #171 I noticed that the AI apparently does not priorise which of its converted cities should be reformed first, i.e. the AI should first reform a city through which it looses most gold.

LynxAbraxas:
Tested this for quite a while and observed no further unexpected changes, so I  think this can go in now.
@MartinGuehmann please feel free to open new PRs/issues for the ideas that are not covered then.

MartinGuehmann:
> I also noticed that the AI does not seem to repair its units, even if they are on the verge to die. Does the AI have a rule for that that just might need a higher priority?

Well, I guess the goal RETREAT was indented for that. However, now the game just uses it as a fall back goal if it has nothing to do for its units. In principle you can add a wounded army bonus to the goal so that it is more likely executed if the army contains a wounded unit.

MartinGuehmann:
> With #171 I noticed that the AI apparently does not priorise which of its converted cities should be reformed first, i.e. the AI should first reform a city through which it looses most gold.
> 
> E.g. here the AI reformed Cortona (12 conversion gold) while Clusium looses 156 gold through the conversion:
> 
> ss_2019-10-12_14:56:15

Well, as a human player I would just reform the city that is next to the reformator. Since this are ordinary attack units, I would reform each city in a turn. The AI would only reform half of its cities (if I interpret the values in strategies.txt correctly).

LynxAbraxas:
> The AI would only reform half of its cities (if I interpret the values in strategies.txt correctly).

Do you think that is intended or by mistake? Which of them do you mean?

MartinGuehmann:
> > The AI would only reform half of its cities (if I interpret the values in strategies.txt correctly).
> 
> Do you think that is intended or by mistake? Which of them do you mean?

`GoalElement { Goal GOAL_REFORM              Priority   450000  MaxEval   1  MaxExec 0.5 PerCity}`

You have MaxEval and MaxExecute. MaxEval are the number of goals the AI should consider and MaxExec is the number of goals to be executed. And that is done PerCity. So for each city the AI evaluates one goal and half of them are executed. However, I haven't looked at the code, so I don't know exactly if the PerCity does exactly this.

In the end whether intentional or by mistake does not matter. It only matters whether it is good behavior. And what is good behavior is up to us to decide.

Anyway, since I am right now quite busy, I won't add anything to this soon, so let's merge.

PR #187, opened 2019-08-28
Tuning the layout of the advance dependency graph

LynxAbraxas:
This PR is for tuning the layout of the advance dependency graph, such as all root nodes on the same rank, each age in a cluster-box and adding info on what each advance brings.


LynxAbraxas:
I won't find the time to realize this in the near future, so closed to serve as a reference. Please feel free to pick it up.

LynxAbraxas:
Ah well, was not able to give up on this one completely, but will not work on:
- each age in a subgraph-box 
- adding info on what each advance brings

LynxAbraxas:
Well, here it is with:
- each age in a cluster-box
- adding info on what each advance brings

LynxAbraxas:
This could be tuned further such that:
- the xlable info is included in the label but that would lead to differently sized nodes (which in turn could be set to a fixed height)
- unit becoming obsolete info
- tile improvements are included
but I will not work on that.

LynxAbraxas:
@MartinGuehmann ready for merging?

williamd67:
@LynxAbraxas: it looks really nice. Although there are still a lot of arrows. I also some arrows going backwards in the ages, probably correct according the data, still a little bit strange.

LynxAbraxas:
Thanks @williamd67.

> Although there are still a lot of arrows.

Well, that is due to the deps. defined in `Advance.txt`. Actually, they are already reduced a lot from transitive reduction.

> I also notice some arrows going backwards in the ages, probably correct according the data, still a little bit strange.

Yeah, the age-assignment is not fully consistent. As far as I know the age is only used for the clustering in the graph but not by the game logic.

MartinGuehmann:
> Yeah, the age-assignment is not fully consistent. As far as I know the age is only used for the clustering in the graph but not by the game logic.

Well, it is also used for the city styles and for a message telling you that you have entered a new age.

You reassigned some advances, but there are still enough advances for each age so that you could enter them. That's good.

Could you change it so that it uses the strings from gl.txt so that "Fluid Breathing" does not become "LIQUID BREATHING APPARATUS". Of course, the same applies for the age, building, and unit names.

You also changed the prerequisite of one advance. That means the number of advances leading to some of the governments could be changed dramatically. I have to look at that.

LynxAbraxas:
> Could you change it so that it uses the strings from gl.txt so that "Fluid Breathing" does not become "LIQUID BREATHING APPARATUS". Of course, the same applies for the age, building, and unit names.

Done.

> You also changed the prerequisite of one advance. That means the number of advances leading to some of the governments could be changed dramatically. I have to look at that.

If this is problematic, I can revert it.

MartinGuehmann:
Looks much better, yet. However, AI Surveillance gives Spy_PLANE, this doesn't look right.

The things that an advance gives are not put nicely in a box. However, if you want to stop working on that now, fine. Then put it into another pull request.

For the advance, I have to check the number of advances you need to reach each government. If that isn't screwed up to much than it is fine.

Was it intentional not to include the Chinese version of the poster?

LynxAbraxas:
> Looks much better, yet. However, AI Surveillance gives Spy_PLANE, this doesn't look right.

Done.

> The things that an advance gives are not put nicely in a box. However, if you want to stop working on that now, fine. Then put it into another pull request.

If put in a box the box will have varying size and would have to be included in the node to be fixed to it. With xlabels dot makes sure that node and extra-label stick to gether nicely.

> For the advance, I have to check the number of advances you need to reach each government. If that isn't screwed up to much than it is fine.

The only changed advance prerequisite is in 1495a15c84b4973f515159a483f9034177da22f6, where only Jurisprudence is skipped.

> Was it intentional not to include the Chinese version of the poster?

Yes, I cannot say if things are right, nor have I any experience in type-setting Chinese.

MartinGuehmann:
> > The things that an advance gives are not put nicely in a box. However, if you want to stop working on that now, fine. Then put it into another pull request.
> 
> If put in a box the box will have varying size and would have to be included in the node to be fixed to it. With xlabels dot makes sure that node and extra-label stick to gether nicely.
> 

Well, it could be in the same box as the advance itself and then have a standard size for all the boxes. In fact the original poster had the stuff an advance gives also under the box. However it was under the box, while for instance we have it left of Neural Programming, below Life Extension, even aligned right, and to the right of Gaia Controller.

In fact, I would prefer one item per line, and then have more space between the advances. 

> > Was it intentional not to include the Chinese version of the poster?
> 
> Yes, I cannot say if things are right, nor have I any experience in type-setting Chinese.

Couldn't you just try and see what happens. On the other hand we won't get anybody to translate it.

LynxAbraxas:
All info in each box now. I won't have the time to work on this any further.

LynxAbraxas:
@MartinGuehmann what do you think? Ready for merging now?

MartinGuehmann:
Well, I preferred having the stuff you get centered, too. And actually, now it would make sense to have the advance title on above the image then below.

But I would consider that as minor, so if you don't want to work on it anymore then we can merge.

LynxAbraxas:
> Well, I preferred having the stuff you get centered, too. 

That way the "grouping" of the types, i.e. the extra space between e.g. units and improvements was hardly visible and I was not able to find a way to increase that space without effecting other spacing.

> And actually, now it would make sense to have the advance title on above the image then below.

That looked a bit odd to me too with the translations now because the spacing of the text towards the images was different if e.g. a g or y was in the advance name or not.

> But I would consider that as minor, so if you don't want to work on it anymore then we can merge.

I will not find the time to work on this any further in the near future,  so if it is not essential let's go with it as is.

MartinGuehmann:
Good, then let's merge.

LynxAbraxas:
This will have to be my last contribution to this great project for the time being. I hope to be back some time and am very much looking forward to all the new contributions then.
Many thanks @MartinGuehmann and @williamd67 for your contributions and all your feedback and help on encountered issues.

williamd67:
@LynxAbraxas thanks for all your contributions and especially for your very positive feedback and support. It is really great to work with you. I will miss the quick tests and their results, you always seems to have ready after my commits.

I hope you will come back when you have time again.

PR #188, opened 2019-08-29
Add extra food, prod. and gold for goods received from trade

LynxAbraxas:
PR trying to implement the simple idea of https://github.com/civctp2/civctp2/issues/148#issuecomment-502468894.


LynxAbraxas:
The built error seems to be a travis problem, compiles fine on GL and seems to have some effect:

Ecbatana receives Mushrooms (which give +15 food and +10 gold):
ss_2019-08-29_21:52:44

Food and gold without this PR:
ss_2019-08-29_21:52:06

and with this PR:
ss_2019-08-29_21:48:01

Need to check if the changes in food and gold are correct.

MartinGuehmann:
> The built error seems to be a travis problem, compiles fine on GL and seems to have some effect:

Yeah, I also checked Travis, and it failed to install the needed packages. So it failed before it tried to compile any of your code.

In fact, I could restart the build, I assume you can do it, too. Well, if you think it is needed.

LynxAbraxas:
> In fact, I could restart the build, I assume you can do it, too. 

When I looked I could not find a way to restart it. It's not essential as I might push another commit that implements some way to see the additional values coming from bought goods. Currently, I'm not sure where to put that info though. Ideally behind each icon in the city tab (e.g. bread XXX (YYY)) but there is hardly any room left. There would be a line of space left by squeezing Citisen, Workers, Slaves together yielding some room before the efficiency bar. 
Hm, and/or exchanging the efficiency bar and the resource icons.

MartinGuehmann:
> When I looked I could not find a way to restart it.

Well, it is on this page: https://travis-ci.com/civctp2/civctp2/builds/125136310. On the right side a button says "Restart Build". You can also restart each build job, individually. I just don't know whether you need privileges for that.

Anyway, you are right. It is restarted with the next commit anyway.

LynxAbraxas:
I was able to make up some space to squeeze in another row for the additional food, prod. and gold from received goods:

ss_2019-08-31_15:38:57

and extra food and extra gold (actually commerce) make a change in the total values reported, compared to the values without this PR:

ss_2019-08-31_15:29:36

However, this is not the case for production:

ss_2019-08-31_15:39:49

and without the PR

ss_2019-08-31_15:30:14

The extra production does show up in the empire manager:

ss_2019-08-31_16:03:54

in regard to the old state

ss_2019-08-31_16:07:00

but also there does have an increase effect in PW.
@martinguehmann Could there be something wrong with the calculations of the production?

LynxAbraxas:
Figured, this PR should also include a commit such that the bonus is only given if the trade route of the good was not pirated.

MartinGuehmann:
> @MartinGuehmann Could there be something wrong with the calculations of the production?

You have two cities, one gets extra food and gold. And the other gets extra production and gold. So that seems to work. Your code looks also alright. And actually, not all goods give extra production or extra food or extra gold. So the behavior looks alright.

One thing, I miss is that if I look at these two rows then I don't know that the second row is about stuff that comes in via trade. Unfortunately, we don't really have space for putting trade somewhere onto the interface, but at least you could add tooltips for the extra icons.

LynxAbraxas:
> And the other gets extra production and gold.

Yes, but the production is not added, it is 569 with or without the PR, nor does it increase the PW, so it seems the extra production neither is used for building things nor for PW nor for unit upkeep, i.e. apparently without any effect.

> Unfortunately, we don't really have space for putting trade somewhere onto the interface, but at least you could add tooltips for the extra icons.

I was wondering if I just should give up the alignment and put goods in front of the second row, or afterwards since goods give commerce there is the space left for the science icon.
I was also wondering if there is a different icon for gold and commerce but could not find any. A gold coin seems to be used for both of them.

MartinGuehmann:
> > And the other gets extra production and gold.
> 
> Yes, but the production is not added, it is 569 with or without the PR, nor does it increase the PW, so it seems the extra production neither is used for building things nor for PW nor for unit upkeep, i.e. apparently without any effect.

I think, I didn't understood your original description. It's probably something with gross and net production. Just move this line below your insertion in CityData.cpp:

`m_collected_production_this_turn = m_gross_production;`

> > Unfortunately, we don't really have space for putting trade somewhere onto the interface, but at least you could add tooltips for the extra icons.
> 
> I was wondering if I just should give up the alignment and put goods in front of the second row, or afterwards since goods give commerce there is the space left for the science icon.

You could put science first, so that you have space for the label.

> I was also wondering if there is a different icon for gold and commerce but could not find any. A gold coin seems to be used for both of them.

As far as I can see Activision uses gold and commerce as synonym.

LynxAbraxas:
> I think, I didn't understood your original description. It's probably something with gross and net production. Just move this line below your insertion in CityData.cpp:

Many thanks @MartinGuehmann, that did the trick, now works as expected:

Before:

ss_2019-09-28_22:29:31
ss_2019-09-28_22:29:43

now:

ss_2019-09-28_22:26:23
ss_2019-09-28_22:26:44

LynxAbraxas:
> Figured, this PR should also include a commit such that the bonus is only given if the trade route of the good was not pirated.

This seems to be quite a challenge to implement because `m_buyingResources` only contains the goods not the corresponding routes. `m_buyingResources` is populated when the trade route is created, so removing the good from the list in case the route is pirated is probably not the best way to go. @MartinGuehmann  any idea on how this could be done nicely? Otherwise it could be left as is and pirating only means to grab the money from the trade routes but not goods. However, then one could not weaken the receiver of goods by pirating.

MartinGuehmann:
> > Figured, this PR should also include a commit such that the bonus is only given if the trade route of the good was not pirated.
> 
> This seems to be quite a challenge to implement because `m_buyingResources` only contains the goods not the corresponding routes. `m_buyingResources` is populated when the trade route is created, so removing the good from the list in case the route is pirated is probably not the best way to go. @MartinGuehmann any idea on how this could be done nicely? Otherwise it could be left as is and pirating only means to grab the money from the trade routes but not goods. However, then one could not weaken the receiver of goods by pirating.

Well, you would have to create a list of non pirates goods in each city on begin turn. Then you could calculate it quickly, which is needed here. Unfortunately, it takes some memory, but it shouldn't be too much.

Anyway, it seems to be a bit more complicated, so that it could go into its own pull request anyway. But that's up to you, if you decide not to implement it now then open an issue for it.

LynxAbraxas:
Label for the trade goods row bonuses:
ss_2019-10-08_23:38:30

LynxAbraxas:
It now makes a difference if a trade route is cancelled:

ss_2019-10-04_23:34:48

ss_2019-10-04_23:29:34

LynxAbraxas:
And it's now possible to boost own cities with inner trade:

ss_2019-10-11_18:50:16

LynxAbraxas:
And it's now possible to boost own cities with inner trade:

ss_2019-10-11_18:50:16

LynxAbraxas:
And it's now possible to boost own cities with inner trade:

ss_2019-10-11_18:50:16

LynxAbraxas:
> Anyway, it seems to be a bit more complicated, so that it could go into its own pull request anyway. But that's up to you, if you decide not to implement it now then open an issue for it.

I'd leave the PR here as is and open an issue for discussing if and how pirating could be taken into account.

MartinGuehmann:
> > Anyway, it seems to be a bit more complicated, so that it could go into its own pull request anyway. But that's up to you, if you decide not to implement it now then open an issue for it.
> 
> I'd leave the PR here as is and open an issue for discussing if and how pirating could be taken into account.

Well then, we are done here and this can go in.

Issue #130, opened 2019-05-03
Tech tree and AI research choice


MartinGuehmann:
Regarding https://github.com/civctp2/civctp2/pull/119#issuecomment-488751731 and  https://github.com/civctp2/civctp2/pull/119#issuecomment-488822643, let's put those into an own issue. Again the two comments:

> > In regard to #129, how about making some AI civs prefer some govs over others making them differ to other AI civs?
> 
> They actually do, however they research all the stuff in the same order, except for a minor variation. Even so all the different personalities have their own advance list, which are just copies of each other and somebody forgot to modify.
> 
> Actually, all the different governments have ranks, 1, 2, 3, and 4 matching the city limits. So in principle, it should be possible that you can research all the governments of the same rank without having any of them. I don't know whether the tech tree is designed like this, but in fact it should be designed like this, so that you really have a choice.
> 
> Right now the AI will research the governments and will use the one of the highest rank available, even so it has not researched its preferred government. And once it got its preferred government it will switch, even so it has been for quite long in a non-preferred government.

> > Even so all the different personalities have their own advance list, which are just copies of each other and somebody forgot to modify.
> 
> Would some random shuffeling suffice or do we need to take advance tree connections into account? I.e. only shuffle them if they are accessable at that stage.

Not really, you can put an advance at the top of list and they AI will research it if you have the prerequisites, but if you don't care to shuffle the prerequisites then the AI may research more than it needs to get to Theocracy or Monarchy.

The other idea is that for instance Theocracy, Monarchy or Republic are supposed to be on the same level, so you should need to research the same number of advances to get to them so that you really have a choice between them.

So I not only propose to make good advance lists for the AI but also to modify the tech tree so that the Governments are indeed on the same level in the tech tree, so that you have to decide what you go for and indeed that you have a choice.

Of course modifying the tech tree also asked the question whether we should do the different governments really belong here or there.

For instance, the Roman Republic ended let's say with Caesar or Augustus, then Rome had an emperor and before the Republic Rome had Kings. The difference between a Roman king and a Roman emperor was that the office of emperor wasn't inherited while the office of king was or maybe I have to look it up.

Anyway, even if the office of Roman king and emperor were similar. You couldn't call emperors kings, because the Romans viewed kings as evil tyrants.

So we could say we went from tyranny to republic to monarchy, but to some extend this are just names.

LynxAbraxas:
[...]

PR #190, opened 2019-09-08
Clerics and Televangelists can now be build under any government

MartinGuehmann:
For now, this is something for testing. This is a way to solve the issue #129 that the AI can only build Clerics and Televangelists for a very short time, since it is only for roughly a few turns in Theocracy the time needed to research Fascism.

There are still alternatives to slove this problem for instance to introduce Fascim later in the tech tree, as suggested in #130.

..\ctp2_data\default\gamedata\Units.txt

### DIFF
--- a/ctp2_data/default/gamedata/Units.txt
+++ b/ctp2_data/default/gamedata/Units.txt
@@ -879,7 +879,6 @@ UNIT_CLERIC {
    EnableAdvance ADVANCE_THEOLOGY
    ObsoleteAdvance ADVANCE_MASS_MEDIA
    UpgradeTo UNIT_TELEVANGELIST
-   GovernmentType GOVERNMENT_THEOCRACY
    ActiveDefenseRange 0
    LossMoveToDmgNone
    MaxFuel 0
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Tested this for some truns now. The AI seems to built and use televangelist appropriately, doing conversions and happyness attacks. However, in the current state of this PR the goverment type Theocracy does not offer any benefits over other gov. types, so I would suggest to some refinement. While introducing Fascim later in the tech tree would help to have the AI in this gov. it would still change to another one later, so it would not be able to come up with clerics/televangelists later in the game. Therefore I would find it interesting if either some civs would stick to Theocracy even if newer govs are available to it. Or making succes of clerics/televangelists and/or amount of conversion gold higher for civs in Theocracy.

LynxAbraxas:
From https://github.com/civctp2/civctp2/pull/225#issuecomment-596218300

> Tested also with #190 before Televangelists can be built. Clerics are built by the AI and used all right, but apparently not if the AI is in Theocracy. So far I could not find any setting that would prevent the building of Clerics for AIs in Theocracy, but in any case that would not be part of this PR but perhaps rather of #190.

MartinGuehmann:
> Could be closed in favor of #410. @MartinGuehmann what do you think?

Indeed. So closed.

Issue #51, opened 2018-12-13
Wrong city name for messages like "slaves put at work at ..."


LynxAbraxas:
Somtimes the names of cities in messages e.g.  "slaves put at work at ..." are wrong (but not always).

MartinGuehmann:
I guess, you added some strings and loaded an old savegame?

LynxAbraxas:
Thank @MartinGuehmann for taking a look at this issue.

> I guess, you added some strings and loaded an old savegame?

No, this even happens in a "vanilla" game, and only sometimes the city names are wrong, not always. 
If I add some strings concerning the game messages or the goods, I get completely different message strings like "citizen starving" instead of "building X was finished".

MartinGuehmann:
I guess then we need it a little bit more specific and also a savegame right before that happens.

Probably, for the slaves you could just put slaves onto the map and then see were the slaves go.

LynxAbraxas:
Hm, how do you place slaves onto the map? Is ther some kind of slave unit?

MartinGuehmann:
> Hm, how do you place slaves onto the map? Is ther some kind of slave unit?

That's supposed to be slavers. One "r" missing.

LynxAbraxas:
Had a look for possible sources of this issue but could not find any but so far. The point where the city goes into the message is here:
https://github.com/civctp2/civctp2/blob/9a1ab16704ccb10c23d14a679adff21501fcd910/ctp2_code/gs/gameobj/ArmyData.cpp#L3057
Possibly the lookup of the city name for the message string is buggy.

LynxAbraxas:
Observing something similar in the case of franchise removal with #191, e.g. I'm told franchise in Sligo closed but the city icon indicates the franchise was removed in Cork:

![ss_2019-10-03_22:01:05](https://user-images.githubusercontent.com/21012234/66162166-1b27fb80-e62e-11e9-8f72-92851d3153c3.png)

If I go  there with a unit, it actually was Cork and not Sligo:

![ss_2019-10-03_22:02:22](https://user-images.githubusercontent.com/21012234/66162304-63471e00-e62e-11e9-9ec2-22f87aace33f.png)

Similar here with Rivas and Makale:

![ss_2019-10-03_22:16:50](https://user-images.githubusercontent.com/21012234/66162337-73f79400-e62e-11e9-9b84-4a74def78857.png)

Both this message and that of slave raids ends up here:
[...]

PR #191, opened 2019-09-28
Message for the victim of a successful reformation

LynxAbraxas:
Tell the victim of a reformation (conversion owner) about a successful reformation, similar to the owner of a franchise about its close up.

### DIFF
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -2370,27 +2370,29 @@ ORDER_RESULT ArmyData::SueFranchise(const MapPoint &point)
 		return ORDER_RESULT_ILLEGAL;
 
 	Unit &	u		= m_array[uindex];
-	Cell *	cell	= g_theWorld->GetCell(point);
+	Unit c = GetAdjacentCity(point);
 
-	if(!cell || cell->GetCity().m_id == 0)
+
+
+	if(c.m_id == 0)
 		return ORDER_RESULT_ILLEGAL;
 
-	if(cell->GetCity().GetOwner() != m_owner) {
-
-		return ORDER_RESULT_ILLEGAL;
-	}
-
-	if(cell->GetCity().GetFranchiseOwner() < 0) {
+	if(c.GetOwner() != m_owner) {
 
 		return ORDER_RESULT_ILLEGAL;
 	}
 
-	if(cell->GetCity().GetFranchiseTurnsRemaining() == 0) {
+	if(c.GetFranchiseOwner() < 0) {
 
 		return ORDER_RESULT_ILLEGAL;
 	}
 
-	if(cell->GetCity().GetFranchiseTurnsRemaining() > 0 && cell->GetCity().GetFranchiseTurnsRemaining() <= g_theConstDB->Get(0)->GetTurnsToSueFranchise()) {
+	if(c.GetFranchiseTurnsRemaining() == 0) {
+
+		return ORDER_RESULT_ILLEGAL;
+	}
+
+	if(c.GetFranchiseTurnsRemaining() > 0 && c.GetFranchiseTurnsRemaining() <= g_theConstDB->Get(0)->GetTurnsToSueFranchise()) {
 
 		return ORDER_RESULT_ILLEGAL;
 	}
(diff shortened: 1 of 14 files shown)


LynxAbraxas:
Using `so->AddGold(static_cast<City>(cell->GetCity()).GetProductionLostToFranchise());` instead of the changes to `Unit.*` and `UnitData.*` yields:
```
ArmyData.cpp:2409:26: error: 'City' does not name a type
  so->AddGold(static_cast<City>(cell->GetCity()).GetProductionLostToFranchise());
```
@MartinGuehmann would it be possible to achieve this with a cast?

MartinGuehmann:
> Using `so->AddGold(static_cast<City>(cell->GetCity()).GetProductionLostToFranchise());` instead of the changes to `Unit.*` and `UnitData.*` yields:
> 
> ```
> ArmyData.cpp:2409:26: error: 'City' does not name a type
>   so->AddGold(static_cast<City>(cell->GetCity()).GetProductionLostToFranchise());
> ```
> 
> @MartinGuehmann would it be possible to achieve this with a cast?

No, you can't do this. There is is no class "City", cities are basically units with additional CityData.

LynxAbraxas:
Basically works:

ss_2019-09-30_23:12:20

ss_2019-09-28_22:09:32

but issue #51 seems to apply to the case of suing a franchise.

LynxAbraxas:
https://github.com/civctp2/civctp2/pull/191/commits/4ec3485db04b6f0f6372ac0d837787a225e30fcf does not seem to solve #51.

Comparing similar functions like `ReformCity`, I wonder if this line
https://github.com/civctp2/civctp2/blob/4ec3485db04b6f0f6372ac0d837787a225e30fcf/ctp2_code/gs/gameobj/ArmyData.cpp#L2405
should be changed to use `c.m_id` like
https://github.com/civctp2/civctp2/blob/4ec3485db04b6f0f6372ac0d837787a225e30fcf/ctp2_code/gs/gameobj/ArmyData.cpp#L4307

MartinGuehmann:
> [4ec3485](https://github.com/civctp2/civctp2/commit/4ec3485db04b6f0f6372ac0d837787a225e30fcf) does not seem to solve #51.
> 
> Comparing similar functions like `ReformCity`, I wonder if this line
> https://github.com/civctp2/civctp2/blob/4ec3485db04b6f0f6372ac0d837787a225e30fcf/ctp2_code/gs/gameobj/ArmyData.cpp#L2405
> 
> 
> should be changed to use `c.m_id` like
> https://github.com/civctp2/civctp2/blob/4ec3485db04b6f0f6372ac0d837787a225e30fcf/ctp2_code/gs/gameobj/ArmyData.cpp#L4307

Well, the class ID, from which Unit is derived, defines the needed cast operators:

```
	operator int() const  { return m_id; };
	operator unsigned int() const  { return m_id; };
	operator unsigned long() const  { return m_id; };
```

Actually, I would prefer them as defined as sint32 etc. operators. Actually, it shouldn't make a difference, whether you use  `c.m_id`. However, considering the definitions of the cast operators, this would be cleaner.

LynxAbraxas:
I see, thanks @MartinGuehmann for that explanation. 
It seems the issure reported above was an unfortunate coincidence. I was testing `TURNS_TO_SUE_FRANCHISE 1`. Due to a `reloadslic` the former messages were gone and just in that turn the franchises in the neighboring cities were removed (with a delay as well). So everything works as expected. I found that this nonobvious behavior deserves some extra info in the messages, so I added the info on the number of turns it takes to remove the franchise.

MartinGuehmann:
And the message should really say: "to close up shop in 1 turn**s**"?

LynxAbraxas:
> And the message should really say: "to close up shop in 1 turns"?

Since it can be more than 1 turn or 0, I decided to go for the plural in any case because it would be quite cumbersome to differentiate that with an extra message. Or is there a simple way to add to the `SlicObject` a string that can be used in the message text?

LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/191/commits/4aa49b95d450b2cbd8526d1ff99353213626f746 the sue messages of units are now correct:

ss_2019-10-05_16:24:34

LynxAbraxas:
This PR is read to go in from my side unless there are any concerns.

MartinGuehmann:
> > And the message should really say: "to close up shop in 1 turns"?
> 
> Since it can be more than 1 turn or 0, I decided to go for the plural in any case because it would be quite cumbersome to differentiate that with an extra message. Or is there a simple way to add to the `SlicObject` a string that can be used in the message text?

Maybe there is, but if you just write "turn(s)" / "Runde(n)" you should be fine.

For the other languages I would just leave the original text and add the English piece. Maybe with a note: Please translate. So that this won't get lost.

LynxAbraxas:
OK, done.

MartinGuehmann:
Well, if you could still add a German translation for:

`REFORM_CITY_VICTIM			"Fremde Inquisitoren haben die Stadt {city[0].name} reformiert und ihre Bürger stehen nicht mehr unter unserem Einfluss. Sie zahlten jede Runde {gold[0].value} Gold."`

LynxAbraxas:
OK, done too.

MartinGuehmann:
Well, then let's merge.

LynxAbraxas:
Message to the victim of a successful franchise also works:
ss_2019-10-25_00:18:32
ss_2019-10-25_23:51:55

PR #195, opened 2019-10-12
Fix cancelling of domestic trade routes

LynxAbraxas:
This PR is meant to fix that under some circumstances domestic trade routes (sender civ == receiver civ) are cancelled by the game logic.
With fe3b57cd409a00acef78855208219a5e7362cb86 this message can appear:

ss_2019-10-12_16:02:53

Note, that also the trade manager then does not list the good (Jellyfish) of the lost trade route (from Esfahan) any more, but it reappears in the list some turns later.

With #188, this can make a serious difference to the growth of a city.
So either this should lead to a message or should not happen for domestic trade routes.


LynxAbraxas:
With 81ae68784a9fdf88d103e8e78d851bd1278cca97 I get in the case of the above message:
```
CityData.cpp L4348: deadRoutes, resource: 68, m_collectingResources: 0, m_sellingResources: 0, i: 1!
CityData.cpp L4422: KillTradeRoute!
TradeRoute.cpp L70: Kill trade route cause: 1, owner: 1, resource: 68!
```

So my current guess is that there is something wrong with populating `m_collectingResources` which causes this condition to be true
https://github.com/civctp2/civctp2/blob/81ae68784a9fdf88d103e8e78d851bd1278cca97/ctp2_code/gs/gameobj/CityData.cpp#L4343
even though the good/resource (jellyfish) is still on the map and within the city radius (of Esfahan).

LynxAbraxas:
> So my current guess is that there is something wrong with populating m_collectingResources

@MartinGuehmann could this happen from introducing a new good during a game ? I.e. for a game that was created when that good did not exist and where the good was placed later on in cheat mode?

Oddly trade routes with that good from other cities seem to work all fine, i.e. are not cancelled by some game logic.

MartinGuehmann:
> > So my current guess is that there is something wrong with populating m_collectingResources
> 
> @MartinGuehmann could this happen from introducing a new good during a game ? I.e. for a game that was created when that good did not exist and where the good was placed later on in cheat mode?

In principle, yes. Especially, if mapping the goods fails. However, that should only happen once. And thus cannot explain what you observe. The AIs cancel the trade routes during the game, without reloading. They may also do after reloading, but not because you added goods.

LynxAbraxas:
Hm, I guess I should postpone this until I start with a fresh game and wait until I observe the same problem there.

LynxAbraxas:
With #205, similar to https://github.com/civctp2/civctp2/pull/205#issuecomment-552100163:
A giant squid trade route exists in the export tab (and on the map) from Cunaxa to Shiraz:

ss_2019-11-07_22:02:19

but does not exist in the import tab nor is its (food) bonus taken into account:

ss_2019-11-07_22:01:38

So I would conclude that somehow the route is in the `TradeSourceList` but not in the `TradeDestinationList` of the corresponding cities.
It is possible to break this stale route and one gets the caravans - 1. Recreating this route then does not help:

ss_2019-11-07_22:04:28

After creation of a new route (5 caravans removed) with giant squid (20 food bonus, showing up in the import tab now) the city food bonus still does not increase (nor in the next turn):

ss_2019-11-07_22:05:32

LynxAbraxas:
> After creation of a new route (5 caravans removed) with giant squid (20 food bonus, showing up in the import tab now but is gone again in the next turn) the city food bonus still does not increase:

Only  what  is in the `tradeDestinationList`
https://github.com/civctp2/civctp2/blob/408bbbaae58a7617d546ca6b2fd27b2d1c73eb64/ctp2_code/gs/gameobj/CityData.cpp#L4413-L4456
gets taken into account for the city bonuses:
https://github.com/civctp2/civctp2/blob/408bbbaae58a7617d546ca6b2fd27b2d1c73eb64/ctp2_code/gs/gameobj/CityData.cpp#L4454

LynxAbraxas:
https://github.com/civctp2/civctp2/pull/195#issuecomment-552104918 seems to be resolved with https://github.com/civctp2/civctp2/pull/195/commits/3278f46c4bfd243149cb00779d355b7795f84b0e but the initally describe problem (with jellyfish) is not solved.

> Note, that also the trade manager then does not list the good (Jellyfish) of the lost trade route (from Esfahan) any more, but it reappears in the list some turns later.

With https://github.com/civctp2/civctp2/pull/195/commits/81ae68784a9fdf88d103e8e78d851bd1278cca97 I get from https://github.com/civctp2/civctp2/pull/195/commits/81ae68784a9fdf88d103e8e78d851bd1278cca97#diff-647de8d8715871ce5f3623bf0a4f0158R4347
```
CityData.cpp L4347: deadRoutes, resource: 68, m_collectingResources: 0, m_sellingResources: 0, i: 1!
```
so it seems that in that case for jellyfish (`routeResource == 68`) `m_collectingResources[routeResource] == 0`, which means that with the new turn jellyfish is not "collected" for some reason. @MartinGuehmann any idea what could cause that considering that it was collected before an the city radius of the collecting city is unchanged?

MartinGuehmann:
> [#195 (comment)](https://github.com/civctp2/civctp2/pull/195#issuecomment-552104918) seems to be resolved with [3278f46](https://github.com/civctp2/civctp2/commit/3278f46c4bfd243149cb00779d355b7795f84b0e) but the initally describe problem (with jellyfish) is not solved.
> 
> > Note, that also the trade manager then does not list the good (Jellyfish) of the lost trade route (from Esfahan) any more, but it reappears in the list some turns later.
> 
> With [81ae687](https://github.com/civctp2/civctp2/commit/81ae68784a9fdf88d103e8e78d851bd1278cca97) I get from [81ae687#diff-647de8d8715871ce5f3623bf0a4f0158R4347](https://github.com/civctp2/civctp2/commit/81ae68784a9fdf88d103e8e78d851bd1278cca97#diff-647de8d8715871ce5f3623bf0a4f0158R4347)
> 
> ```
> CityData.cpp L4347: deadRoutes, resource: 68, m_collectingResources: 0, m_sellingResources: 0, i: 1!
> ```
> 
> so it seems that in that case for jellyfish (`routeResource == 68`) `m_collectingResources[routeResource] == 0`, which means that with the new turn jellyfish is not "collected" for some reason. @MartinGuehmann any idea what could cause that considering that it was collected before an the city radius of the collecting city is unchanged?

m_collectingResources is reset every turn, also when the city grows, and when the city is initialized. As far as I could see, we have here copied code, which may not do the same thing as it should be. Actually, m_collectingResources should only be changed when the resources are collected.

LynxAbraxas:
The loss of the trade route seems to be sovled with edf8aafc653798fcbaf4796f863640b6e0b5d2a0. This probably also solves the oberseved issue that sometimes goods within the city influence do not appear as tradeable good in the trade manager.
It seems that originally
https://github.com/civctp2/civctp2/blob/9cbaf854611358b55e8e3c6bbe11b546e00ef1ca/ctp2_code/gs/gameobj/CityData.cpp#L1983-L1985
no depencende on partSquaredRadius existed.
@MartinGuehmann  the corresponding changes seem to come from you. Can you take a look if the dependence was unintended?

MartinGuehmann:
> The loss of the trade route seems to be sovled with [edf8aaf](https://github.com/civctp2/civctp2/commit/edf8aafc653798fcbaf4796f863640b6e0b5d2a0). This probably also solves the oberseved issue that sometimes goods within the city influence do not appear as tradeable good in the trade manager.
> It seems that originally
> https://github.com/civctp2/civctp2/blob/9cbaf854611358b55e8e3c6bbe11b546e00ef1ca/ctp2_code/gs/gameobj/CityData.cpp#L1983-L1985
> 
> 
> no depencende on partSquaredRadius existed.
> @MartinGuehmann the corresponding changes seem to come from you. Can you take a look if the dependence was unintended?

I think that the original idea was to only collect the good if it is in a ring that is being worked on. So to have it like in Civ2 and CTP1, where you have to put a worker onto that tile. Actually today, I would like to have it so that no matter how many rings are worked on, the good is collected. This actually simplifies a few things, and you don't have to recompute less stuff. The whole idea of the new resource collection stuff was, to just compute all the stuff once and then check how much you actually need by simple formula.

The current way is very time consuming, since you have to test all the slide settings one by one, assign all the specialists. And see whether everything fits. This is just brute force and takes quite some time if you have 200 or something cities. However, in the end it got too complicated and I stopped. The problem with the one formula approach you also have dissolve some dependencies. Happiness depends on production, so you add entertainers, but with more entertainers you produce less, and so you pollute less, which means you need less entertainers to cure happiness.

And of course, once you are finished with the new resource process, the city should produce more or less the same as with the old way of collecting and processing resources. That's quite challenging.

LynxAbraxas:
> I think that the original idea was to only collect the good if it is in a ring that is being worked on.

Hm, not sure about that since in `AdjustSizeIndices` it is still
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_code/gs/gameobj/CityData.cpp#L8936
While I would find it intersting to have "collect the good if it is in a ring that is being worked on", I would also want some kind of indication of this ring being worked on, otherwise it is very hard to understand why a good is collected once but not always. Perhaps some kind of border line in addition to the city influence border.

In any case I do wonder why this ring being worked on changes/decreases in size eventhough the city grows. Could that be the actual bug of this issue?

LynxAbraxas:
> So to have it like in Civ2 and CTP1

How was it in CTP1? Can't find anything relating. The workers are just listed under labor:
ss_2020-01-03_21:10:38

LynxAbraxas:
Any way, since https://github.com/civctp2/civctp2/commit/edf8aafc653798fcbaf4796f863640b6e0b5d2a0 solves the issue and equalizes the code in relation to `AdjustSizeIndices`, should we make the correction for the "ring being worked on" its own issue and use this PR to solve the initial issue?

LynxAbraxas:
>>    In any case I do wonder why this ring being worked on changes/decreases in size eventhough the city grows. Could that be the actual bug of this issue?
> That's because you turn workers into specialists and thus you remove them from the field. 

That I would understand, but I did not do that, so some automatism must have lead to such a change. Any idea which one that could be?

> in CTP1, I think you can right click onto the city and than a window pops up with the workers, where you can move them around.

Many thanks for pointing that out. I was not aware of that. Created #247 for that. However, that will not be so simple to realize, so I would suggest to merge this PR (which also equalizes this with the code of `AdjustSizeIndices`) and possibly switch back the the former behavior if #247 or something similar is realized. Although still then it would be bad that a trade route would get lost due to worker reordering or creating specialist. This should at least lead to a confirmation message.

LynxAbraxas:
Since #247 can be regarded as a separate issue, I also created #250. Meanwhile, I'd suggest to merge this PR, because #250 does not appear to be realizable easily.

LynxAbraxas:
> Since #247 can be regarded as a separate issue, I also created #250. Meanwhile, I'd suggest to merge this PR, because #250 does not appear to be realizable easily.

MartinGuehmann:
> > Since #247 can be regarded as a separate issue, I also created #250. Meanwhile, I'd suggest to merge this PR, because #250 does not appear to be realizable easily.
> 
> @MartinGuehmann What do you think? Ready for merge?

Sure we can merge, basically it just contains one change that matters, two changes that are more for the ease of the player and some comments added.

PR #199, opened 2019-10-12
Correct city name told in enslave settler message

LynxAbraxas:
This PR solves #51, which only concerned enslaving settlers. It seems the bug is caused by different results for the slave-city of `IsSlaveRaidPossible` and `GetSlaveCity`.

With this PR:
ss_2019-10-12_19:36:09
ss_2019-10-12_19:39:12

Without this PR:
ss_2019-10-12_19:33:41
ss_2019-10-12_19:29:33

This PR also adds an eye point to the message to ease jumping to the city in order to check its status with the new slave.


MartinGuehmann:
Then this is solved and can go in.

PR #200, opened 2019-10-19
Debug liberate city

LynxAbraxas:
Currently when conquering a city and choosing "Liberate" the conquering units are stuck in the then vandal/barbarian city. 

ss_2019-10-20_01:00:35

They cannot be moved out (due to ZOC?) and are killed over the next turn. This PR tries to solve this by either giving the units over to the barbarians or by sending them home (like when expelled).

### DIFF
--- a/ctp2_code/gs/gameobj/armyevent.cpp
+++ b/ctp2_code/gs/gameobj/armyevent.cpp
@@ -627,6 +627,7 @@ STDEHANDLER(ArmyGetExpelledOrderEvent)
 	SlicObject *so = new SlicObject("42UnitExpelled");
 	so->AddCivilisation(player);
 	so->AddRecipient(victim);
+	so->AddLocation(pos);
 	g_slicEngine->Execute(so);
 
 	return GEV_HD_Continue;
--- a/ctp2_code/gs/slic/slicfunc.cpp
+++ b/ctp2_code/gs/slic/slicfunc.cpp
@@ -7546,22 +7546,50 @@ SFN_ERROR Slic_Liberate::Call(SlicArgList *args)
 
 	//if(!args->GetInt(0, cause))
 	//	return GEV_HD_Continue;
-
+    
     city.ResetCityOwner(PLAYER_INDEX_VANDALS, FALSE, CAUSE_REMOVE_CITY_DIPLOMACY); // must be before sending armies home otherwise NearestFriendlyCityWithRoom returns the current city
 
-    //// send conquering units home
+    //// send conquering armies home, based on code from ArmyData::Expel
     Cell *cell = g_theWorld->GetCell(city.RetPos());
     sint32 i, n = cell->GetNumUnits();
 
+    bool 		foundCity	= false;
     MapPoint 		cpos;
-    UnitDynamicArray    revealed;
+    CellUnitList 	expelled;
     for(i = 0; i < n; i++) {
-	Unit u = cell->AccessUnit(i);
-	if(u.NearestFriendlyCityWithRoom(cpos, n, u.GetArmy())){
-	    u.SetPosition(cpos, revealed);
-	    }
-	else{
-	    u.Kill(CAUSE_REMOVE_ARMY_EXPELLED_NO_CITIES, PLAYER_INDEX_VANDALS);
+	    Unit u = cell->AccessUnit(i);	    
+	    foundCity = u.NearestFriendlyCityWithRoom(cpos, n, u.GetArmy());
+	    expelled.Insert(u);
+	}
+
+    n = expelled.Num();
+    if(n > 0) {
+	for(i = 0; i < n; i++) {
+	    if(foundCity) {
+		Army newArmy;
+		if(expelled[i].GetArmy().Num() > 1) {
+		    newArmy = g_player[expelled[i].GetOwner()]->GetNewArmy(CAUSE_NEW_ARMY_EXPELLED);
+		    } else {
+		    newArmy = expelled[i].GetArmy();
+		    }
+		
+		g_gevManager->AddEvent(GEV_INSERT_AfterCurrent, GEV_GetExpelledOrder,
+		    GEA_Army, newArmy,
+		    GEA_MapPoint, cpos,
+		    GEA_Player, PLAYER_INDEX_VANDALS,
+		    GEA_End);
+		
+		if(expelled[i].GetArmy().Num() > 1) {
+		    g_gevManager->AddEvent(GEV_INSERT_AfterCurrent, GEV_AddUnitToArmy,
+			GEA_Unit, expelled[i],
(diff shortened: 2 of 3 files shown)


LynxAbraxas:
@MartinGuehmann how to best sent home the conquering units? While  https://github.com/civctp2/civctp2/pull/200/commits/264c864341fedb3e581449a9fb262dfb6ab5dd5d works as expected, the game crashes over the next turn with 9159654bd7b3379a1a2b3c56bccd8930d nor are the units sent home.

MartinGuehmann:
As far as I can see you just set the positions of the units, but you don't set the army, actually you have to look at the move army code, there is also something called teleport.

LynxAbraxas:
Thanks @MartinGuehmann  for that hint. I used the code from expell order (https://github.com/civctp2/civctp2/pull/200/commits/e218814250dbd271b16f4529d2d6e93aafbf3939) and the game now does not crash any more but the units/armies still remain in the liberated city:

ss_2019-10-20_15:02:37

Any idea what could still be wrong?

MartinGuehmann:
> Any idea what could still be wrong?

The question is if the expel order is used in the game. For now I would think it is a left over from CTP1. I don't think that units are expelled in CTP2

LynxAbraxas:
Many thanks @MartinGuehmann for that hint, I did not think of that because expelling is possible and also used by the AI but possibly that uses different code. Do you know what code is used in CTP2 for that?

MartinGuehmann:
> Many thanks @MartinGuehmann for that hint, I did not think of that because expelling is possible and also used by the AI but possibly that uses different code. Do you know what code is used in CTP2 for that?

Well, actually I thought of the army retreat code, when you agreed to a leave our lands treaty. You are right the AI does some opportunity actions on special units. Check the AI code for these opportunity actions. There you should find the expel code. Probably, expelling is handled by some slic event.

LynxAbraxas:
> Well, actually I thought of the army retreat code, when you agreed to a leave our lands treaty.

Is that for CTP1 or CTP2? I think that would be really interesting to enable because currently the AI does not bother removing its armies when it has agreed to withdraw its troops, most often violating that agreement in ther very next round. So often it is only possible to remove an army/unit from e.g. an important railroad by declaring war.

MartinGuehmann:
> > Well, actually I thought of the army retreat code, when you agreed to a leave our lands treaty.
> 
> Is that for CTP1 or CTP2? I think that would be really interesting to enable because currently the AI does not bother removing its armies when it has agreed to withdraw its troops, most often violating that agreement in ther very next round. So often it is only possible to remove an army/unit from e.g. an important railroad by declaring war.

Well, that is CTP1. Basically, retreating is teleporting the armies to a friendly city.

Making the AI retreating units by pulling them back you would have to add a priority bonus to the retreat goal for armies in foreign territory with the retreat treaty. And of course they need some time for retreating.

LynxAbraxas:
>  but the units/armies still remain in the liberated city:

Figured that was due to the city not being given to the barbarians before sending home the units/armies, so the closest city for the expel target was the city itself.

LynxAbraxas:
> The question is if the expel order is used in the game. For now I would think it is a left over from CTP1. I don't think that units are expelled in CTP2. So here could still be something wrong.

They are, if you use the expell action (which normally is only possible for civilian units) which then uses the very same code.

LynxAbraxas:
Eventhough 099e00af7fada1c478f59af4d0849a9d75ca430d uses the exact same code as in `ArmyData::Expel`
https://github.com/civctp2/civctp2/blob/408bbbaae58a7617d546ca6b2fd27b2d1c73eb64/ctp2_code/gs/gameobj/ArmyData.cpp#L2455-L2499
only the last unit of the conquering army is actually expelled.
However, if the unit to be expelled is added to the newly created army (as in b18045dbacda3464368367bc73efbd3ddfc2b195) it works:

ss_2019-10-27_19:15:07

 It also works if the conquering army is used (a7ad82ba83b461dad72b19f6278e9e457bf674e7, not creating a new army for each unit) which leads the effect that the army as a whole is expelled:

ss_2019-10-27_19:36:43

This makes me wonder why the code in  `ArmyData::Expel` works as expected. @MartinGuehmann do you have an explanation for why it works there even though the units are not added to the newly created armies before expelling?

LynxAbraxas:
Optimization idea for expell code tracked with #207.

LynxAbraxas:
Basically good to go in. 

I wonder if the liberated city should be given some units as inital garrison. In the current state it is undefendet initially and the Barbarians can only built ancient units. @MartinGuehmann  what do you think? 
Actually I would find it most interesting to not give the city to the Barbarians but a new civ. That way it could inherit the state (advances, age, etc) from its former civ (before the conquering) and could built units comparable to those before the conquering.
Would those be splitter civs like mentioned in fae7f93c733d822abe848687b6a4bf0cfbf69a22? However, I've never observed that such a new splitter civ turned up in a game.

LynxAbraxas:
@MartinGuehmann any ideas/comments or should it go in?

MartinGuehmann:
> Basically good to go in.
> 
> I wonder if the liberated city should be given some units as inital garrison. In the current state it is undefendet initially and the Barbarians can only built ancient units. @MartinGuehmann what do you think?

Maybe, the Barbarians should be able to build current units as the civs around them. If you want to give them a unit then it should be one unit. The problem with those splitter civs is that they are conquered easily by their neighbors, whether they have some units or not doesn't matter. Actually, if you give those cities a garrison than you also have to units to splitter civs. However, here it would be easy since the current garrison must only change loyalty.

For the splitter civs: This is an option in userprofile.txt, but can also set at the start of a new game. However, the new value applies for all game: New and existing.

> Actually I would find it most interesting to not give the city to the Barbarians but a new civ. That way it could inherit the state (advances, age, etc) from its former civ (before the conquering) and could built units comparable to those before the conquering.

That's a bit tricky, since when you decide you already own the city, here you would grant the city independence and it would inherit your stuff. Otherwise you have to save the former owner.

> Would those be splitter civs like mentioned in [fae7f93](https://github.com/civctp2/civctp2/commit/fae7f93c733d822abe848687b6a4bf0cfbf69a22)? However, I've never observed that such a new splitter civ turned up in a game.

Well, again that are your settings.

Anyway, this can go in. If you wanna address any of the issues you brought up then open a new issue or a new pull request.

LynxAbraxas:
The fixes in this PR also come into play when giving a city to another civ:

ss_2019-11-17_00:24:32

and works as expected:

ss_2019-11-17_00:25:47

LynxAbraxas:
> For the splitter civs: This is an option in userprofile.txt, but can also set at the start of a new game. However, the new value applies for all game: New and existing.

@MartinGuehmann what option is that in `userprofile.txt`? I can only find `NonRandomCivs`, which does not match concerning its description.

PR #202, opened 2019-10-20
Indicate how many units have got their order for current turn in progress bar

LynxAbraxas:
This PR adds an indication of how many units have got their order for current turn in a progress bar:

ss_2019-10-20_20:00:25

So far the progress bar was only used to display the progress of the other civs, i.e. when it was not the human players turn to play, and was black (unused) when it is the human players turn to give order to units/armies.


LynxAbraxas:
@MartinGuehmann Any idea how to not count units/armies that follow order from former truns, e.g. those that are fortified or sleeping? Currently the progress bar is already filled by those when a new turn begins.

MartinGuehmann:
> @MartinGuehmann Any idea how to not count units/armies that follow order from former truns, e.g. those that are fortified or sleeping? Currently the progress bar is already filled by those when a new turn begins.

Well, that's a bit tricky. You have to count the armies at the begin-turn that need orders. That is also a variable that must be saved in the save game. And then you can calculate the value you want. But here comes the tricky part: If I unfortify an army then it needs new orders, so that has then to modify the initial value.

Now you can hack it a bit more or just leave it as it is.

However, in both cases the progress bar needs a proper tooltip message, explaining what it is about, e.g.: Percent of armies that have orders for this turn, or percent of armies that need and got new orders this turn.

LynxAbraxas:
> That is also a variable that must be saved in the save game.

True. I guess it only makes sense to try to implement that additional feature if there is still some unused variable that could be used for storing that value.

By the way I noticed that the file saved by quick-save is smaller than that generated by a regular save, i.e. where you can specify the file name. @MartinGuehmann is that correct, i.e. expected?

MartinGuehmann:
> By the way I noticed that the file saved by quick-save is smaller than that generated by a regular save, i.e. where you can specify the file name. @MartinGuehmann is that correct, i.e. expected?

Yes, that is expected. The regular saves contain the world map as an image, while the quick and autosaves saves don't.

LynxAbraxas:
I figured a simple way that does not need saving a variable is to count just those armies that are sleeping or entrenched. With https://github.com/civctp2/civctp2/pull/202/commits/10e8ba69b814e755bb5cb5408c8a1fb70012dc10 the dark player color indicates those:

ss_2019-10-25_21:47:36

Activating some fortified units in a city has the expected effect on the progress bar:

ss_2019-10-25_21:49:00

LynxAbraxas:
> However, in both cases the progress bar needs a proper tooltip message, explaining what it is about, e.g.: Percent of armies that have orders for this turn, or percent of armies that need and got new orders this turn.

Hm, but where to but all that text? It already has a tooltip. Is it possible to swithch that depending on the context? On the other hand: I'd think most gamer will understand what the bar indicates without a tooltip.

MartinGuehmann:
> > However, in both cases the progress bar needs a proper tooltip message, explaining what it is about, e.g.: Percent of armies that have orders for this turn, or percent of armies that need and got new orders this turn.
> 
> Hm, but where to but all that text? It already has a tooltip. Is it possible to swithch that depending on the context? On the other hand: I'd think most gamer will understand what the bar indicates without a tooltip.

What does it already say? Anyway, with the dark and light player color, it is already better to understand. So that might be already good enough.

LynxAbraxas:
> What does it already say?

Actually it's empty.
Tested for a while and otherwise good to go in.

MartinGuehmann:
Shouldn't it be: "Armies with order**s**"?
And: "Truppen mit Befehlen"?
It could also be: "Armeen mit Befehlen". The translation of "army" is not consistent.

LynxAbraxas:
Done. Can go in unless you have an idea where the `Update()` could be placed instead. However I did not notice any performance drop with the current implementation.

MartinGuehmann:
> Done. Can go in unless you have an idea where the `Update()` could be placed instead. However I did not notice any performance drop with the current implementation.

Well, that should be about the user interface, where speed is not so super critical. Of course you don't want to slow it down unnecessarily. But so far, the code looks good. For sure, if you may have 300 armies or so then it might slow down, but that should be minor.

So, let's merge.

PR #204, opened 2019-11-03
Avoid population-cheat by not using SettleSize of settler-units

LynxAbraxas:
So far it was possible to build an Urban Planner or a Sea Engineer costing only one population and ordering that to setlle in an existing city (possibly the same as the building city) increasing its size by the units `SettleSize` (3 by default). This way it was easy to cheat up the city sizes. With the PR settling in  an existing city only increase the population count by that what the building of the unit has cost in population (so far only 1).

### DIFF
--- a/ctp2_code/gs/gameobj/Player.cpp
+++ b/ctp2_code/gs/gameobj/Player.cpp
@@ -2928,7 +2928,7 @@ bool Player::SettleInCity(Army &settle_army)
 		if(settle_army[i].CanSettle(pos, true))
 		{
 			Unit c = g_theWorld->GetCity(pos);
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Works as expected now.

Joining just gives one pop:
ss_2019-11-07_22:21:41

ss_2019-11-07_22:21:49

Building new city gives still 3 pop:
ss_2019-11-07_22:39:53

ss_2019-11-07_22:40:36

LynxAbraxas:
Tested this for a while and seems to work as expected. @MartinGuehmann  What do you think? Ready for merge?

MartinGuehmann:
> Tested this for a while and seems to work as expected. @MartinGuehmann What do you think? Ready for merge?

Well, the code looks good, so it can go in.

PR #205, opened 2019-11-05
Message to receiver when a trade route is created, import tab in trade manager

LynxAbraxas:
This PR implements a message to the receiver when a trade route is created and gives an extra tab in the trade manger listing the imports, which has become of interest with #188.


LynxAbraxas:
Seems also to work as expected:

ss_2019-11-09_00:49:31

This new functionality also shows that there is a bug in the trade creation code, possible the same as #195 tries to solve. 

Here it seems a route carrying bananas to Llanelli was created and then deleted again within the same turn (without a message). The route is neither listed in the import tab nor is its bonus added to the city:
ss_2019-11-09_00:50:01

LynxAbraxas:
Similar situation here. There is 60 food bonus before the creation of a new route:

ss_2019-11-07_22:04:28

but after creation of a route (5 caravans removed) with giant squid (20 food bonus, showing up in the import tab) the city food bonus does not increase (nor in the next turn):

ss_2019-11-07_22:05:32

LynxAbraxas:
@MartinGuehmann  should the message from https://github.com/civctp2/civctp2/pull/205/commits/deffb530da2ff8bd95108fa43a5a7d01d8fdff0f only be issued for non-domestic trade?
Does it make sense to have a "break route" button in the import tab? It would give a means to cancel trade the AI instantiated in case the player wants to save the money instead of getting the good bonus.

LynxAbraxas:
NB: The diplomacy intelligence window also contains a 3-tab layout, which is defined slightly differnetly:
https://github.com/civctp2/civctp2/blob/408bbbaae58a7617d546ca6b2fd27b2d1c73eb64/ctp2_data/default/uidata/layouts/DiplomacyDetails.ldl#L108-L325

MartinGuehmann:
> @MartinGuehmann should the message from [deffb53](https://github.com/civctp2/civctp2/commit/deffb530da2ff8bd95108fa43a5a7d01d8fdff0f) only be issued for non-domestic trade?
> Does it make sense to have a "break route" button in the import tab? It would give a means to cancel trade the AI instantiated in case the player wants to save the money instead of getting the good bonus.

It would be good be notified of international trade routes. In fact the AI should create those, and I think they should also increase the diplomatic regard. But that may be another issue.

If you add the break route button then you need to explain the player what it brings to break a route. I don't think that this is not too obvious.

LynxAbraxas:
> Here it seems a route carrying bananas to Llanelli was created and then deleted again within the same turn (without a message). The route is neither listed in the import tab nor is its bonus added to the city:

That seems to happen if there are not enough caravans for the route and is solved with 1121f81a71d76c1a30faa9649735eab352fb2d8d.

Tested for a while and seems to work as expected:
ss_2019-11-19_22:11:39

LynxAbraxas:

MartinGuehmann:
> @MartinGuehmann ready for merge?

Looks good, so let's merge.

Issue #201, opened 2019-10-20
Working functionalities


LynxAbraxas:
This is actually not an issue (therefore closed) but intended as a list to collect info on working functionalities that got tested.

LynxAbraxas:
Action of a cyber ninja to incite a revolution seems to work correctly:

![ss_2019-10-16_22:11:35](https://user-images.githubusercontent.com/21012234/67158191-68e27a80-f335-11e9-9575-3dd4344698d3.png)

LynxAbraxas:
Action of a cyber ninja to plant a nuke seems to work correctly:

![ss_2019-10-16_22:14:19](https://user-images.githubusercontent.com/21012234/67158215-9b8c7300-f335-11e9-97f7-5e4821f98671.png)

Not sure if this should have any diplomatic effects, apparently it has not.

LynxAbraxas:
Destruction of a pirating ("intention" to pirate) unit also removes the pirating flag from corresponding trade routes:

![ss_2019-10-13_22:41:01](https://user-images.githubusercontent.com/21012234/68531655-da9a5c80-0314-11ea-8ffa-fe010dd68eac.png)
![ss_2019-10-13_22:41:26](https://user-images.githubusercontent.com/21012234/68531596-45976380-0314-11ea-85da-01d116dbbf77.png)
![ss_2019-10-13_22:43:01](https://user-images.githubusercontent.com/21012234/68531599-4b8d4480-0314-11ea-97a1-69ce03fe9217.png)

LynxAbraxas:
Counter-proposals from the AI seem to be very rare, but seem to happen:

![ss_2019-08-03_17:42:03](https://user-images.githubusercontent.com/21012234/69500404-0f083e00-0efb-11ea-842f-9f2896e6cdb6.png)

LynxAbraxas:
AI asking to stop an enacted embargo:

![ss_2019-11-14_22:31:51](https://user-images.githubusercontent.com/21012234/69827653-94039800-1218-11ea-9e5b-eaa33b117c7d.png)

LynxAbraxas:
Winning the game by a scientific victory works:
![ss_2020-01-03_20:39:26](https://user-images.githubusercontent.com/21012234/71747590-5bd32300-2e70-11ea-9835-8e6d27316391.png)
[...]

PR #214, opened 2019-11-16
Listing a standing embargo in the diplomacy manager

LynxAbraxas:
So far the embargo icon was only shown when at war but not when only an embargo was enacted.


LynxAbraxas:
Enacted embargos to other civs now shown:
ss_2019-11-28_19:30:37

LynxAbraxas:
Tested for a while and seems to work as expeced.

LynxAbraxas:
For the reference: https://github.com/civctp2/civctp2/issues/201#issuecomment-559583598

LynxAbraxas:

MartinGuehmann:
> Tested for a while and seems to work as expeced.
> @MartinGuehmann ready for merge?

If it works then let's merge.

PR #218, opened 2019-11-29
Improved  force field blending

LynxAbraxas:
This PR offers improved  force field blending.

Before:

ss_2019-11-29_22:15:49

with this PR:

ss_2019-11-29_22:09:22

### DIFF
--- a/ctp2_code/gfx/spritesys/UnitActor.cpp
+++ b/ctp2_code/gfx/spritesys/UnitActor.cpp
@@ -1443,20 +1443,13 @@ void UnitActor::DrawForceField(bool fogged)
 	Pixel16 *cityImage = g_tiledMap->GetTileSet()->GetImprovementData((uint16)which);
 
 	if (g_tiledMap->GetZoomLevel() == k_ZOOM_LARGEST) {
-	    if (fogged)
-		g_tiledMap->DrawBlendedOverlayIntoMix(cityImage, m_x + nudgeX, m_y + nudgeY, k_FOW_COLOR, k_FOW_BLEND_VALUE);
-	    else
-		g_tiledMap->DrawColorizedOverlayIntoMix(cityImage, m_x + nudgeX, m_y + nudgeY, 0x0000);
+	    g_tiledMap->DrawDitheredOverlayIntoMix(cityImage, m_x + nudgeX, m_y + nudgeY, fogged);
 	    }
 	else {
-	    if (fogged)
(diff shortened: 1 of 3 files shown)


LynxAbraxas:
With 26ad54d926062bf1bdf29121a60e670a20cf39f8:

ss_2019-11-29_21:18:37

which reveals there are "holes" in the force-fields of water cities.
@MartinGuehmann any idea if that is intended or an artifact?

MartinGuehmann:
> With [26ad54d](https://github.com/civctp2/civctp2/commit/26ad54d926062bf1bdf29121a60e670a20cf39f8):
> 
> ss_2019-11-29_21:11:27
> 
> which reveals there are "holes" in the force-fields of water cities.
> @MartinGuehmann any idea if that is intended or an artifact?

If you look for it in the original version you can also see it. I guess that was intended that this piece would have 100% transparency. I guess you could alter agecitystyle.txt so that it uses the land force fields. Just remove the ForceField part. Alternatively you could look for another force field type in the tile file with TileEdit. When I added this I thought this was the intended water force field, but it could also be the space force field from CTP1.

LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/218/commits/4bffd9f7d968d086a75696a2707ac75110225778
ss_2019-12-07_18:49:30
ss_2019-12-07_18:51:39
looks all fine.

LynxAbraxas:
With https://github.com/civctp2/civctp2/pull/218/commits/8bae348fc9835651b593d24a08e64db0909218b4 there is no hole in the force field:

ss_2019-12-07_22:17:31

In ctp1 it looked like this:

ss_2019-12-07_22:09:38

ss_2019-12-07_22:05:12

I'm not sure what the hole is supposed to indicate.
@MartinGuehmann shall we keep it as is or revert https://github.com/civctp2/civctp2/pull/218/commits/8bae348fc9835651b593d24a08e64db0909218b4?

MartinGuehmann:
> I'm not sure what the hole is supposed to indicate.

I don't know either. The force field looks like the one for space cities, but the order in the tile file does not make any sense. If I look at this with TileEdit then it is there in CTP1 and CTP2. Judging from the screenshot of CTP1, the underwater cities also use the graphics from the land city.

> @MartinGuehmann shall we keep it as is or revert [8bae348](https://github.com/civctp2/civctp2/commit/8bae348fc9835651b593d24a08e64db0909218b4)?

I would prefer to keep it like this. Since you removed the WIP label, I assume that I can just merge and thus will do it.

PR #220, opened 2019-12-01
UnitManager: Refresh UpkeepButton when Readiness slider changes, fix #219

### LINKED ISSUE #219: UnitManager: upkeep costs not refreshed when readiness changes
Currently when Readiness slider position changes, upkeep change does not get reflected on upkeep button. To see how the cost change, one has to either close unit manager and then bring it back or press the upkeep button (actually press it twice to keep the same unit, since single press changes the format from percent to total and vice versa).

DEVoytas:
Currently when Readiness slider position changes, upkeep change does not
get reflected on the upkeep button. To see how the cost change, one has
to either close unit manager and then bring it back or press the upkeep
button (actually press it twice to keep the same unit, since single
press changes the format from percent to total and vice versa).

With this change the military upkeep cost on upkeep button are
refreshed dynamically as the position of readiness slider changes.

Additionally code duplication is decreased by moving upkeep calculations
to the dedicated function.

fixes #219

### DIFF
--- a/ctp2_code/ui/interface/unitmanager.cpp
+++ b/ctp2_code/ui/interface/unitmanager.cpp
@@ -487,6 +487,34 @@ void UnitManager::UpdateTacticalList()
 	m_tacticalList->BuildListEnd();
 }
 
+static void UpdateUpkeepButton(ctp2_Button *butt, Player *pl, bool useTotalFormat)
+{
+	Assert(butt);
+	Assert(pl);
+
+	MBCHAR buf[k_MAX_NAME_LEN];
+
+	if(useTotalFormat) { // total format
+		pl->m_readiness->RecalcCost();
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Cool, thanks @DEVoytas for this PR, will test it.

LynxAbraxas:
Tested this and seems to work as expected now, many thanks @DEVoytas for the fix.
@MartinGuehmann  ready for merge?

MartinGuehmann:
I have a formatting thing: "updateUpkeepButton" is a function even so it starts lower case. In general, functions in the code start upper case and variables lower case. So far I only remember the builtutil_ prefixed functions and similar prefixed ones.

I would prefer it upper case unless it is wide spread.

DEVoytas:
Thank you for pointing that out @MartinGuehmann, changed to upper case as requested.
And also thanks for testing @LynxAbraxas.

MartinGuehmann:
OK, then we can merge.

LynxAbraxas:
Thanks @DEVoytas, more PRs on other issues are very welcome;-)

PR #225, opened 2019-12-14
Adjusting the tech tree

LynxAbraxas:
PR relating to #129 (and #190), trying to reoder some dependencies of the advance graph (tech tree) in order to have more advances between theocracy and fascism such that the AI longer stays with theocracy where clerics and televangelist can be built, which represent a very interesting game concept that currently is cut very short.


LynxAbraxas:
@MartinGuehmann should fascism even be made dependent on explosives or mass-media such that it is more likely that the AI comes up with televangelists or should televangelists be made available earlier?

LynxAbraxas:
And in regard to https://github.com/civctp2/civctp2/issues/129#issuecomment-559194400, should the default strategy have thocracy on top to further increase the chance for many clerics/televangelists?

MartinGuehmann:
> @MartinGuehmann should fascism even be made dependent on explosives or mass-media such that it is more likely that the AI comes up with televangelists or should televangelists be made available earlier?

That doesn't really make sense, because the AI would then switch to another government that offers more cities.

The alternative would be to put the televangelists into later government.

MartinGuehmann:
> And in regard to [#129 (comment)](https://github.com/civctp2/civctp2/issues/129#issuecomment-559194400), should the default strategy have thocracy on top to further increase the chance for many clerics/televangelists?

No, because the AIs don't use the default strategy. The default strategy is just a fallback strategy. The AIs use :

STRATEGY_MILITARIST_DEFAULT
STRATEGY_DIPLOMATIC_DEFAULT
STRATEGY_ECONOMIC_DEFAULT
STRATEGY_SCIENTIST_DEFAULT
STRATEGY_ECOTOPIAN_DEFAULT

And those have different orders of government.

LynxAbraxas:
> The alternative would be to put the televangelists into later government.

Yes, something like an modern form of theocracy. Would it be tricky to introduce an additional goverment or would it be better to adjust an existing one?

MartinGuehmann:
> > The alternative would be to put the televangelists into later government.
> 
> Yes, something like an modern form of theocracy. Would it be tricky to introduce an additional goverment or would it be better to adjust an existing one?

I don't think so, we could just add it to the end to the database, or put it even in between. The in between solution might be a bit problematic since we shift the database indices, which might be saved in the games. However, I would expect that the game would handle the index shift.

Alternatively, we could put them into Democracy, since televangelists are a US-American phenomenon. Actually, as far as I know you can also find them in South America.

LynxAbraxas:
Hm, or, even though not related to reality/history, into communism which so far suffers from interest in ctp2 due to its settings.

LynxAbraxas:
https://github.com/civctp2/civctp2/pull/225/commits/9d3ea46887c6fcfb61e6fa169932df78efeab352 such that "advanced undersea mines" are available before "undersea mega mines" as currently:
ss_2019-12-08_23:25:03

MartinGuehmann:
> [9d3ea46](https://github.com/civctp2/civctp2/commit/9d3ea46887c6fcfb61e6fa169932df78efeab352) such that "advanced undersea mines" are available before "undersea mega mines" as currently:
> ss_2019-12-08_23:25:03

That's a bit odd, but you can change what the advances enable. I also renamed this PR, because advance graph would only apply in this context to the poster, even so you are right that in computer science it is not a tree, but just a graph.

LynxAbraxas:
Have now a game where at least one of the AI civs chose Theocracy:
ss_2020-01-28_21:10:53
Let's see if it turns up with Clerics and uses them properly.

LynxAbraxas:
Even though there are now 2 civs who chose Theocracy neither of them has built Clerics for many turns (>40) 

ss_2020-02-10_22:34:19
ss_2020-02-10_22:35:12

even though they seem to be able to built them.
So the problem might not only be the time they stay in Theocracy but possibly also some setting that prevents AIs in Theocracy to choose to built Clerics.
@MartinGuehmann any idea? While I tested some turns with #190 the AI came up with Televangelists (and used them all right), but none of them was in Theocracy.

LynxAbraxas:
Tested also with #190 before Televangelists can be built. Clerics are built by the AI and used all right, but apparently not if the AI is in Theocracy. So far I could not find any setting that would prevent the building of Clerics for AIs in Theocracy, but in any case that would not be part of this PR but perhaps rather of #190.

LynxAbraxas:
https://github.com/civctp2/civctp2/pull/225/commits/e1d5787806ff860c18a0c5706a46ea5bbc3c4b96 is reverted because it is difficult to propagate the changes back. However the reduced graph from GV's `tred` is helpful in case other tech tree modifications are done in future. An advance can only depend on 2 advances, with tred unnecessary direct dependencies are removed, making space for additional other possible dependencies (that are not implied).

LynxAbraxas:

MartinGuehmann:
Could you still update the GreatLibrary.txt files? Shall I also merge in the Polish text files of #278 since you would also like to update those? Originally, I wanted to make sure that I have the same strings in these files before, I merge them in.

LynxAbraxas:
> Could you still update the GreatLibrary.txt files?

In which way? The changed dependencies are already reflected in the in-game views of the tech tree.

> Shall I also merge in the Polish text files of #278 since you would also like to update those?

No, better someone with knowledge of Polish should do that in the scope of #278;-)

MartinGuehmann:
> > Could you still update the GreatLibrary.txt files?
> 
> In which way? The changed dependencies are already reflected in the in-game views of the tech tree.

That is not the only place where that matters. All these advances have the Requires fields:

[ADVANCE_ADVANCED_COMPOSITES_PREREQ]
Requires:
<L:DATABASE_ADVANCES,ADVANCE_CONSERVATION>Conservation<e>
<L:DATABASE_ADVANCES,ADVANCE_SUPERSONIC_FLIGHT>Supersonic Flight<e>
[END]

> 
> > Shall I also merge in the Polish text files of #278 since you would also like to update those?
> 
> No, better someone with knowledge of Polish should do that in the scope of #278;-)

That's just moving around some entries, so would would yo need to understand some Polish for that?

LynxAbraxas:
> > > Could you still update the GreatLibrary.txt files?
> > 
> > 
> > In which way? The changed dependencies are already reflected in the in-game views of the tech tree.
> 
> That is not the only place where that matters. All these advances have the Requires fields:
> 
> [ADVANCE_ADVANCED_COMPOSITES_PREREQ]
> Requires:
> <L:DATABASE_ADVANCES,ADVANCE_CONSERVATION>Conservation
> <L:DATABASE_ADVANCES,ADVANCE_SUPERSONIC_FLIGHT>Supersonic Flight
> [END]

OK, so now 8220f95fce4c7721589360c6990e350056a12207 contains the automatically determined advance prerequisites as defined in Advance.txt and fdcc0cc18a8bb3007bf247029c66db0515d343db reflects the changes made in the dependencies.
Similar to https://github.com/civctp2/civctp2/pull/321#issuecomment-646981947 this also contains the changes for awk to work properly (f675a2d7f4377d1aa7c70ad0cec856665c322361, c3be28896e262d2dc79c29d2eea1293da1211eb1, e8881e400d1ee97d710af4777005f407c3a5e696), which needs to be tested if that has any negative effect on windows. @MartinGuehmann could you do that? Otherwise ready for merging?

MartinGuehmann:
The line endings don't seem to change anything on Windows. So this can go in.

PR #227, opened 2020-01-03
Increase MaxRadius and MaxNeeded for power satellite

LynxAbraxas:
This way the science victory (SV) can be achieved with less agression because the oblelisk range can be further increased by building more power sats:
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_code/gs/gameobj/gaiacontroller.cpp#L760-L769
This is particularly a problem when playing with huge or gigantic maps.
Before:
ss_2020-01-01_21:32:20
with the PR:
ss_2020-01-01_22:22:32

### DIFF
--- a/ctp2_data/default/gamedata/EndGameObjects.txt
+++ b/ctp2_data/default/gamedata/EndGameObjects.txt
@@ -15,7 +15,7 @@ ENDGAME_POWER_SATELLITE {
     Building IMPROVE_POWER_SATELLITE
 	
 	MinNeeded 10
-	MaxNeeded 20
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
That should be straight forward. In principle, all civilizations should be able to achieve the science victory, however still under all circumstances the big civilizations have an advantage here. And that's true for all kind of science victories we have in civ games. Actually, this coverage condition makes it much worse, but that is a design decision from the time when Activision developed the game.

So let's merge.

Issue #209, opened 2019-11-09
Allow to expell dipomats only if at war


LynxAbraxas:
In relation to the examplary case of #208:
Since #177, dipomats cannot be expelled any more, so either the message of attacking diplomats should be change or not issued for diplomats. 
Probalbly the best would be to realize that a diploma can only be expelled if being at war with the dipomat's civ.

LynxAbraxas:
Or give diplomats the ability to expell other diplomats.

PR #232, opened 2020-01-03
make Emphatic Diplomats un-expelleable

LynxAbraxas:
missing from 6b68e4ae:
ss_2019-12-31_00:11:46

### DIFF
--- a/ctp2_data/default/gamedata/Units.txt
+++ b/ctp2_data/default/gamedata/Units.txt
@@ -1504,7 +1504,6 @@ UNIT_EMPATHIC_DIPLOMAT {
    MaxFuel 0
    IgnoreZOC
    NoZoc
-   CanBeExpelled
    CantCaptureCity
    EstablishEmbassy
    ThrowParty


LynxAbraxas:
It might be interesting to give (empahtic) diplomats the special action to expell other diplomats. Currently there is no way of getting rid of diplomats blocking e.g. roads other than killing them. 
The change of  #177 did make a real difference concerning the diplomacy interactions of AI civs, so reverting that would not be a good solotion to the above problem.

LynxAbraxas:
Another option might be to give (empahtic) diplomats the `CanBeSued` tag. However that would not solve the above problem as long as one cannot built lawyers, so giving  (empahtic) diplomats a similar ability might be the best solution.

MartinGuehmann:
> It might be interesting to give (empahtic) diplomats the special action to expell other diplomats (see also #209). Currently there is no way of getting rid of diplomats blocking e.g. roads other than killing them.
> The change of #177 did make a real difference concerning the diplomacy interactions of AI civs, so reverting that would not be a good solotion to the above problem.

It is indeed a problem that right now we cannot get rid of Diplomats except by killing them. Actually, does it happen that they block roads? I would assume that they would just move to their next task.

I think we should go for the expel by other Diplomats option.

Anyway, for now it is just consequent to make Emphatic Diplomats unexpellable. So let's merge.

LynxAbraxas:
> Actually, does it happen that they block roads? 

Possibly not intentionally, but I have had the situation that a diplomat was stationary on one of my main meglevs for many turns and really causing me troubles because the way around it did cost some essential moving points.

MartinGuehmann:
> > Actually, does it happen that they block roads?
> 
> Possibly not intentionally, but I have had the situation that a diplomat was stationary on one of my main meglevs for many turns and really causing me troubles because the way around it did cost some essential moving points.

Well, then we have to think here about a solution.

LynxAbraxas:
OK, made a note here: https://github.com/civctp2/civctp2/issues/209#issuecomment-560128802.

PR #241, opened 2020-01-05
Use/test run.sh in GL-CI

LynxAbraxas:
So far `run.sh` is well tested to work for running CTP2 through the GL-DI (docker image). Even though GL-CI testing uses the GL-DI, commands from `run.sh` are used (doublicated) in `.gitlab-ci.yml` because Xorg and Xvfb need different settings.
This PR is meant for making use of `run.sh` in GL-CI in order to avoid dublication and to test  `run.sh` in the scope of GL-CI.


LynxAbraxas:
OK, this seems to work now.
@MartinGuehmann Ready for merge?

MartinGuehmann:
Then we merge.

PR #23, opened 2018-11-05
Dockerfile and .travis.yml to test docker build

LynxAbraxas:
All-in-one-commit PR as an alternative to https://github.com/civctp2/civctp2/pull/13.

PR #242, opened 2020-01-05
Docker4master

LynxAbraxas:
This PR is only meant as a reference for #13 (which used `LynxAbraxas:master` instead of `LynxAbraxas:docker4master`) and was closed instantly because #23 was merged in favor of the approach taken here.
If needed, the commits of this PR are basically the history of https://github.com/civctp2/civctp2/pull/23/commits/a1605044108202121c10e5042d1150292e3d5efc.


LynxAbraxas:
Closed instantly because #23 was merged in favor of the approach taken here.

Issue #144, opened 2019-06-01
const cleanup


LynxAbraxas:
Only issue left from #1, see https://github.com/civctp2/civctp2/issues/1#issuecomment-494090167.

MartinGuehmann:
const cleanup refers to missing const keywords in the code. Just to clarify that, so that nobody has to follow the links.

MartinGuehmann:
This should be handled by 709199abf09c63b80bc1467f0850a124e4f2f266. Originally, it was probably also about compiler warnings.

Issue #85, opened 2019-02-03
Convert old branches to tags


LynxAbraxas:
There are still quite a few branches on https://github.com/civctp2/civctp2/branches, some are very old and remainders of the SVN import. To avoid others basing contribution on those but to still keep a ref on them (to avoid loosing the commits), I would suggest to convert the branches:
```
RolandTaverner
ScottGrant
linux-merge
linux-rework
port-trunk
trunk
```
into tags (possibly without a message, such that they are not regarded as releases by GH). This might even be appropriate for:
```
SaveGameConverter
linux
```
as they can be converted back to a brach if necessary.

MartinGuehmann:
I just don't know how to convert them to tags, I can create tags on GitHub and I can delete branches on GitHub, too.

For sure those can go:

```
ScottGrant
linux-merge
linux-rework
port-trunk
trunk
```
I think ScottGrant just contains one commit. The linux and port trunk are linux branch merge attempts. Then we still have:

```
RolandTaverner
SaveGameConverter
linux
```
For the linux branch, I still don't know whether I have everything, but that seems to be of low importance. Then there is something left from the RolandTaverner branch. Basically, that are the Visual Project files, that might contain something. There seems to have happened something with them. In master, we have the project files in I converted from my VS 2008 project files, the ones in the branch RolandTaverner have been created independently.

Unfortunately, the project configurations are in a different order so that I cannot just compare them with a merge/diff program. So the task would be to reorder the stuff so that I can compare, and then pick what seems to be good. That would be at least one final commit to that branch. But it has also low priority.

Then we have SaveGameConverter, with pull request #89 in mind I wouldn't be so sure if nothing would come anymore.

LynxAbraxas:
> I just don't know how to convert them to tags, I can create tags on GitHub and I can delete branches on GitHub, too.

Basiacally you would create a tag pointing to the corresponding branch tip first, then delete the branch, using the GH UI is probably the best.

MartinGuehmann:
Through GitHub it is a bit ugly, basically you have to create a release and if you wanna have it just a tag then you delete the release and the tag is left. For now I created tags for the following branches:

```
linux-merge
linux-rework
port-trunk
trunk
```

I still wanna look at ScottGrant's branch.

LynxAbraxas:
[...]

PR #243, opened 2020-01-05
Docker4linux

LynxAbraxas:
This PR is the equivalent to #242 (#13) but for the linux branch.


LynxAbraxas:
Closed instantly because the linux branch is basically fully integrated into master now: #1 (see also #85 and #144).

PR #244, opened 2020-01-11
Save known CityWall and ForceField 

LynxAbraxas:
So far it seems that known City-Walls and Force-Fields are not saved such that when re-loading a game cities in FOW are always drawn without City-Walls/ Force-Fields. 
Force-Field seen and saved in unseen cell:
ss_2020-01-11_21:55:08
After a save (quick or normal) and a re-load the force field is not drawn any more:
ss_2020-01-11_22:01:45

### DIFF
--- a/ctp2_code/gfx/spritesys/UnitActor.cpp
+++ b/ctp2_code/gfx/spritesys/UnitActor.cpp
@@ -338,11 +338,6 @@ void UnitActor::Initialize(void)
 
 	m_hidden = FALSE;
 
-	m_isFortifying = FALSE;
-	m_isFortified = FALSE;
-	m_hasCityWalls = FALSE;
-	m_hasForceField = FALSE;
-
 	m_shieldFlashOnTime = 0;
 	m_shieldFlashOffTime = 0;
 
--- a/ctp2_code/gfx/spritesys/UnitActor.cpp
+++ b/ctp2_code/gfx/spritesys/UnitActor.cpp
@@ -2323,10 +2323,10 @@ void UnitActor::Serialize(CivArchive &archive)
 		}
 
 		archive << m_facing;
-		archive.PutUINT8((uint8)m_isFortified);
-		archive.PutUINT8((uint8)m_isFortifying);
-		archive.PutUINT8((uint8)m_hasCityWalls);
-		archive.PutUINT8((uint8)m_hasForceField);
+		archive.PutUINT8((uint8)m_isFortified); // was not saved before #244 and is only saved here to fill the fromer 4 bytes, i.e. slot could be reused
+		archive.PutUINT8((uint8)m_isFortifying); // was not saved before #244 and is only saved here to fill the fromer 4 bytes, i.e. slot could be reused
+		archive.PutUINT8((uint8)m_hasCityWalls); // without saving, known city walls are not drawn after a game reload (bad for sending armies, slavers)
+		archive.PutUINT8((uint8)m_hasForceField); // without saving, known force fields are not drawn after a game reload (bad for sending armies)
 		archive << m_size;
 		archive.PutUINT8((uint8)m_isUnseenCellActor);
 
--- a/ctp2_code/gfx/spritesys/UnitActor.cpp
+++ b/ctp2_code/gfx/spritesys/UnitActor.cpp
@@ -2328,7 +2328,10 @@ void UnitActor::Serialize(CivArchive &archive)
 		}
 
 		archive << m_facing;
-		archive << m_lastMoveFacing;
+		archive.PutUINT8((uint8)m_isFortified);
+		archive.PutUINT8((uint8)m_isFortifying);
+		archive.PutUINT8((uint8)m_hasCityWalls);
+		archive.PutUINT8((uint8)m_hasForceField);
 		archive << m_size;
 		archive.PutUINT8((uint8)m_isUnseenCellActor);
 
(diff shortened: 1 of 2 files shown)


LynxAbraxas:
@MartinGuehmann What is `m_lastMoveFacing` needed for? Can we re-use its slot to save the bools? Not sure if `m_isFortified` and `m_isFortifying` are needed but I used them in  efad3488713acc4fa2d26f88a1462115cd796ca9 to fill the 4 bytes.

LynxAbraxas:
@MartinGuehmann What is `m_lastMoveFacing` needed for? Can we re-use its slot to save the bools? Not sure if `m_isFortified` and `m_isFortifying` are needed but I used them in  efad3488713acc4fa2d26f88a1462115cd796ca9 to fill the 4 bytes.

LynxAbraxas:
Hm, https://github.com/civctp2/civctp2/commit/efad3488713acc4fa2d26f88a1462115cd796ca9 seems not to solve this issue.
@MartinGuehmann any idea where the problem might lie?

LynxAbraxas:
Same for city walls. When newly seen, city wall is drawn:
ss_2020-01-12_20:10:25
then moving away such that city is in FOW, city walls are still drawn:
ss_2020-01-12_20:15:42
but when reloading the game at that state the city walls in FOW are not drawn any more:
ss_2020-01-12_20:14:17

@MartinGuehmann Any suggestions what I could try to solve this? Is efad3488713acc4fa2d26f88a1462115cd796ca9 correct in principle?

MartinGuehmann:
> @MartinGuehmann Any suggestions what I could try to solve this? Is [efad348](https://github.com/civctp2/civctp2/commit/efad3488713acc4fa2d26f88a1462115cd796ca9) correct in principle?

Check out void UnitActor::Initialize(void). I think it even overwrites stuff that already has been initialized in the constructor.

LynxAbraxas:
Lovely, now works as expected, many, many thanks @MartinGuehmann for pointing that out. I had not seen that overriding.
Now, is it OK to use the slot of `m_lastMoveFacing` for this? To my understanding  `m_lastMoveFacing` is only meant to recover the last facing position of the unit. If that is correct, I guess we can well live without that being saved. Then, is there something more important to save with 8 or 16 bit than `m_isFortified` `m_isFortifying`?

MartinGuehmann:
> Lovely, now works as expected, many, many thanks @MartinGuehmann for pointing that out. I had not seen that overriding.
> Now, is it OK to use the slot of `m_lastMoveFacing` for this? To my understanding `m_lastMoveFacing` is only meant to recover the last facing position of the unit. If that is correct, I guess we can well live without that being saved. 

I think this is correct. However, if I remember correctly this only applies to CTP1. In CTP2, the units face into the default position, which I guess is a bug and should be fixed. Actually, I wonder what the deference between m_facing and m_lastMoveFacing is. I guess these don't need 32 bits to be saved.

> Then, is there something more important to save with 8 or 16 bit than `m_isFortified` `m_isFortifying`?

In principle those can be even saved with one bit.

LynxAbraxas:
> Actually, I wonder what the deference between m_facing and m_lastMoveFacing is.

Me too.

> I guess these don't need 32 bits to be saved.

True, I'd think sint8 (or uint8) would be sufficient since the max seems to be
https://github.com/civctp2/civctp2/blob/e404a9abe18ce22ddfc4c12ed7f6b0c24bd0ba5f/ctp2_code/gfx/spritesys/FacedSprite.h#L49
which makes sense since sprite files have 5 and the 3 others are created by mirroring.
However sint32 is used throughout the code for `m_facing` and similar.

MartinGuehmann:
> > Actually, I wonder what the deference between m_facing and m_lastMoveFacing is.
> 
> Me too.
> 
> > I guess these don't need 32 bits to be saved.
> 
> True, I'd think sint8 (or uint8) would be sufficient since the max seems to be
> https://github.com/civctp2/civctp2/blob/e404a9abe18ce22ddfc4c12ed7f6b0c24bd0ba5f/ctp2_code/gfx/spritesys/FacedSprite.h#L49
> 
> 
> which makes sense since sprite files have 5 and the 3 others are created by mirroring.
> However sint32 is used throughout the code for `m_facing` and similar.

In principle, you could reduce the memory size of UnitActor objects, by reducing the memory requirements and reordering the variables then you just have to restore savegame compatibility. Is just the question how much this is worth. You don't have thousands of UnitActor objects in one game but maybe just a few hundred. So compared to the Goal objects that's not much.

Of course, you can view it as exercise for optimizing memory usage, and you may also find all the gaps were you could squeeze in more stuff into a savegame.

LynxAbraxas:
> Of course, you can view it as exercise for optimizing memory usage, and you may also find all the gaps were you could squeeze in more stuff into a savegame.

Yes, but am lacking the time for this exercise;-) and would rather spent that which I have on more pressing matters. Since this PR seems to work as expected, I'd say it can go in and if more slots for saving some vars are needed, there are the comments that point to this PR for further details.

MartinGuehmann:
> > Of course, you can view it as exercise for optimizing memory usage, and you may also find all the gaps were you could squeeze in more stuff into a savegame.
> 
> Yes, but am lacking the time for this exercise;-) and would rather spent that which I have on more pressing matters. Since this PR seems to work as expected, I'd say it can go in and if more slots for saving some vars are needed, there are the comments that point to this PR for further details.

Well, since you are happy with it we can merge.

PR #246, opened 2020-01-13
Don't show city capture options for cities with 1 pop

madand:
When capturing a city with population of 1, it is automatically razed yet
the City Capture Options dialog is still shown. The dialog buttons can be
clicked but it has no effect, since city is already destroyed.

Note: the observed behavior is with AE Mod scernario. I haven't played vanilla
for decades, so I'm not entirely sure that cities lose population after being
captured in vanilla game.

### DIFF
--- a/ctp2_code/gs/gameobj/CityEvent.cpp
+++ b/ctp2_code/gs/gameobj/CityEvent.cpp
@@ -201,7 +201,7 @@ STDEHANDLER(CaptureCityEvent)
 		if (
 			(g_theProfileDB->IsCityCaptureOptions())
 			//(g_theProfileDB->GetValueByName("CityCaptureOptions"))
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
> When capturing a city with population of 1, it is automatically razed yet
> the City Capture Options dialog is still shown. The dialog buttons can be
> clicked but it has no effect, since city is already destroyed.
> 
> Note: the observed behavior is with AE Mod scernario. I haven't played vanilla
> for decades, so I'm not entirely sure that cities lose population after being
> captured in vanilla game.

Yes, they do and are destroyed, when the population would drop to zero. However, the population is not at zero, when this code runs.

In fact these things E added are not really well tested.

Your fix should be very straight forward and thus can be merged.

Issue #192, opened 2019-10-11
Effect of pirating on city trade good bonuses


LynxAbraxas:
This issue is for discussing if and how pirating should effect the city trade good bonuses introduced with #188.

LynxAbraxas:
An implementation that takes pirating into account should be possible but might be a bit chellanging: https://github.com/civctp2/civctp2/pull/188#issuecomment-538781772

LynxAbraxas:
While taking pirating into accound would make the concept of pirating even more interesting, this would basically also demand that the AI is made aware of this concept of weakening other player's cites, and on the other hand the AI should become upset about pirating routes it receives a bonus from. So far only the sender (apparently) becomes upset (@MartinGuehmann do you know if that is actually taken into account in the code?).

LynxAbraxas:
Without the implementatio of loosing the bonus, pirating on domestic trade routes has no effect for the sender==receiver. Oddly, it seems the pirating civ gets the gold value of that route which is otherwise not taken into account for domestic trade:
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_code/gs/gameobj/CityData.cpp#L4463-L4475

It might be appropriate to give the full or a partial route value to the civ of the domestic trade in case of implementing #215.

LynxAbraxas:
> While taking pirating into accound would make the concept of pirating even more interesting, this would basically also demand that the AI is made aware of this concept of weakening other player's cites, and on the other hand the AI should become upset about pirating routes it receives a bonus from. So far only the sender (apparently) becomes upset (@MartinGuehmann do you know if that is actually taken into account in the code?).

In addition the AI should be taught to not pirate a route to one of its own cities, like here:
![ss_2019-12-16_23:23:25](https://user-images.githubusercontent.com/21012234/71752608-fb4be200-2e7f-11ea-9990-beeb672bc374.png)

LynxAbraxas:
While #249 implements that the bonus is lost for the receiver in case of pirating, the issue that the AI should not pirate routes it is a receiver of (i.e. gets the bonus) is currently still unresolved. Ideas on such an implementation are discussed here https://github.com/civctp2/civctp2/pull/259#issuecomment-581062571. 
[...]

PR #249, opened 2020-01-16
Pirate also the bonus given to a trade route destination

LynxAbraxas:
This PR solves #192:
ss_2020-01-16_21:04:35
ss_2020-01-16_21:05:22

### DIFF
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -190,6 +190,7 @@
 #include "profileDB.h"
 #include "radarmap.h"
 #include "RandGen.h"                    // g_rand
+#include "ResourceRecord.h"             // g_theResourceDB
 #include "RiskRecord.h"
 #include "Scheduler.h"
 #include "SelItem.h"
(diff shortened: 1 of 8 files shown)


LynxAbraxas:
So what is left is to decide if the pirating civ should get the bonus. If so, how to determine which city should get it?
@MartinGuehmann Is it possible to determine the city that built one of the pirating units?

MartinGuehmann:
> So what is left is to decide if the pirating civ should get the bonus. If so, how to determine which city should get it?

Which city is easy: Just the closest.

> @MartinGuehmann Is it possible to determine the city that built one of the pirating units?

This is actually a Civ2, where the city had to pay shields for its units. I doubt that this is saved in CTP2. However, if it were saved, then you should find it in the UnitData object, it should be the unit parent city ID or something like that. But there is nothing like this.

LynxAbraxas:
> Which city is easy: Just the closest.

True, I like that idea. Probably the code that determines where to send a unit (expell, enslaving) could be used (reduced by the check if the destination can hold another army/unit).
However, after investigating this a bit, I wonder how to pass the captured bonus to the closest city. Currently I can only think of storing this in a variable of `CityData` which then gets read when calculating food, prod. and gold during turn transition, but that would need to be saved in a save-game, since there we cannot use the `m_tradeDestinationList` or similar because the city itself is in general not involved in the trade route connection.

LynxAbraxas:
> how to pass the captured bonus to the closest city.

@MartinGuehmann Any ideas on realizing this without the need for another slot in the save-game format?

MartinGuehmann:
> > how to pass the captured bonus to the closest city.
> 
> @MartinGuehmann Any ideas on realizing this without the need for another slot in the save-game format?

I think you have to save it, otherwise it gets messed up after a reload. And looping through all the trade routes of all players checking whether they are pirated and then figuring out what the nearest city is, is just brute force and would really get horrible with more trade routes and cities.

So m_buildInfrastructure and m_buildCapitalization are not used and are stored as BOOL in the save game, which gives you two sint32. Unfortunately these things are 32 bits so your new variables should be 32 bits, too. So now you have space for food and production, now you just need gold. Also m_terrainWasPolluted seems to be unused, and what the point of m_terrainImprovementWasBuilt and m_improvementWasBuilt is I don't know. Maybe these are not used, either.

LynxAbraxas:
Thanks for all these pointers, I'll give it  a try with those.

LynxAbraxas:
I'd say that passing the pirated bonus to the nearest city should be a PR on its own, created #257 to track that.

MartinGuehmann:
Then let's merge.

Issue #223, opened 2019-12-01
AI weakness partially due to loss of gold


LynxAbraxas:
It seems one remaining reason for the AI to be weaker than a human player is that the AI is very bad at book-keeping about its gold. This limits its choice on special order that cost gold.

Assuming that the AI optimize its cities concerning:
- citizen jobs
- improvements
- gold vs science ratio

Ways for the AI to increase its gold would be:
1. establishing more valuable trade routes (#155)
2. pirating more trade routes
3. build more buildings that increase gold
4. spent less gold on rush-buy?
5. give less gold (diplomatically) to other players

LynxAbraxas:
Getting many 
```
Assertion (0 <= delta) Failed in File:Gold.cpp, Line:215
```
with the debug version for [dbgSue.zip](https://github.com/civctp2/civctp2/files/3936133/dbgSue.zip).

LynxAbraxas:
> Ways for the AI to increase its gold would be:
> 
>     1. establishing more valuable trade routes (#155)

Done, with #252.

PR #251, opened 2020-01-19
WIP: Gold given even for domestic trade routes

LynxAbraxas:
In conjunction to https://github.com/civctp2/civctp2/issues/192#issuecomment-570700310, this PR enables receiving gold even for domestic trade routes, e.g. paid by the citizens from their wages.
Currently, domestic trade does not yield any gold
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_code/gs/gameobj/CityData.cpp#L4463-L4475
Therefore, before #188, domestic trade routes did not make any sense at all. Now with #188 the destination at least gets the bonus, and with #249 this bonus is lost when the route is pirated. The pirating civ so far does not get the bonus but the gold the route is worth.
It can be argued that the sending city gets the gold from the receiving city, and belonging to the same civ for domestic trade routes, not yielding any net gold to the civ. However, giving the civ some gold (not removed from another player in case of domestic trade) would have the positive effect that the AI (establishing many domestic trade routes) would gain more gold in regard to #223.

Suggestions and other ideas on this are welcome.

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -4473,6 +4473,9 @@ void CityData::CalculateTradeRoutes(bool projectedOnly)
 					g_player[route.GetSource().GetOwner()]->AddGold(route.GetGoldInReturn());
 				}
 			}
+			else{ // gold even for domestic trade route (e.g. paid by the citizens from their wages)
+			    g_player[m_owner]->AddGold(route.GetGoldInReturn());
+			    }
 
 			if(!route.IsActive())
 			{


LynxAbraxas:
In conjunction to https://github.com/civctp2/civctp2/pull/252#issuecomment-579869865, and since the current code does not discriminate between gold given for domestic and non-domestic trade
https://github.com/civctp2/civctp2/blob/ea3274c3d43176670ac71f6bc75703fac18a226e/ctp2_code/gs/gameobj/CityData.cpp#L4510-L4538
closed in favor for #255.

Issue #254, opened 2020-01-29
AI trade route creation should take bonus and the need of a city into account


LynxAbraxas:
While #252 improves the AI trade route creation concerning the trade profit (gold, one point of #223, #155), #252 does not include an implementation that the AI trade route creation takes the value of the good-bonus (#188) into account nor how much a target city actually needs some boni. E.g. a city in a glacier or mountain region should offer an extra price for goods which have a high food bonus. A human player can take that into account and accept very low `valuePerCaravan` in such cases for domestic trade routes but the AI currently cannot. For such additional value `GetGoldInReturn()` (apparently a remainder from CPT1, https://github.com/civctp2/civctp2/pull/252#issuecomment-579869865) 
https://github.com/civctp2/civctp2/blob/ea3274c3d43176670ac71f6bc75703fac18a226e/ctp2_code/gs/gameobj/TradeRouteData.h#L180
could be used in addition to currently only used `GetValue()`
https://github.com/civctp2/civctp2/blob/ea3274c3d43176670ac71f6bc75703fac18a226e/ctp2_code/gs/gameobj/TradeRouteData.cpp#L591-L599
adjusted in #252 
https://github.com/civctp2/civctp2/blob/ff46029ea9d91a06c4ad2face2616b30b2a76d82/ctp2_code/gs/gameobj/TradeRouteData.cpp#L591-L597
That would also allow to have one portion of the gold profit come from the trade itself (like the transitional trade #126) or be argued to be paid by the citizen (from their wages) of the target city and the other portion to be paid by the treasury, i.e. gold subtracted from the receiving player.

Please post ideas how to realize a balanced and fair rating of how much a bonus is worth to a city.

PR #255, opened 2020-01-30
Reduce trade route value for domestic trade

LynxAbraxas:
PR replacing #251, reducing the trade profit (gold) for domestic trade routes. Apparently, no gold was given for domestic trade in CTP1
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/gs/gameobj/CityData.cpp#L4460-L4475
but in CTP2:
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/gs/gameobj/CityData.cpp#L4513-L4541
This is fine in general, but with #254 currently leads to more domestic trade routes than before. This PR tries to counter act this by reducing the value of domestic trade by not taking the distance to a good into account in case of domestic trade, since the good is within the territory.

### DIFF
--- a/ctp2_code/gs/gameobj/tradeutil.cpp
+++ b/ctp2_code/gs/gameobj/tradeutil.cpp
@@ -58,10 +58,12 @@ sint32 tradeutil_GetTradeValue(const sint32 owner, const Unit & destination, sin
 	return 0;
 
     double baseValue = g_theWorld->GetGoodValue(resource); // good value varies between a min and a map depening on its frequency on the map
-    double distance = static_cast<double>(destination.GetCityData()->GetDistanceToGood(resource));
-    sint32 totalValue = sint32(baseValue * distance);
 
     PLAYER_INDEX const  tradePartner = destination.GetOwner();
+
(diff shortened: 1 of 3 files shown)


LynxAbraxas:
> This PR tries to counter act this by reducing the value of domestic trade by not taking the distance to a good into account in case of domestic trade, since the good is within the territory.

This does not work because `baseValue < 1`, so instead introduced a reduction coefficient, with 74e93f336bdbfd5f73643db4a68e2f7b8fcb5b78 domestic trade routes are valued as before:
ss_2020-01-31_00:08:17
but with 75527f90abbf0afeb62020940e4308464adcacc6 their value is halved but foreign trade is unaffected:
ss_2020-01-31_22:23:14

LynxAbraxas:
The value for `DOMESTIC_TRADE_COEF` in `Const.txt` might need to be adjusted in case #254 is implemented.

LynxAbraxas:

MartinGuehmann:
Well then, we can merge.

PR #256, opened 2020-01-30
Drawing trade routes according the the last discovered state (created or cancelled)

LynxAbraxas:
Currently trade routes are drawn globally even in FOW, except in unexplored tiles. This exposes never seen trade routes as well as their source and destination cities. Here the barbarian trade route gives away the position of an unseen city and that Arzawa has been conquered by barbarians (eventhough that was not the case when it was last seen):
ss_2020-01-30_16:38:31
similar here:
ss_2020-01-30_16:36:57

This exposure makes it very easy for a human player to send units for pirating to trade routes that would otherwise not be know or attack cities otherwise unknown.
This PR reduced the drawing of trade routes to those currently seen or as last seen, i.e. the actual trade route situation might be different to what is show in a unseen cell, similar to former and current borders.


LynxAbraxas:
This PR also needs changes to the tile-info, such that it does not give away more info than when last seen.

LynxAbraxas:
While 6f468b57a0950a675d8825813617b3a4ae26e2b9 and f0d6bce605c48ecaed6cca519b09f5f016595894 kind of work
ss_2020-01-30_17:00:23
ss_2020-01-30_17:03:14

the current changes do not allow to save what was seen in the unseen cell object. While the world map does save the trade routes for each cell (`g_theWorld->GetCell(p)->GetTradeRoute(i)`), the unseen cell object does not seem to have this.
@MartinGuehmann Do you see a chance to save this info also in the unseen cell object? Possibly, this could be reduced to just a flag which player has a trade route in this cell, such that it could be stored in a single `uint32` because `k_MAX_PLAYERS` is defined as 32.

LynxAbraxas:
> Possibly, this could be reduced to just a flag which player has a trade route in this cell, such that it could be stored in a single uint32 because k_MAX_PLAYERS is defined as 32.

This simplification however would demand that trade routes in unseen cells would only be indicated by e.g. a colored dot in the center of the tile because for the drawing of the current trade route segments the previous/next point of the route is needed (which would not be saved with such a simplification).

MartinGuehmann:
> @MartinGuehmann Do you see a chance to save this info also in the unseen cell object? Possibly, this could be reduced to just a flag which player has a trade route in this cell, such that it could be stored in a single `uint32` because `k_MAX_PLAYERS` is defined as 32.

You can do what I have done with m_visibleCityOwner, I squeezed it into the installation list. So you can do it here too.

Additionally, you still have a spot for 8 bits. I just have no idea what you could do with that.

Well, you could still connect the routes if the neighbor tile has a route, from the same player. It shouldn't matter which route it is.

LynxAbraxas:
> You can do what I have done with m_visibleCityOwner, I squeezed it into the installation list. So you can do it here too.

Many thanks for that cool idea. How long can the installation list be? Perhaps it would then be better to store a set of pointers to the routes of the global map that were seen. That of would of course need some research what these pointers point to if the route was killed in the mean time and that one can rule out that it would point to a possibly newly created one. I'm not sure at the moment if a route that is killed is removed from the global route list or just marked as gone.

> Additionally, you still have a spot for 8 bits. I just have no idea what you could do with that.

Not sure what you mean with that., What spot do you refer to?

> Well, you could still connect the routes if the neighbor tile has a route, from the same player. It shouldn't matter which route it is.

In some cases (e.g. branch points) this could matter, but all in all this should be the smaller problem to solve.

In general, do you think this change is a good idea or is it not fitting the general game concepts? I'm not sure in this case. Not seeing the above described details without going there makes it more challenging.
Do you think we will be able to "teach" this the AI? Even if, would it make it too difficult for the AI to make use of pirating regularly?

MartinGuehmann:
> > You can do what I have done with m_visibleCityOwner, I squeezed it into the installation list. So you can do it here too.
> 
> Many thanks for that cool idea. How long can the installation list be? Perhaps it would then be better to store a set of pointers to the routes of the global map that were seen. That of would of course need some research what these pointers point to if the route was killed in the mean time and that one can rule out that it would point to a possibly newly created one. I'm not sure at the moment if a route that is killed is removed from the global route list or just marked as gone.

I think as many as you want within the limits of the datatype, but be careful with that you have many UnseenCell objects: MapHeight * MapWidth * PlayersCurrentlyInTheGame. This is actually and upper bound, but just insert the values for ultra gigantic maps and have 32 players, and you are already using quite some memory with the current implementation in master.

> > Additionally, you still have a spot for 8 bits. I just have no idea what you could do with that.
> 
> Not sure what you mean with that., What spot do you refer to?

Check out the serialization code there you have 8 unused bits.

> In general, do you think this change is a good idea or is it not fitting the general game concepts? I'm not sure in this case. Not seeing the above described details without going there makes it more challenging.

That is the real question: Traders just can't shut their mouth, so they will tell you where their stuff is coming from. And you can still follow a trade route to find the trade partners. Actually, I tend to leave the display of trade routes as it is now. You are addressing an issue that is an issue in an unexplored world. In an explored world you would know the cities that are involved. It makes it also difficult to draw these trade routes properly, basically you are left with open ends on the maps, and that is getting very cluttering.

> Do you think we will be able to "teach" this the AI? Even if, would it make it too difficult for the AI to make use of pirating regularly?

Well, then you have to go for something that the can see, or it would go for something it thinks it sees, that makes it at least so that you have to modify the algorithm for the AI to find it.

LynxAbraxas:
> And you can still follow a trade route to find the trade partners. 

Yes, and that is what this PR is meant to promote, that you actually have to follow a trade route to find the trading cities (in case they are not known already). Currently the drawn trade route gives away this information without having to follow the route. That way a human player has an advantage, since the AI is not able to conclude that there must be a city at each route's end.
And even later in the game (if all cities are know and no cities get lost or newly created) it would be much more a challenge to pirate if one needs to find an active trade route first. Currently it is far too easy for a human player to figure out a hot-spot of trade routes (in the unseen regions) and go there and pirate many routes with just one unit.

MartinGuehmann:
> > And you can still follow a trade route to find the trade partners.
> 
> Yes, and that is what this PR is meant to promote, that you actually have to follow a trade route to find the trading cities (in case they are not known already). Currently the drawn trade route gives away this information without having to follow the route. That way a human player has an advantage, since the AI is not able to conclude that there must be a city at each route's end.

In principle, the AI knows where all the cities are. However, the ones it hasn't explored yet, get an unexplored bonus for most of the goals, for that you should know their location.

> And even later in the game (if all cities are know and no cities get lost or newly created) it would be much more a challenge to pirate if one needs to find an active trade route first. Currently it is far too easy for a human player to figure out a hot-spot of trade routes (in the unseen regions) and go there and pirate many routes with just one unit.

That would be the easiest thing to teach the AI. I think currently the AI goes for some pint that is close to some position. So the AI would just have to go for the spots with many trade routes. Of course, it shouldn't pirate accidentally routes of civs, it is friends with.

LynxAbraxas:
So far I had the following ideas how to realize this PR:

1. one sint32 for each unseen tile (e.g. `visibleTradeRouteOwners`, similar to city visibility), however that way no correct connections between tiles are possible, so seen trade routes in unseen tiles could only be drawn with a dot (as in the radar window) and the tile info could only give away the info of what civs have a trade route there but not how many. Adding an additional uint8 would allow to save the total number of trade routes but not the subtotals of each owner. This would demand:

2. a unit8 for each civ (i.e. 32 * uint8 per unseen tile) storing the # of trade routes of that civ. an additional unit8 for each civ (i.e. 2 * 32 * uint8 per unseen tile) would allow to save the seen directions of routes such that they could be drawn connected exactly as in the seen state.

3. one sint32 per trade route storing which civ has seen this route (e.g. `seenByCiv`, similar to city visibility). Basing the drawing on this still would reveal unseen cities but one could argue that the traders met on the route tell about the source and destination cites or their route, similar to gossip. So to be consistent the unseen cities could just be made visible, i.e. added to the state of the unseen cell. One problem of this approach is then still that trade routes cannot be drawn any more when the trade route is cancelled. However this is one of the points of the idea of this PR, it should force a human player to actually only get to know that a trade route is gone when at least one of its tiles are within a seen tile. Unseen tiles should just remember the last seen state. So to realize that trade routes could be marked as cancelled and when seen again in that state the bit in `seenByCiv` removed. However that way trade routes cannot be deleted until `seenByCiv` is zero. That way trade route data would accumulate for much longer than now but this still might be the least memory demanding solution.

4. dynamic array (smart) pointers to the seen trade routes running in an unseen tile could be stored. This is probably the most memory demanding solution, since also then trade routes could not be deleted until no unseen tile refers to them.

My current favorite is 3. @MartinGuehmann What do you think? Would the extended accumulation of trade route data be uncritical? Possible `isActive` of the trade route data could be used to store the state of being cancelled/killed. If there is no unused sint32 any more in the saved trade route data there would be the option to save it as the 2nd sint32 of https://github.com/civctp2/civctp2/pull/263/commits/7c3eecc52e1eb095c0f4dd54af74bcdf4fc5522d.

MartinGuehmann:
> My current favorite is 3. @MartinGuehmann What do you think? Would the extended accumulation of trade route data be uncritical? Possible `isActive` of the trade route data could be used to store the state of being cancelled/killed. If there is no unused sint32 any more in the saved trade route data there would be the option to save it as the 2nd sint32 of [7c3eecc](https://github.com/civctp2/civctp2/commit/7c3eecc52e1eb095c0f4dd54af74bcdf4fc5522d).

If the AI does not change so often its mind about trade routes then it should be uncritical. As alternative. It could still be shown that these routes are canceled. That shouldn't interfere too much with what you want, but of course it is just unlogical.

Also keep in mind that not only armies see trade routes, but also cities and GlobeSats.

What actually did isActive before? If it was unused then you should just reuse it if you need it.

LynxAbraxas:
Trying to realize 3. I am not sure where to place the code that sets `AddSeenByBit` if a tile of a trade route is within the vision range of a unit. In 56b1ab721edde836fda155810085cf5186669e70 I placed it in `ArmyData::CheckWasEnemyVisible` because there the visibility of cities is set but this seems not to work for the trade route (possibly because it is not a unit).
Newly created trade route visible to the owner (as intended)
ss_2020-02-15_23:15:18
but not to third party (as intended)
ss_2020-02-15_23:17:02
However, also not after being within vision
ss_2020-02-15_23:17:37
Visibility so far also not working for receiver
ss_2020-02-15_23:16:43

@MartinGuehmann Any idea where to better place `AddSeenByBit` than in `ArmyData::CheckWasEnemyVisible`? Or is the code missing something else?

LynxAbraxas:
Just noticed that `owner == paying_for != m_destinationCity->GetOwner()`, with
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/gs/gameobj/PlayerEvent.cpp#L556-L558
and
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/gs/gameobj/Player.cpp#L3071
Possibly, it should be `destCity.GetOwner()` in 
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/gs/gameobj/PlayerEvent.cpp#L556-L558
On the other hand `m_paying_for` might not be needed at all.

LynxAbraxas:
> However, also not after being within vision

Solved with a3536a8409170a42860650292f94f7a154bd30e0
While it works nearly as expected now, the changes of 9f89e832b7971f1be1ad348f1af5225e0384db2d for some reason prevent that tiles that are not any more within a units vision become unseen tiles:
ss_2020-02-16_16:43:37

@MartinGuehmann any idea why that happens?

LynxAbraxas:
OK, making the scope of the case statement local makes it work again (c6b7a8ae5d38bd695a30e009c4f5628850bf4556), still I'd be very thankful for an explanation why that is necessary now.

LynxAbraxas:
Is there something like an `EndOfAllTruns` where `TradeRoute::RemoveUnseenRoute` (from f6fd858a506858a9a238267d8030868996812523) could be placed?

MartinGuehmann:
> OK, making the scope of the case statement local makes it work again ([c6b7a8a](https://github.com/civctp2/civctp2/commit/c6b7a8ae5d38bd695a30e009c4f5628850bf4556)), still I'd be very thankful for an explanation why that is necessary now.

I don't know: It might be something found here: https://stackoverflow.com/questions/92396/why-cant-varniables-be-declared-in-a-switch-statement

However, what does not fit here is that your compiler should have complained. Anyway, with the label explanation, you should put any variable declared into its own scope.

LynxAbraxas:
Many thanks @MartinGuehmann for that pointer. I see, all  other variables were already initialized before the switch statement. A complaint by the compiler would definitely be helpful here...

MartinGuehmann:
> Many thanks @MartinGuehmann for that pointer. I see, all other variables were already initialized before the switch statement. A complaint by the compiler would definitely be helpful here...

Maybe there is some option that could enable strict checking.

LynxAbraxas:
In the current state, the concept is that only an *active* trade route exchanges goods, gives gold and bonus and can be pirated (like the original behavior). Only if it is seen when active it will be drawn on the map (FOW or not) for that player. However, instead of the former killing of the route it now gets deactivated (but not killed/removed). If a route is deactived, it will not be drawn any more on the map (FOW or not) for that player if visited again. If all civs that have seen that route also have visited it again after it was deactivated, the route can be eventually be removed completely (this still needs a proper placement of `TradeRoute::RemoveUnseenRoute` from f6fd858a506858a9a238267d8030868996812523).
Now the problem that I am facing is that all calls that iterate over the routes of a cell like
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/gs/gameobj/ArmyData.cpp#L1674-L1676
must check now if the route is active, like:
```
if(route.IsActive() && route->GetPiratingArmy().m_id == m_id) {
```
This made me wonder if it would be appropriate to define an iterator for trade routes with an option to iterate over all routes or only the active ones. The closest concept I could find so far would be https://stackoverflow.com/questions/774293/how-to-create-an-iterator-over-elements-that-match-a-derived-type-in-c but instead of a `dynamic_cast` check if the route is active. I have not worked with self-defined iterators before, so am wondering if that is likely introducing more hassle than it would solve and if it is actually possible to define an interator with such an option. @MartinGuehmann Do you have experience with that?
A search for `GetNumTradeRoutes` yields about 10 places in the code base that would need the addition of `route.IsActive()`, so perhaps it would be better to go for just adjusting those (despite being less elegant).

LynxAbraxas:
> Many thanks @MartinGuehmann for that pointer. I see, all other variables were already initialized before the switch statement. A complaint by the compiler would definitely be helpful here...

Read through the answers again and it is still not clear to me why scoping solved that issue since the variable `cell` is not used in `case CIRCLE_OP_SUBTRACT:` or `case CIRCLE_OP_ADD_RADAR:`. A complaint by the compiler would only be that it is not initialized in case of jumps to those labels but since `cell` is not used there the lack of its initialization should not have any effect on `case CIRCLE_OP_SUBTRACT:` or `case CIRCLE_OP_ADD_RADAR:`. 
@MartinGuehmann  Any other ideas how to explain this odd behavior?

MartinGuehmann:
> @MartinGuehmann Any other ideas how to explain this odd behavior?

Not really, I just googled for it and that was what I found. On the other hand this code has another variable initialized in one of case statements and the original code also put it into a block. So I assume it still has to do something with that I dug up. But I haven't understood it fully, either.

MartinGuehmann:
> This made me wonder if it would be appropriate to define an iterator for trade routes with an option to iterate over all routes or only the active ones. The closest concept I could find so far would be https://stackoverflow.com/questions/774293/how-to-create-an-iterator-over-elements-that-match-a-derived-type-in-c but instead of a `dynamic_cast` check if the route is active. I have not worked with self-defined iterators before, so am wondering if that is likely introducing more hassle than it would solve and if it is actually possible to define an interator with such an option. @MartinGuehmann Do you have experience with that?

Not really, but I would expect that you have to overload the ++operator or operator++.

LynxAbraxas:
> So I assume it still has to do something with that I dug up. But I haven't understood it fully, either.

Many thanks anyway for looking into this.

> Not really, but I would expect that you have to overload the ++operator or operator++.

Yes, and apparently quite a lot more, so decided not to go for the iterator for now, because the code now modified https://github.com/civctp2/civctp2/pull/256/commits/8bb3aa37b97f2a000b083b884296077a4473619b would also have to be modified to use the iterator.
Most thinks work now as expected, except that
https://github.com/civctp2/civctp2/pull/256/commits/6dccdf5419ba658103321c539355a113c5cfc3e3 leads to a segfault when it is some other civ turn: 
```
Thread 1 "ctp2" received signal SIGSEGV, Segmentation fault.
0x000055555608e20d in TradeRouteData::SeenBy (this=0x0, player=1)
    at TradeRouteData.cpp:164
warning: Source file is more recent than executable.
164		return m_seenBy & (1 << player);
(gdb) bt
#0  0x000055555608e20d in TradeRouteData::SeenBy (this=0x0, player=1)
    at TradeRouteData.cpp:164
#1  0x000055555608d163 in TradeRoute::SeenBy (this=0x7fffffffdfd0, player=1)
    at TradeRoute.cpp:212
#2  0x00005555558b1cb9 in RadarMap::RenderTrade (this=0x55555e30d66c, surface=
    0x55555e31070c, position=..., worldpos=..., player=0x555562cc91bc)
    at ui/aui_ctp2/radarmap.cpp:932
#3  0x00005555558b3355 in RadarMap::RedrawTile (this=0x55555e30d66c, 
    point=0x7fffffffe0a0) at ui/aui_ctp2/radarmap.cpp:1419
#4  0x000055555608e856 in TradeRouteData::GeneratePath (this=0x555573dbbcbc)
    at TradeRouteData.cpp:282
#5  0x000055555608d83d in TradeRouteData::TradeRouteData (this=0x555573dbbcbc, 
    route=..., source=..., dest=..., owner=6, sourceType=ROUTE_TYPE_RESOURCE, 
    sourceResource=29, paying_for=6, gold_in_return=0)
    at TradeRouteData.cpp:108
#6  0x000055555608bf14 in TradePool::Create (this=0x555560ae15dc, 
    sourceCity=..., destCity=..., owner=6, sType=ROUTE_TYPE_RESOURCE, 
    sResource=29, paying_for=6, gold_in_return=0) at TradePool.cpp:51
#7  0x000055555605d193 in Player::CreateTradeRoute (this=0x5555695400ec, 
    sourceCity=..., sourceType=ROUTE_TYPE_RESOURCE, sourceResource=29, 
    destCity=..., paying_for=6, gold_in_return=0) at Player.cpp:3071
#8  0x000055555607b84c in SendGoodEvent::GEVHookCallback (
    this=0x55555674d500 <s_SendGoodEvent>, gameEventType=GEV_SendGood, 
    args=0x555574ec1dcc) at PlayerEvent.cpp:558
#9  0x0000555555fc05f4 in GameEventHook::Activate (this=0x55555efc72bc, 
    args=0x555574ec1dcc, startIndex=0, resumeIndex=@0x5555747ec2cc: 1)
    at GameEventHook.cpp:254
#10 0x0000555555fbdab8 in GameEventManager::ActivateHook (this=0x55555efa253c, 
    type=GEV_SendGood, args=0x555574ec1dcc, startIndex=-1, 
    resumeIndex=@0x5555747ec2cc: 1) at GameEventManager.cpp:511
#11 0x0000555555fc001b in GameEvent::Process (this=0x5555747ec2bc)
    at GameEvent.cpp:112
#12 0x0000555555fbd7a8 in GameEventManager::ProcessHead (this=0x55555efa253c)
    at GameEventManager.cpp:440
#13 0x0000555555fbd6d7 in GameEventManager::Process (this=0x55555efa253c)
    at GameEventManager.cpp:415
#14 0x0000555555fbd613 in GameEventManager::ArglistAddEvent (
    this=0x55555efa253c, insert=GEV_INSERT_Tail, type=GEV_EndTurn, 
    argList=0x555574f5a41c) at GameEventManager.cpp:395
#15 0x0000555555fbd451 in GameEventManager::AddEvent (this=0x55555efa253c, 
    insert=GEV_INSERT_Tail, type=GEV_EndTurn, argTypes=0x7fffffffe940, 
    args=0x7fffffffe938) at GameEventManager.cpp:357
#16 0x0000555555fbc941 in GameEventManager::AddEvent (this=0x55555efa253c, 
    insert=GEV_INSERT_Tail, type=GEV_EndTurn, t1=GEA_Player, a1=5, t2=GEA_End)
    at GameEventManager.cpp:152
#17 0x0000555555f6d894 in dh_endTurn (itemAction=0x5555743ccdec, 
    seq=0x5555743d66ec, executeType=DHEXECUTE_NORMAL) at director.cpp:3183
#18 0x0000555555f65e39 in Director::HandleNextAction (this=0x55555bad059c)
    at director.cpp:733
#19 0x0000555555f406b8 in UnitActor::Process (this=0x555560355b4c)
    at UnitActor.cpp:844
#20 0x0000555555f6700b in Director::ProcessActiveUnits (this=0x55555bad059c)
    at director.cpp:1198
#21 0x0000555555f63bb8 in Director::Process (this=0x55555bad059c)
    at director.cpp:292
#22 0x0000555555780ae1 in CivApp::ProcessUI (this=0x555556cae5ac, 
    target_milliseconds=30, used_milliseconds=@0x7fffffffeb0c: 0)
    at ctp/civapp.cpp:2403
#23 0x0000555555780f4a in CivApp::Process (this=0x555556cae5ac)
    at ctp/civapp.cpp:2550
#24 0x0000555555776d68 in CivMain (iCmdShow=2, pSzCmdLine=0x7fffffffecb8)
    at ctp/civ3_main.cpp:1694
#25 0x0000555555776905 in main (argc=2, argv=0x7fffffffecb8)
    at ctp/civ3_main.cpp:1440
```
Do you have any idea how those changes could lead to a trade route whose `TradeRouteData` points to zero?

MartinGuehmann:
You have modified the code a lot, I would say. So you can expect to have overlooked something. First thing you stack trace says is that your source file is more recent than you executable. I would fix that first and hope that this was it. :D

However, I expect that it won't be that easy. I guess there is some collection of TradeRoute objects, where the TradeRoute was not removed from. Basically a TradeRoute is an ID. And if you remove the TradeRouteData from the pool, such an ID could still be around.

I know that is quite vague, for more I would have dig through the code.

LynxAbraxas:
> You have modified the code a lot, I would say. So you can expect to have overlooked something.

Yes, I fear so... The source file is more recent than the executable due to switching branches with git, so that is expected.
Odd is, that using `SeenBy` works for all the other changes made, just not for the radar map when it is another civ's turn.

LynxAbraxas:
Just had an insight: It seems to happen during `Player::CreateTradeRoute` which calls `TradeRouteData::GeneratePath` which in turn calls `RadarMap::RedrawTile` which wants to access the `TradeRouteData` but I think that is not yet assigned to the `TradeRoute` object at that time. I'll investigate this.

LynxAbraxas:
> Just had an insight: It seems to happen during `Player::CreateTradeRoute` which calls `TradeRouteData::GeneratePath` which in turn calls `RadarMap::RedrawTile` which wants to access the `TradeRouteData` but I think that is not yet assigned to the `TradeRoute` object at that time. I'll investigate this.

Yes, this was the problem, fixed with bf62f927a25df7fd2a658e27321e37e33a7474d6.

LynxAbraxas:
This PR is basically ready now, I will continue testing it for a while before I remove the WIP.

efef27405e95c2f0f8720e902958292063dbef60 was needed to prevent a segfault, but as far as I understand, that only was due to the save-game being in some intermediate state from testing the various commits of this PR. So basically efef27405e95c2f0f8720e902958292063dbef60 could be reverted, on the other hand it does not seem to hurt either. @MartinGuehmann what would you say? Revert or leave as is?

LynxAbraxas:
With 53a400c7d113966199c25cdd2a7dda255392e578 the trade route is now also revealed for tiles that have been visible before moving off to another one. That way routes that were created by other civs over the last turn appear, however so far only after the unit moved off:
No route so far within the vision of the Assurian's Hoplite:
ss_2020-02-29_17:36:28
Created a Persian route that has a tile within the vision of the Assurian's Hoplite:
ss_2020-02-29_17:37:16
Back to the vision of the Assurian's Hoplite, the new route is not visible:
ss_2020-02-29_17:37:58
but when moving the Hoplit/vision off, the route becomes visible:
ss_2020-02-29_17:38:23
The same if the route gets deleted:
ss_2020-02-29_17:41:16
ss_2020-02-29_17:43:12

It would be nicer if a newly created route is set to be seen for all civs whose vision contains at least one tile of the the newly created route. However, I have no idea so far how to realize that, so 53a400c7d113966199c25cdd2a7dda255392e578 is the best I can think of at the moment.

LynxAbraxas:
> It would be nicer if a newly created route is set to be seen for all civs whose vision contains at least one tile of the the newly created route. 

Realized with 5673e16712ae5ac4db6ef96065310e658b18b549.

LynxAbraxas:
Tested this for many turns and is working very well now. Trade routes are only drawn if they got discovered and their drawing is only removed when it was discovered that they do not exist any more.
Unknown trade route not within vision:
ss_2020-02-29_23:57:58
route revealed when in vision:
ss_2020-02-29_23:58:45
That way there is still something new and interesting to discover even if the whole map is explored. This makes it more difficult for a human to take AI overwhelming advantage of pirating because routes cancelled (due to too much pirating) and reestablished else where need to be found first. In addition, cities of a route are revealed such that the AI also knows about them like a human would otherwise know from the endings of trade routes.
Yanbu not known before new trade route is established:
ss_2020-03-02_23:23:44
but it is afterwards:
ss_2020-03-02_23:19:50
The chosen implementation does not break compatibility to existing save-games and overall only increases the size of a save-game by the amount of inactive (cancelled but not yet removed) trade routes.

LynxAbraxas:

LynxAbraxas:
Not yet ready for merge... While 14813282878794386c7d8c75efef2b3fde5cbff6 works as expected:
ss_2020-03-07_10:06:45
I am not really happy with calling
https://github.com/civctp2/civctp2/blob/14813282878794386c7d8c75efef2b3fde5cbff6/ctp2_code/gs/gameobj/Vision.cpp#L693
because it does far more than needed and it was not used before and needed a correction
https://github.com/civctp2/civctp2/blob/14813282878794386c7d8c75efef2b3fde5cbff6/ctp2_code/gs/gameobj/Vision.cpp#L124-L128
However, with the code before 14813282878794386c7d8c75efef2b3fde5cbff6 or when I replace `AddExplored(point, 0);` with `m_array[point.x][point.y] |= k_EXPLORED_BIT;`, the city name and icons are not drawn:
ss_2020-03-07_11:01:09
and the info on the city tile is "unexplored".
Currently, it is not clear to me what code I am missing from `AddExplored(point, 0);` that would do the trick without the extra fuzz. Any idea what it could be?

LynxAbraxas:
Seems it just needs a coord convert for `m_array` indexing and an unconvert for the unseen tile:
ss_2020-03-07_21:01:51

MartinGuehmann:
I guess you are now done with this. Ideally, I would like to merge #263 first. So that we don't have so many inactive trade routes around. However, it could be just that then we just get much more active routes so that it doesn't really matter.

LynxAbraxas:
Yes, merging #263 first might be better (or not matter). 
I am struggeling with https://github.com/civctp2/civctp2/pull/256/commits/f85b9598a16a4ac19b427a3f7c08d67886059804 which causes domestic cities with trade routes to be redrawn constantly such that e.g. units on that tile do not get visible, so merging this should wait until that issue is resolved or f85b9598a16a4ac19b427a3f7c08d67886059804 given up or reverted. However it would make hunting for trade routes even more interesting.

LynxAbraxas:
e1d7e0431c6dcf40558390166a56526e5c003e22 works, so it seems `IsVisible` or equivalently `(m_array[point.x][point.y] & k_VISIBLE_REFERENCE_MASK) > 0` is false for an owned tile which I find very odd but is also discriminated here:
https://github.com/civctp2/civctp2/blob/45cd01bb9f54502c2dfa13e8acd40558fb2af7c0/ctp2_code/gs/gameobj/Vision.cpp#L254-L258

LynxAbraxas:
With #280, trade route cost will not be infinite (10000). This was an indirect indicator that no route path could be found, which is the case if the there is no path on explored tiles to a target. This could happen if a city is known but surrounded by unexplored tiles, which (as far as I can imagine) could only happen so far since b7a1e76c633170446c2d6add0e641ef4cf416f43, however was already considered in the code before: If astar was not able to find a way the trade route would be marked invalid here
https://github.com/civctp2/civctp2/blob/b2aaae74c063d8e86b281ffb826d252fe30a8c55/ctp2_code/gs/gameobj/TradeRouteData.cpp#L109
However, testing if a trade route is valid only happens sparsely in the code, which was not a problem before, because such a situation could only happen before b7a1e76c633170446c2d6add0e641ef4cf416f43 if a unit was magically placed in unexplored terrain in the cheat mode.
So to avoid invalid routes during normal game play, I decided to not only reveal the source and destination cities of a trade route but all tiles of its path: 86319cdd86f11c2ca03167196909daa9f2ea8d7a. This also seems quite logical since if the traders of a route tell about their source and destination cities, they also know the tiles of the path of their trade route.

LynxAbraxas:
Just as a note: `Vision::IsExplored` needs to be passed a point
https://github.com/civctp2/civctp2/blob/b2aaae74c063d8e86b281ffb826d252fe30a8c55/ctp2_code/gs/gameobj/Vision.cpp#L164-L173
which needs to be run through `Unconvert` before, as in 8685f1c5431e2e5d85139b8ea23542b50b3e4f35. Otherwise all compiles and even runs but the result of `Vision::IsExplored` is inconsistent!

LynxAbraxas:
Tested this for many turns now and all works as expected. With https://github.com/civctp2/civctp2/pull/256/commits/4d3783e6083cab1e0b174f8a797b752cc0824232 all conflicts due to merges to master since the start of this PR are resolved.

MartinGuehmann:
Well, then let's merge.

LynxAbraxas:
Just as a demo: 
Before discovery of a new trade route:
ss_2020-03-30_01:00:32
whose traders (when discovered) tell about the source city (Edo) size, the transition through a new city (Tell Rimah) and the tiles along the route (here loss of a fishery beside Edo):
ss_2020-03-30_01:01:11
Visiting that trade route again some turns later (Diplomat close to Edo) reveals that the transit city Tell Rimah was captured by the Yamamoto:
ss_2020-03-30_02:10:41

LynxAbraxas:
Receiving a good from a formerly unknown city now reveals that city:
ss_2020-04-04_01:19:52
ss_2020-04-04_01:18:18

LynxAbraxas:
Creating a trade route to a foreign city (Rome) reveals (and updates) the current status:
ss_2020-04-12_23:00:09
ss_2020-04-12_23:00:35

LynxAbraxas:
Regular investigation of known trade routes and checking for new ones can reveal a lot of remote details:
ss_2020-04-20_01:00:04
ss_2020-04-20_01:00:37
ss_2020-04-21_23:04:29
ss_2020-04-21_23:05:13

williamd67:
Nice, so talking to traders will give a lot of information. Sounds like reality.  Now you have to add secretive traders that will not tell you anything or give false information to keep their monopoly ;)

Issue #231, opened 2020-01-03
Pollution Pact agreed target value and count-down not listed any where 


LynxAbraxas:
While it is possible to agree to a Pollution pact, the value for the targeted reduction is not listed any where, which makes it very difficult to check that one meets the pact's expectations:
![ss_2019-12-31_21:21:32](https://user-images.githubusercontent.com/21012234/71751676-993dad80-2e7c-11ea-957e-ca32a06479be.png)
In addition it would be helpful to have a count-down for when the target value has to be reached, as well as for the other side.

LynxAbraxas:
So far this info has to be remembered by the player or saved via screen shots:
![ss_2019-12-15_00:17:06](https://user-images.githubusercontent.com/21012234/71752858-fe939d80-2e80-11ea-960a-5d78c96e7a0a.png)
![ss_2019-12-15_00:18:46](https://user-images.githubusercontent.com/21012234/71752861-018e8e00-2e81-11ea-9838-0c4faab4635b.png)
![ss_2019-12-15_00:19:16](https://user-images.githubusercontent.com/21012234/71752786-afe60380-2e80-11ea-9243-14152274058d.png)
![ss_2019-12-15_00:19:48](https://user-images.githubusercontent.com/21012234/71752795-b7a5a800-2e80-11ea-925e-60e110bc9611.png)

LynxAbraxas:
Similar problems when agreeing on nano-tech weapon reduction:
![ss_2019-12-16_22:39:43](https://user-images.githubusercontent.com/21012234/71753685-9a260d80-2e83-11ea-8936-2c524f04c2d3.png)

LynxAbraxas:
While ivestigating solutions for #245, I found some code that seems to implement a listing of agreements made and received:
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_data/default/uidata/layouts/diplomacywindow.ldl#L438
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_data/default/uidata/layouts/diplomacywindow.ldl#L453
https://github.com/civctp2/civctp2/blob/bf9a81223e9c1ec266364807555fe9ffc7b75b60/ctp2_data/default/uidata/layouts/diplomacywindow.ldl#L483

However, `DiplomacyWindow::UpdateProposalList`, which would populate and update these lists
[...]

PR #258, opened 2020-02-01
WIP: Introducing listing of diplomatic interactions

LynxAbraxas:
This PR tries to solve #231 by introducing a listing of diplomatic interactions in a new tab in the diplomacy manger.

### DIFF
--- a/ctp2_code/ui/interface/diplomacywindow.cpp
+++ b/ctp2_code/ui/interface/diplomacywindow.cpp
@@ -309,6 +309,7 @@ void DiplomacyWindow::ShowSections(uint32 sections)
 void DiplomacyWindow::Update()
 {
 	UpdateCreateList();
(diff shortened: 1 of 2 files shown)


LynxAbraxas:
With ceb15be4626ab061ab14191518161b08d8c1927b I get a listing:
ss_2020-02-01_19:34:06
but it seems the code behind the proposal management is not used (probably from CPT1) so calls to
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/diplomacy/diplomat.cpp#L3372-L3389
return `Diplomat::s_badNewProposal`, i.e. are useless.
So to get this PR into a reasonable state those problems need to be addressed.

Feel free to base new PRs on this one or reopening this with new contributions.

Issue #117, opened 2019-04-14
piracy bounty is not assigned in case multiple civs pirate the same route in the same turn


LynxAbraxas:
The discussion in #75 and https://github.com/civctp2/civctp2/pull/76#issuecomment-482749127 showed that the bounty of piracy is not assigned  in the case that multiple civs pirate the same route in the same turn.

One way to determine a disticinct winner of the concurrent pirating civs would be to assign the bounty to the civ whose unit is closest to the trade route source along the trade route.

Currently, no civ gets the bounty if  multiple civs pirate the same route in the same turn, i.e. the message of #71 does not appear  (as might be expected) nor does the victim get informed  (as might be expected as well, see #75) nor can the victim ask to stop piracy by diplomacy (as might be expected too, see #116).

LynxAbraxas:
Relating to https://github.com/civctp2/civctp2/issues/75#issuecomment-483064108

> Yes, in fact you only have to count the nodes of the path, I think I have done something like this with the empire center calculation.

That's a nice idea simple to establish, thanks for that @MartinGuehmann . Can you point me to where you did this in the code?
(BTW, you can easily get a perma-link to some code on GH, see e.g. https://help.github.com/en/articles/creating-a-permanent-link-to-a-code-snippet just make sure it uses the commit SHA and not a branch name, in that case hitting 'y' convertes the link into the current commit SHA)

MartinGuehmann:
> Relating to [#75 (comment)](https://github.com/civctp2/civctp2/issues/75#issuecomment-483064108)
> 
> > Yes, in fact you only have to count the nodes of the path, I think I have done something like this with the empire center calculation.
> 
> That's a nice idea simple to establish, thanks for that @MartinGuehmann . Can you point me to where you did this in the code?
> (BTW, you can easily get a perma-link to some code on GH, see e.g. https://help.github.com/en/articles/creating-a-permanent-link-to-a-code-snippet just make sure it uses the commit SHA and not a branch name, in that case hitting 'y' convertes the link into the current commit SHA)

Well let's try this:

https://github.com/civctp2/civctp2/blob/b3ac27230a0cfba4cba2c912c81ab5abf18fa3ae/ctp2_code/gs/gameobj/Player.cpp#L9730-L9748

Basically, here I did it, but it is actually the other way around, I know the the numbers of tiles to move and I want to know the position. But well, you just have to copy, paste, and modify RestoreIndexAndCurrentPos:

[...]

PR #259, opened 2020-02-01
WIP: Corrections to goals

LynxAbraxas:
This PR is meant for some corrections/improvements to the goal code.

### DIFF
--- a/ctp2_code/ai/strategy/goals/Goal.cpp
+++ b/ctp2_code/ai/strategy/goals/Goal.cpp
@@ -2885,8 +2885,21 @@ bool Goal::IsTargetImmune() const
 
 	if(order_record->GetUnitPretest_EstablishEmbassy())
 	{
+	        // do not go for installing an embassy if one already is installed for the civ of that city
 		if(g_player[m_playerId]->HasEmbassyWith(m_target_city->GetOwner()))
 			return true;
+
+		// do not go for installing an embassy if civ of target city has wonder that forbids so
+		if(wonderutil_GetCloseEmbassies(g_player[m_target_city->GetOwner()]->m_builtWonders)) { // code from ArmyData::EstablishEmbassy
+		    for(sint32 w = 0; w < g_theWonderDB->NumRecords(); w++) {
+			if((g_player[m_target_city->GetOwner()]->m_builtWonders & ((uint64)1 << w)) &&
+			    !wonderutil_IsObsolete(w)) {
+			    return true;
+			    }
+			}
+		    }
+		    
+
 	}
 
 	if(order_record->GetUnitPretest_CanCreateFranchise())
@@ -2933,7 +2946,7 @@ bool Goal::IsTargetImmune() const
 			return true;
 	}
 
-	if(!wonderutil_GetProhibitSlavers(g_theWonderTracker->GetBuiltWonders()))
+	if(!wonderutil_GetProhibitSlavers(g_theWonderTracker->GetBuiltWonders())) // wonder affects all civs
 	{
 		// Abolitionist has to go to cities with slaves
 		if(    order_record->GetUnitPretest_CanInciteUprising()
--- a/ctp2_code/ai/strategy/goals/Goal.cpp
+++ b/ctp2_code/ai/strategy/goals/Goal.cpp
@@ -2949,9 +2949,11 @@ bool Goal::IsTargetImmune() const
 			sint32 popCount;
 			m_target_city.GetCityData()->GetPop(popCount);
 
-			// Slavers must go to cities with population to enslave, give an extra point so that the city isn't killed on conquest
-			if(popCount <= 2)
+			// Slavers must go to cities with population to enslave, i.e. OK for pop >= 2
+			if(popCount <= 1)
 				return true;
+			// city wall check should be added if fuzzy-logic is implemented
+			// city walls reduce chance of success but do not prevent it completely
 		}
 
 		if(      order_record->GetUnitPretest_CanEnslaveSettler()
--- a/ctp2_code/ai/strategy/goals/Goal.cpp
+++ b/ctp2_code/ai/strategy/goals/Goal.cpp
@@ -2891,12 +2891,7 @@ bool Goal::IsTargetImmune() const
 
 		// do not go for installing an embassy if civ of target city has wonder that forbids so
 		if(wonderutil_GetCloseEmbassies(g_player[m_target_city->GetOwner()]->m_builtWonders)) { // code from ArmyData::EstablishEmbassy
-		    for(sint32 w = 0; w < g_theWonderDB->NumRecords(); w++) {
-			if((g_player[m_target_city->GetOwner()]->m_builtWonders & ((uint64)1 << w)) &&
-			    !wonderutil_IsObsolete(w)) {
-			    return true;
-			    }
-			}
+		    return true;
 		    }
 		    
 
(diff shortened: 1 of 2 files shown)


LynxAbraxas:
I have the impression that the additions of 00752a6233c710add9d98111d569ba224367d9ae do check if the player has a wonder that forbids to install embassies but the code that then checks if the wonder is obsolete only checks if that player has a wonder that is obsolete not if that particular wonder is obsolete. The code is from
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/gs/gameobj/ArmyData.cpp#L3566-L3577
which seems to suffer the same problem. It seems there is no code that would return the current wonder that forbids the creation of embassies (probably because that could be multiple wonders).

LynxAbraxas:
Concerning #256, I wonder what the AI uses for determining if a unit/army should go for pirating just the seen cell info or the global map info (as available to the human player int this particular case). Appropriate and fair for general orders would be the unseen cell info but that does not exist so far for trade routes.
`Goal::IsTargetImmune` only checks of the target city has trade routes
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/strategy/goals/Goal.cpp#L2920-L2927
I wonder though where the target city comes from in that case where basically every position on the route is equally suitable (without #117) for pirating.

LynxAbraxas:
@MartinGuehmann Do you have any answers concerning the cell info used for pirating goals?

LynxAbraxas:
@MartinGuehmann Any idea where and how one could place a check for the intercept-trade-goal (pirate-goal) such that the AI does not pirate routes it is the receiver of (i.e. gets the bonus, see https://github.com/civctp2/civctp2/issues/192#issuecomment-570716436)?
Should that go here
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/strategy/goals/Goal.cpp#L3087-L3105
or here
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/strategy/goals/Goal.cpp#L1050-L1091

MartinGuehmann:
> @MartinGuehmann Any idea where and how one could place a check for the intercept-trade-goal (pirate-goal) such that the AI does not pirate routes it is the receiver of (i.e. gets the bonus, see [#192 (comment)](https://github.com/civctp2/civctp2/issues/192#issuecomment-570716436))?
> Should that go here
> https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/strategy/goals/Goal.cpp#L3087-L3105
> 
> 
> or here
> https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/strategy/goals/Goal.cpp#L1050-L1091

That is a tricky question, for sure it is not the second one, because that just selects the target position, but does not select whether the AI should go for that trade route. The problem is that you have here two target owners. And thus both must be checked. I tend to the first one, but I cannot exclude that there are better places.

LynxAbraxas:
@MartinGuehmann  any idea why https://github.com/civctp2/civctp2/commit/00752a6233c710add9d98111d569ba224367d9ae might not lead to the disired effect that diplomats of other civs do not go for installing an embassy in one of my cities? I have the "forbidden city" wonder and no wonder that is obsolete (and the action of installing an embassy also always fails for the other civs, but the AI keeps trying).

MartinGuehmann:
> @MartinGuehmann any idea why [00752a6](https://github.com/civctp2/civctp2/commit/00752a6233c710add9d98111d569ba224367d9ae) might not lead to the disired effect that diplomats of other civs do not go for installing an embassy in one of my cities? I have the "forbidden city" wonder and no wonder that is obsolete (and the action of installing an embassy also always fails for the other civs, but the AI keeps trying).

Well, the code seems to the AI from trying to establish embassies, if you have the forbidden wonder and no wonder that is obsolete. Obviously, that isn't what you want, but for testing it should be fine.

You could now return true if you just have the forbidden wonder, at least you could check whether the problem is here.

By the way, you should use Player::HasWonder that makes the code easier to read.

LynxAbraxas:
> Well, the code seems to the AI from trying to establish embassies, if you have the forbidden wonder and no wonder that is obsolete. Obviously, that isn't what you want, but for testing it should be fine.

Sorry, I don't get here what you mean. 
The intention of the code of https://github.com/civctp2/civctp2/commit/00752a6233c710add9d98111d569ba224367d9ae is that the AI does not try to install embassies in cities of a civ that has a wonder that forbids to do so (like the "forbidden city" wonder).

MartinGuehmann:
> > Well, the code seems to the AI from trying to establish embassies, if you have the forbidden wonder and no wonder that is obsolete. Obviously, that isn't what you want, but for testing it should be fine.
> 
> Sorry, I don't get here what you mean.
> The intention of the code of [00752a6](https://github.com/civctp2/civctp2/commit/00752a6233c710add9d98111d569ba224367d9ae) is that the AI does not try to install embassies in cities of a civ that has a wonder that forbids to do so (like the "forbidden city" wonder).

It just doesn't do that: It checks whether the target owner has a wonder that closes embassies. If yes, it checks whether the target owner has a wonder that is not obsolete. If yes, the target is immune.

Actually, I have just checked wonderutil_GetCloseEmbassies, it already checks whether the wonder is obsolete. Maybe it is worth checking where else wonderutil_IsObsolete is used without reason.

LynxAbraxas:
There is definitely something wrong with the goals for Diplomats:
ss_2020-03-06_14:55:34
I suspect they try to install embassies even when at war.

LynxAbraxas:
Concerning https://github.com/civctp2/civctp2/pull/259#issuecomment-581060775
> Concerning #256, I wonder what the AI uses for determining if a unit/army should go for pirating just the seen cell info or the global map info (as available to the human player int this particular case). Appropriate and fair for general orders would be the unseen cell info but that does not exist so far for trade routes.

To determine the pirating position of a goal, it seems the AI uses the `GetTradeSourceList` of the target city
https://github.com/civctp2/civctp2/blob/71571fc29255a13215a289e00668ad055e776f46/ctp2_code/ai/strategy/goals/Goal.cpp#L1059
which consists of the actual active routes, i.e. the AI does not use the (un)seen cell info (like a human must). So actually the AI cheats (has an unfair advantage here), but it is so bad at pirating any way or does so very rarely, that it might be quite appropriate to leave this as is for now. 
It would be easy to skip each route in case `!route.SeenBy(m_owner)` but that would not add the evaluation of seen routes that are now inactive (as a human would with #256)
It would also be possible to improve the evaluation of the best pirating position here by introducing a check if there are tiles along the routes of the target city that contain multiple (known) trade routes.

However, a proper redesign of the evaluation would be to check all routes of the map, skip those that do not belong to the owner of the target city, check if there are tiles with multiple routes (possibly of other civs), trade off amount of routes of a tile versus shortest distance. 
Even nicer would be if the AI could be taught to move along a route to a tile with multiple routes, perserving its last movement point for pirating in that turn, and continue to move in the next turn until the tile with multiple routes is reached. This would correspond to how a clever human would do this.

LynxAbraxas:
The correction concerning establishing embassies is solved in #330.
I won't find the time to improove the pirating goals in the near future, so closed to serve as a reference. Please feel free to pick it up.

PR #261, opened 2020-02-04
Adjustments for general GL-CI runners

LynxAbraxas:
Regression for #241, which apparently does not work as is for all of the available GL-CI runners. As devices are not needed for Xvfb (at least not as long as screen casting from Xvfb does not work) nor guaranteed to be available on every shared runner by default, they are now move to the docker parameters when used with X11.


MartinGuehmann:
Well, I assume I can merge it.

LynxAbraxas:
Yes, works for me, now also on a general GL-CI runner, so hopefully will also work for https://gitlab.com/civctp2/civctp2.

MartinGuehmann:
Then let's merge.

Issue #237, opened 2020-01-05
A* path algorithm needs corrections


LynxAbraxas:
Currently, the A* path algorithm does not seem to take into account to hold sufficient distance when passing enemy cities (civ at war). This can lead to the loss of units e.g. helicopter or plane due to active air defence or boats if city has artillery.

In addition the current A* algorithm seems to be wrong at finding optimal paths for (modern) units, see e.g. these paths for adjacent tiles:

![ss_2019-11-24_21:34:50](https://user-images.githubusercontent.com/21012234/71784111-f27a1e00-2fef-11ea-9683-ffb5319f2eb4.png)
![ss_2019-11-24_21:34:37](https://user-images.githubusercontent.com/21012234/71784112-f27a1e00-2fef-11ea-9260-888426a1841e.png)

![ss_2019-11-24_21:11:24](https://user-images.githubusercontent.com/21012234/71784116-058cee00-2ff0-11ea-86e4-f28eade1eb7b.png)
![ss_2019-11-24_21:11:44](https://user-images.githubusercontent.com/21012234/71784117-0756b180-2ff0-11ea-81f1-5ebc8dd91252.png)

![ss_2019-11-18_22:37:00](https://user-images.githubusercontent.com/21012234/71784120-0e7dbf80-2ff0-11ea-81a3-d431947606f7.png)
![ss_2019-11-18_22:37:06](https://user-images.githubusercontent.com/21012234/71784123-10478300-2ff0-11ea-8f52-7abe6b9d7f6f.png)

LynxAbraxas:
Problem also exist for non-modern units:
![ss_2020-01-14_22:59:22](https://user-images.githubusercontent.com/21012234/72462776-2dd7d200-37d2-11ea-84db-45b05b7557c4.png)
![ss_2020-01-14_22:59:28](https://user-images.githubusercontent.com/21012234/72462778-2f08ff00-37d2-11ea-95d5-4e0f30320d13.png)

LynxAbraxas:
> Regarding 'active defense' I assume that the A* algorithm should take visible active defense units into account assuming that enemy cities contains active defense units.  Of course the owner of those units should be at war with the active player.

Yes.

> This has only effect when bypassing such units and not as it is an end position.

A slightly different but similar situation is when there is no active defense but a unit that could attack, e.g. if the A* leads a ship just aside a city/artillery and end the turn in reach of bombardment.

[...]

Issue #224, opened 2019-12-01
AI unintentionally declares war (runs over units)


LynxAbraxas:
Even with #101, I still observe that the AI attacks civilian units when not at war with the unit owner. I have the impression that the AI has ordered such a unit/army with something different than actually attacking, but when it encounters a civilian unit while going to the point of order, it attacks that unit unconditionally (similar to #91). 

Here e.g. a Sumarian tank's order is "Group to (2, 112), Seige (9, 112)" but on its way there it comes across a Persian lawyer and attacks it (eventhough it had a cease-fire agreement with the Persians):

On its way, north of Isin:
![ss_2019-11-28_23:06:13](https://user-images.githubusercontent.com/21012234/69917107-d7e9dd80-1462-11ea-8987-c715df15324f.png)

attacking Lawyer (north west of Kish):
![ss_2019-11-28_23:06:26](https://user-images.githubusercontent.com/21012234/69917103-d3bdc000-1462-11ea-9d13-439b67207925.png)

Order still unchanged ("Group to (2, 112), Seige (9, 112)") after killing Lawyer:
![ss_2019-11-28_23:06:45](https://user-images.githubusercontent.com/21012234/69917102-d3bdc000-1462-11ea-9a33-934230256015.png)

Similarly, I also observed that a unit/army of a human player given an order (that includes a go-to) attacks another civ's unit/army if encountered along the way. However, if that would lead to war, the human player is shown the message that asks if one want's to declare war. Since that message/question does not exist for the AI, this could possibly be the reason for the unintentional declaration of war by running-over.

LynxAbraxas:
Getting this assert with the debug version for the attack of the lawyer:
```
Assertion (!g_player[owner]->IsRobot() || Diplomat::GetDiplomat(owner).HasWarOrDesiresPreemptivelyWith(defender->GetOwner())) Failed in File:armyevent.cpp, Line:810
```

LynxAbraxas:
Getting this assert with the debug version for the attack of the lawyer:
```
Assertion (!g_player[owner]->IsRobot() || Diplomat::GetDiplomat(owner).HasWarOrDesiresPreemptivelyWith(defender->GetOwner())) Failed in File:armyevent.cpp, Line:810
```
for  [dbgSue.zip](https://github.com/civctp2/civctp2/files/3936133/dbgSue.zip).

LynxAbraxas:
My current impression is that this originates from
[...]

PR #263, opened 2020-02-08
Improve AI trade route management

LynxAbraxas:
PR to improve AI trade route management.


LynxAbraxas:
@MartinGuehmann  Any idea how to ensure that the trade route cancelling
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/CityManagement/governor.cpp#L5167-L5173
is processed before trade route creation
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/CityManagement/governor.cpp#L5206-L5211
?
I have hit the case where a trade route gets created first, hitting
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/gs/gameobj/Player.cpp#L3027
which kills the current route of that good, and then creates the new one but then the kill even from 
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/CityManagement/governor.cpp#L5167-L5173
comes afterwards, killing the newly created route again.

LynxAbraxas:
> I have hit the case where a trade route gets created first, hitting
> 
> https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/gs/gameobj/Player.cpp#L3027
> 
> which kills the current route of that good, and then creates the new one but then the kill event from
> 
> https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/CityManagement/governor.cpp#L5167-L5173
> 
> comes afterwards, killing the newly created route again.

Figured now that the event processing was working all right, but a unit in the way of a move-order lead to the problem of #224, in particular https://github.com/civctp2/civctp2/issues/224#issuecomment-583874528, which caused war with the civ that the newly created trade routes were initiated with...
Since the problem of running into some other civ cannot be foreseen when the governor creates new routes, it does not make much sense to bind the creation of new routes to 
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/diplomacy/diplomat.cpp#L5160-L5170
instead solving #224 should solve also this issue here.

LynxAbraxas:
The assertion of 2f429d8aa76548e9b214b84c1476e7adf35cde5a was issued in this case

LynxAbraxas:
@MartinGuehmann is the way of saving sint32 instead of double in fbd9526435fcb9ab4813b4c6bba918bd0dfc42f9 OK? Or would it be better to convert the sint32 to double for saving and back when loading? That would preserve save-games and would not need a converter (7c3eecc52e1eb095c0f4dd54af74bcdf4fc5522d).

MartinGuehmann:
> @MartinGuehmann is the way of saving sint32 instead of double in [fbd9526](https://github.com/civctp2/civctp2/commit/fbd9526435fcb9ab4813b4c6bba918bd0dfc42f9) OK? Or would it be better to convert the sint32 to double for saving and back when loading? That would preserve save-games and would not need a converter ([7c3eecc](https://github.com/civctp2/civctp2/commit/7c3eecc52e1eb095c0f4dd54af74bcdf4fc5522d)).

No you can't do it like this: First, you don't load and save the variable twice. You do expect to have different things in the two parts of the double, anyway. On Wikipedia, we have some details about it: https://en.wikipedia.org/wiki/Double-precision_floating-point_format

So that is something that cannot be dealt easily with.

So you would cast it to a double and save it, and then cast it from a double to a sint32 back on loading. It is just the question whether this defeats what you are trying to do here.

LynxAbraxas:
> No you can't do it like this: First, you don't load and save the variable twice. You do expect to have different things in the two parts of the double, anyway.

It does work (with the convert) but it is not compatible with the former save-game format. So does 7aed8a758d32a0f20d0e63fc2288ef92226f49e4 fit your expectations?

> It is just the question whether this defeats what you are trying to do here.

The point of fbd9526435fcb9ab4813b4c6bba918bd0dfc42f9 is not saving space in the save-game but to consistently regard the cost of a trade route as an int instead of casting back an for as was the case before c4e573640751fa67a88dd0731723e25189328151.

LynxAbraxas:
@MartinGuehmann  Do you have any idea what `!city.CD()->HasResource(g)`
is meant for in this if-statement here:
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/ai/CityManagement/governor.cpp#L5088-L5097
i.e. do you think the changes in fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8 make sense?

MartinGuehmann:
> > No you can't do it like this: First, you don't load and save the variable twice. You do expect to have different things in the two parts of the double, anyway.
> 
> It does work (with the convert) but it is not compatible with the former save-game format. So does [7aed8a7](https://github.com/civctp2/civctp2/commit/7aed8a758d32a0f20d0e63fc2288ef92226f49e4) fit your expectations?

That looks good. Just as a note it is not just about the alignment, but also about the data representation that needs to be conserved. 

> > It is just the question whether this defeats what you are trying to do here.
> 
> The point of [fbd9526](https://github.com/civctp2/civctp2/commit/fbd9526435fcb9ab4813b4c6bba918bd0dfc42f9) is not saving space in the save-game but to consistently regard the cost of a trade route as an int instead of casting back an for as was the case before [c4e5736](https://github.com/civctp2/civctp2/commit/c4e573640751fa67a88dd0731723e25189328151).

Well, you are still casting once, that could still change the value slightly after a reload. And if it does then the question is still whether it is important.

MartinGuehmann:
> @MartinGuehmann Do you have any idea what `!city.CD()->HasResource(g)`
> is meant for in this if-statement here:
> https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/ai/CityManagement/governor.cpp#L5088-L5097
> 
> 
> i.e. do you think the changes in [fc09b2b](https://github.com/civctp2/civctp2/commit/fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8) make sense?

It is quite hard to understand, especially if you don't know what `!city.CD()->HasResource(g)` is good for in the first place.

LynxAbraxas:
> Well, you are still casting once, that could still change the value slightly after a reload. 

Since we know that we have had an int to save, I would expect that there is no precision/rounding error in casting to double and back. Storing the sint32 as a double is basically only wasting space but preserves alignment and data representation.

> It is quite hard to understand, especially if you don't know what !city.CD()->HasResource(g) is good for in the first place.

True, and I still am in doubt what it might have been meant for but cannot make any sense of it. I'll test this for a few more turns and if no unexpected behavior show up I would otherwise leave fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8 as.

MartinGuehmann:
> True, and I still am in doubt what it might have been meant for but cannot make any sense of it. I'll test this for a few more turns and if no unexpected behavior show up I would otherwise leave [fc09b2b](https://github.com/civctp2/civctp2/commit/fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8) as.

I am really unconformable if you don't really know what your change now does. It is fine if you don't know what the original code does as long as you know what your new code does and what it is supposed to fix.

LynxAbraxas:
In regard to https://github.com/civctp2/civctp2/pull/263#issuecomment-583875350

> Since the problem of running into some other civ cannot be foreseen when the governor creates new routes, it does not make much sense to bind the creation of new routes to

the discussion in https://github.com/civctp2/civctp2/pull/264#issuecomment-586577271 showed that it does make sense to bind the creation of new routes to what `DesireWarWith` evaluates because `DesireWarWith` determines the decision on declaring war in opportune situations. Since the result of the evaluation can change in the turns between the route creation and the occurrence of the opportune situation,  it is necessary that `DesireWarWith` also takes the value of standing trade routes into account.

LynxAbraxas:
> the discussion in https://github.com/civctp2/civctp2/pull/264#issuecomment-586577271 showed that it does make sense to bind the creation of new routes to what `DesireWarWith` evaluates because `DesireWarWith` determines the decision on declaring war in opportune situations. Since the result of the evaluation can change in the turns between the route creation and the occurrence of the opportune situation,  it is necessary that `DesireWarWith` also takes the value of standing trade routes into account.

Thinking about this further, binding the creation of new routes to what `DesireWarWith` evaluates is not a good idea, because `DesireWarWith` evaluates to true much more often war actually happens, so such binding would prevent the creation of valuable trade routes even though it might never come to war, i.e. if no opportune situation occurs. So I would say it would only make sense to take the value of standing trade routes into account  in `DesireWarWith` and as such would be its own PR or should go into #264.

LynxAbraxas:
> > True, and I still am in doubt what it might have been meant for but cannot make any sense of it. I'll test this for a few more turns and if no unexpected behavior show up I would otherwise leave [fc09b2b](https://github.com/civctp2/civctp2/commit/fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8) as.
> 
> I am really unconformable if you don't really know what your change now does. It is fine if you don't know what the original code does as long as you know what your new code does and what it is supposed to fix.

Checked this again, it comes from the original code
https://github.com/civctp2/civctp2/blob/d614fbdf705db334a4b45037bbcc735142d22016/ctp2_code/ai/CityManagement/governor.cpp#L3414-L3415
and I think it is meant to prevent the cancellation of standing routes if a good of type g is still available.
https://github.com/civctp2/civctp2/blob/45cd01bb9f54502c2dfa13e8acd40558fb2af7c0/ctp2_code/gs/gameobj/CityData.cpp#L6565-L6569
`city.CD()->HasResource(g)` is true if subtracting the number of sold goods g from the number of collected goods g is still positive, i.e. there are goods of type g left over.
So currently
https://github.com/civctp2/civctp2/blob/45cd01bb9f54502c2dfa13e8acd40558fb2af7c0/ctp2_code/ai/CityManagement/governor.cpp#L5088-L5097
standing trade routes are not considered for cancellation if the city still has a good of type g. While this makes sense, this also prevents the trade route from being considered for cancellation due to pirating (and to free the caravans for a better offers). However, since under common map conditions it is unlikely that a city has a good multiple times within its influence, most routes would be considered due to pirating when reaching `MaxPiracyEvents`. So this inappropriate condition was hardly ever met.

MartinGuehmann:
> standing trade routes are not considered for cancellation if the city still has a good of type g. While this makes sense, this also prevents the trade route from being considered for cancellation due to pirating (and to free the caravans for a better offers). However, since under common map conditions it is unlikely that a city has a good multiple times within its influence, most routes would be considered due to pirating when reaching `MaxPiracyEvents`. So this inappropriate condition was hardly ever met.

I would say, having a good twice is still realistic, but even I think the change in [fc09b2b](https://github.com/civctp2/civctp2/commit/fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8) is the right one. So thanks for clearing this up.

LynxAbraxas:
I have an idea for a better way to not cancel standing trader routes if the good is available multiple times but to still consider cancelling in case of too much pirating. I will investigate that and replace https://github.com/civctp2/civctp2/commit/fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8 accordingly.

LynxAbraxas:
> I have an idea for a better way to not cancel standing trader routes if the good is available multiple times but to still consider cancelling in case of too much pirating. I will investigate that and replace [fc09b2b](https://github.com/civctp2/civctp2/commit/fc09b2be763ac0fa6fd46ceddd92d8b8778f8ce8) accordingly.

@MartinGuehmann how about bb9fec8dca6318694069af750a84b416ee3b0814? Standing trade routes then only get considered for cancelling due to too much pirating but not in case the good is still available from another collected source.

LynxAbraxas:
Tested this for many turns now and all in all seems to improve the AI trade route management. A human can still do better but putting all those decisions into code is really hard. For example the AI still sometimes creates trade routes and cancels them the next turn because it got new caravans that allow to create a better route, a human would just wait for a turn and create that final route then, avoiding to loos a caravan just for the cancelling, however also missing the profit for one turn then.
It seems to me that the AI only starts with serious/excessive trading when it has reached the max. number of cities its gov allows and those have built the important things. Probably the building of caravans is lower in priority than the other things (which is appropriate). So many new trade routes appear only after 500AD, but there are a few that appear quite quickly after the advance trade is researched.
I'm not sure at what values casting an int to a double and back leads to changed values but probably not within the value range that can be expected for `m_transportCost`. Ideally the sint32 should be saved as is but that would break compatibility to existing save-games.

LynxAbraxas:

LynxAbraxas:
Not ready yet, just had an idea on how the processing could be sped up.

MartinGuehmann:
> Tested this for many turns now and all in all seems to improve the AI trade route management. A human can still do better but putting all those decisions into code is really hard. For example the AI still sometimes creates trade routes and cancels them the next turn because it got new caravans that allow to create a better route, a human would just wait for a turn and create that final route then, avoiding to loos a caravan just for the cancelling, however also missing the profit for one turn then.

For the new caravans you could check what the AI has in its build queues, I but I think then the more difficult part is to figure out for how many turns to postpone the creation of the trade route at some point at some point the costs for waiting should be higher than loosing that caravan, since there was also some trade to gain.

LynxAbraxas:
> For the new caravans you could check what the AI has in its build queues, I but I think then the more difficult part is to figure out for how many turns to postpone the creation of the trade route at some point at some point the costs for waiting should be higher than loosing that caravan, since there was also some trade to gain.

Yes, and I learned from all my attempts to improve trade stuff, that it is quite hard to not make other things worse or even break other things. So I'd leave this as it is.

LynxAbraxas:
> Not ready yet, just had an idea on how the processing could be sped up.

https://github.com/civctp2/civctp2/pull/263/commits/bf4e15e3756d76f19d9516fdb77ef9fec16bff3b does so. But the actual idea was to speed up `tradeutil_GetAccurateTradeDistance` which slows down the game execution quite a bit because it calls `g_theTradeAstar.FindPath`, most notably when the governor or the trade manager evaluates (around a 100 times or more) the possible new routes. However, it leads to the general astar, which I do not want to touch. Since there seem to be some bugs there (#237), a future PR should perhaps consider using an optimized implementation of astar, e.g. from boost.

LynxAbraxas:
So this PR is ready for merging.

MartinGuehmann:
> > For the new caravans you could check what the AI has in its build queues, I but I think then the more difficult part is to figure out for how many turns to postpone the creation of the trade route at some point at some point the costs for waiting should be higher than loosing that caravan, since there was also some trade to gain.
> 
> Yes, and I learned from all my attempts to improve trade stuff, that it is quite hard to not make other things worse or even break other things. So I'd leave this as it is.

OK.

> > Not ready yet, just had an idea on how the processing could be sped up.
> 
> [bf4e15e](https://github.com/civctp2/civctp2/commit/bf4e15e3756d76f19d9516fdb77ef9fec16bff3b) does so. But the actual idea was to speed up `tradeutil_GetAccurateTradeDistance` which slows down the game execution quite a bit because it calls `g_theTradeAstar.FindPath`, most notably when the governor or the trade manager evaluates (around a 100 times or more) the possible new routes. However, it leads to the general astar, which I do not want to touch. Since there seem to be some bugs there (#237), a future PR should perhaps consider using an optimized implementation of astar, e.g. from boost.

Even if you had fix the astar, it would still be slower than your solution, so I think this is still better.

> So this PR is ready for merging.

Well, then let's merge.

PR #264, opened 2020-02-09
WIP: Debug unintional war by AI

LynxAbraxas:
PR to solve #224.

### DIFF
--- a/ctp2_code/ai/diplomacy/diplomat.cpp
+++ b/ctp2_code/ai/diplomacy/diplomat.cpp
@@ -5154,7 +5154,10 @@ void Diplomat::SendGreeting(const PLAYER_INDEX & foreignerId)
 }
 bool Diplomat::HasWarOrDesiresPreemptivelyWith(const PLAYER_INDEX foreignerId) const
 {
-	return g_player[m_playerId]->HasWarWith(foreignerId) || (/*m_personality->GetAlignmentEvil() || m_personality->GetTrustworthinessChaotic() &&*/ DesireWarWith(foreignerId));
+	return g_player[m_playerId]->HasWarWith(foreignerId)
+	    || (m_personality->GetAlignmentEvil()
+	       || (m_personality->GetTrustworthinessChaotic() && DesireWarWith(foreignerId))
+	       );
 }
 
 bool Diplomat::DesireWarWith(const PLAYER_INDEX foreignerId) const
--- a/ctp2_code/ai/diplomacy/diplomat.cpp
+++ b/ctp2_code/ai/diplomacy/diplomat.cpp
@@ -5154,10 +5154,7 @@ void Diplomat::SendGreeting(const PLAYER_INDEX & foreignerId)
 }
 bool Diplomat::HasWarOrDesiresPreemptivelyWith(const PLAYER_INDEX foreignerId) const
 {
-	return g_player[m_playerId]->HasWarWith(foreignerId)
-	    || (m_personality->GetAlignmentEvil()
-	       || (m_personality->GetTrustworthinessChaotic() && DesireWarWith(foreignerId))
-	       );
+	return g_player[m_playerId]->HasWarWith(foreignerId) || (/*m_personality->GetAlignmentEvil() || m_personality->GetTrustworthinessChaotic() &&*/ DesireWarWith(foreignerId));
 }
 
 bool Diplomat::DesireWarWith(const PLAYER_INDEX foreignerId) const
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -7230,6 +7230,7 @@ bool ArmyData::MoveIntoForeigner(const MapPoint &pos)
 	      )
 	  )
 	{
+	fprintf(stderr, "%s L%d:\n", __FILE__, __LINE__);
 		Battle(pos, defender);
 	}
 	else
@@ -7292,6 +7293,7 @@ bool ArmyData::MoveIntoForeigner(const MapPoint &pos)
 			// Have to check wehther we desire war
 			if(VerifyAttack(UNIT_ORDER_MOVE_TO, pos, defense_owner))
 			{
+	fprintf(stderr, "%s L%d:\n", __FILE__, __LINE__);
 				Battle(pos, defender);
 			}
 
@@ -7403,6 +7405,7 @@ void ArmyData::Battle(const MapPoint &pos, CellUnitList & defender)
 bool ArmyData::VerifyAttack(UNIT_ORDER_TYPE order, const MapPoint &pos,
 							sint32 defense_owner)
 {
+	fprintf(stderr, "%s L%d:\n", __FILE__, __LINE__);
 	if (g_player[m_owner]->HasWarWith(defense_owner))
 	{
 		return true;
@@ -7438,6 +7441,7 @@ bool ArmyData::VerifyAttack(UNIT_ORDER_TYPE order, const MapPoint &pos,
 	}
 	else
 	{
+	fprintf(stderr, "%s L%d:\n", __FILE__, __LINE__);
 		return true;
 	}
 
--- a/ctp2_code/gs/gameobj/armyevent.cpp
+++ b/ctp2_code/gs/gameobj/armyevent.cpp
@@ -786,15 +786,11 @@ STDEHANDLER(ArmyMoveEvent)
 										   GEA_Army, army,
 										   GEA_End);
 
-					if(!extra){
-					    fprintf(stderr, "%s L%d:\n", __FILE__, __LINE__);
+					if(!extra)
 						return GEV_HD_Continue;
-					    }
 
-					if (newPos != extraOrderPos){
-			    fprintf(stderr, "%s L%d: newPos: %d,%d; extraOrderPos: %d,%d\n", __FILE__, __LINE__, newPos.x, newPos.y, extraOrderPos.x, extraOrderPos.y);
+					if (newPos != extraOrderPos)
 						return GEV_HD_Continue;
-					    }
 				}
 			}
 
@@ -813,10 +809,8 @@ STDEHANDLER(ArmyMoveEvent)
 
 			Assert(!g_player[owner]->IsRobot() || Diplomat::GetDiplomat(owner).HasWarOrDesiresPreemptivelyWith(defender->GetOwner()));
 
-			    fprintf(stderr, "%s L%d: newPos: %d,%d; extraOrderPos: %d,%d\n", __FILE__, __LINE__, newPos.x, newPos.y, extraOrderPos.x, extraOrderPos.y);
 			if(newPos != extraOrderPos){ // about to unintionally bump into another civ's unit
 			        //// so just clear the order for reconsideration
-			    fprintf(stderr, "%s L%d: newPos: %d,%d; extraOrderPos: %d,%d\n", __FILE__, __LINE__, newPos.x, newPos.y, extraOrderPos.x, extraOrderPos.y);
 			        DPRINTF(k_DBG_GAMESTATE, ("Army 0x%lx clears current oders via event due to unit in the way\n", army.m_id));
 				g_gevManager->AddEvent(GEV_INSERT_AfterCurrent,
 									   GEV_ClearOrders,
--- a/ctp2_code/gs/gameobj/armyevent.cpp
+++ b/ctp2_code/gs/gameobj/armyevent.cpp
@@ -809,8 +809,17 @@ STDEHANDLER(ArmyMoveEvent)
 
 			Assert(!g_player[owner]->IsRobot() || Diplomat::GetDiplomat(owner).HasWarOrDesiresPreemptivelyWith(defender->GetOwner()));
 
-			if(g_player[owner]->IsRobot() && army->CanFight(*defender) || g_player[owner]->IsHuman())
-			{
+			if(newPos != extraOrderPos){ // about to unintionally bump into another civ's unit
+			        //// so just clear the order for reconsideration
+			        DPRINTF(k_DBG_GAMESTATE, ("Army 0x%lx clears current oders via event due to unit in the way\n", army.m_id));
+				g_gevManager->AddEvent(GEV_INSERT_AfterCurrent,
+									   GEV_ClearOrders,
+									   GEA_Army, army,
+									   GEA_End);
+				return GEV_HD_Continue;
+			}
+			else{ // not a move order
+			    if(g_player[owner]->IsRobot() && army->CanFight(*defender) || g_player[owner]->IsHuman()){
 				DPRINTF(k_DBG_GAMESTATE, ("Army 0x%lx gets event to attack foreigner\n", army.m_id));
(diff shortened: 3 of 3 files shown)


LynxAbraxas:
@MartinGuehmann What do you think of 4a59fd4b69d234756bb8a6a1929d46fbfe5f48df as a solution?
What other orders should be considered as non-aggressive here? I guess boat-to should be one in case a transporter is joined by a military unit that could attack.

MartinGuehmann:
> @MartinGuehmann What do you think of [4a59fd4](https://github.com/civctp2/civctp2/commit/4a59fd4b69d234756bb8a6a1929d46fbfe5f48df) as a solution?
> What other orders should be considered as non-aggressive here? I guess boat-to should be one in case a transporter is joined by a military unit that could attack.

No, [4a59fd4] is not the solution, you have to go for HasWarOrDesiresPreemptivelyWith. HasWarOrDesiresPreemptivelyWith is used by the path finding algorithm. If you do this, the AI will try to march over that unit again. You should only change the parentheses. Even so, the ones you added just do what the code does anyway, and thus should not be added, since they are superfluous.

However, they should enclose GetAlignmentEvil and GetTrustworthinessChaotic. The change here is that only the AIs with these personality traits would march over foreign units, without declaring war first.

And now let's go to the main problem: You should not change ArmyData::VerifyAttack, otherwise evil AIs could not just march over foreign units. And right now, this is more or less the only way AI can declare war onto another civ, it shouldn't be changed, either. And actually this is also the reason why the code in HasWarOrDesiresPreemptivelyWith is disabled.

For now the AI only declares war on weaker civilizations and only if it has no war with another civilization. And of course if it does not like that civ so much. And here is the problem, what is the right power balance of all the other civs it would declare war upon.

LynxAbraxas:
> you have to go for HasWarOrDesiresPreemptivelyWith. 

But `HasWarOrDesiresPreemptivelyWith` is used only in the assertion in `STDEHANDLER(ArmyMoveEvent)`. Should it be part of the if-statement?
In any case, I do not think this would help in this situation because `HasWarOrDesiresPreemptivelyWith` evaluates to true due to
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/diplomacy/diplomat.cpp#L5160-L5170
evaluating to true in this case. While this should be an issue of balancing, I still think that unintentional bumping into another civ's unit is a different problem. 
With https://github.com/civctp2/civctp2/commit/4a59fd4b69d234756bb8a6a1929d46fbfe5f48df the order is cleared in that case but also if the order is to attack, probably because that internally is also using the move-order to get to the attack-order target point. So I thought, that probably most of the time the unit is met at a position along the order path that lies before the order target, i.e. `newPos != extraOrderPos` in `STDEHANDLER(ArmyMoveEvent)`, however bf23463a7b7c165857867773d3880f05fe4839fe does not work as expected either (not clearing the order in this case, so apparently the condition `newPos != extraOrderPos` is not met).

> You should not change ArmyData::VerifyAttack, otherwise evil AIs could not just march over foreign units. 

Was not the problem, because `HasWarOrDesiresPreemptivelyWith` evaluates to true any way in that case, so removed that commit (b286d63a2a934982145d7f987e527b05e4f4658e).

> and only if it has no war with another civilization. 

That does not seem to be a condition, because the current issue happens even though the attacker is already at war with other civs.

> otherwise evil AIs could not just march over foreign units.

While I agree that this should be possible for evil AIs, the unintentional bumping into a unit needs to lead to a reconsideration. Fore example in another case I had moved a cleric in the way of an army of a Hoplite and a Settler. So when it was their turn to move they bumped into my cleric and so the Hoplite attacked, lost and then the Settler continued the attack and lost as well. This is an action that I do not think would happen if the AI had reconsidered the order for that army.
And I think the key to achieve this is to clear the order in case of unintentional bumping into a unit, then the AI's logic will re-evaluate the possible order, now seeing and considering the unit in the way also taking into account `HasWarOrDesiresPreemptivelyWith`. If that then evaluates to true, it should issue a new attack-order with the order target being the unit in the way. But even an evil AI will not attack a cleric with a Hoplite for protection of a Settler.

So for the mere purpose of testing, do you have an idea how to change bf23463a7b7c165857867773d3880f05fe4839fe such that it leads to the described clearing of the order followed by a reconsideration?

MartinGuehmann:
> > you have to go for HasWarOrDesiresPreemptivelyWith.
> 
> But `HasWarOrDesiresPreemptivelyWith` is used only in the assertion in `STDEHANDLER(ArmyMoveEvent)`.

No, that's wrong. HasWarOrDesiresPreemptivelyWith is used at quite some places, for instance in UnitAstar::CheckUnits. That means when you re-evaluate your order then the AI will chose the same path, marching over the unit with whose civ it desires war with. And there are actually, a few other places where HasWarOrDesiresPreemptivelyWith is used.

> While I agree that this should be possible for evil AIs, the unintentional bumping into a unit needs to lead to a reconsideration. Fore example in another case I had moved a cleric in the way of an army of a Hoplite and a Settler. So when it was their turn to move they bumped into my cleric and so the Hoplite attacked, lost and then the Settler continued the attack and lost as well.

Well, then there is something wrong with the strength calculation, normally I would expect that the cleric loses.

LynxAbraxas:
> HasWarOrDesiresPreemptivelyWith is used at quite some places

I know, I meant within STDEHANDLER(ArmyMoveEvent) it is only used in the assertion.

>  That means when you re-evaluate your order then the AI will chose the same path, marching over the unit with whose civ it desires war with.

Yes, but only if it finds that its own unit/army has a chance of winning in an attack against the newly appeared unit/army. That is where I would expect the AI code to not consider an attack with a Hoplite guarding a Settler, what ever unit is in the way, but instead take a path around.

> Well, then there is something wrong with the strength calculation, normally I would expect that the cleric loses.

I think that is all fine since there is always some chance/randomness in the calculation and the Hoplite's strength is not in attacking.

Anyway, even if you do not think it would solve the given issue, do you have an idea how to change bf23463 such that it leads to the described clearing of the order followed by a reconsideration in case another civ's unit is unexpectedly in the way?

MartinGuehmann:
> Anyway, even if you do not think it would solve the given issue, do you have an idea how to change [bf23463](https://github.com/civctp2/civctp2/commit/bf23463a7b7c165857867773d3880f05fe4839fe) such that it leads to the described clearing of the order followed by a reconsideration in case another civ's unit is unexpectedly in the way?

Actually, the code deals with the fact that a unit may have been invisible:

https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/gs/gameobj/armyevent.cpp#L772-L795

There might be a problem if this does not work: newPos != extraOrderPos

This would also affect your code. However, so far as I can see it should be fine.

However, if you look at UnitAstar::CheckUnits then the AI would also chose a path across the stealth units if it is just to supposed to expel them:

https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/robot/pathing/unitastar.cpp#L239-L277

However, this is not considered in ArmyMoveEvent, and that is the thing that should make the assertion fire. 

Basically, stealth units should be expelled here if the AI has no war or does not desire war, preemptively.

LynxAbraxas:
> There might be a problem if this does not work: newPos != extraOrderPos

My current impression is, that the move-to-order is also used to just move from one tile to the next (according to the A* path), such that `newPos == extraOrderPos` more often than expected, even if the higher order has a different target position.

> However, if you look at UnitAstar::CheckUnits then the AI would also chose a path across the stealth units if it is just to supposed to expel them:

Hm, apparently not only for expelling:
Original path of order:
ss_2020-02-11_19:42:44
Italian unit moved into the path:
ss_2020-02-11_19:43:40
order cleared (due to bf23463a7b7c165857867773d3880f05fe4839fe?):
ss_2020-02-11_19:48:03
new path to same target pos does not take the Italian unit into account:
ss_2020-02-11_19:52:21
but instead again just run over the unit (leading to war):
ss_2020-02-11_21:00:51
I would either expect that the AI then decides to attack the Italian unit which should lead to an attack-order with target 26,65 or that the AI decides to go around the unit (or rest for one turn).

MartinGuehmann:
> order cleared (due to [bf23463](https://github.com/civctp2/civctp2/commit/bf23463a7b7c165857867773d3880f05fe4839fe)?):

No, the AI clears orders after each turn anyway, maybe even after each process cycle and then re-evaluates the situation.

> new path to same target pos does not take the Italian unit into account:

Actually, it does as you can see in the piece of unitastar code I quoted. It is just that it wants to start a war and the unit it runs over is so weak that it can win the fight. That is an opportunity action. So you don't need for that an attack goal.

All what we have to figure out here is whether the AI should really desire war here.

LynxAbraxas:
> No, the AI clears orders after each turn anyway, maybe even after each process cycle and then re-evaluates the situation.

Without https://github.com/civctp2/civctp2/commit/bf23463a7b7c165857867773d3880f05fe4839fe it does not clear the order in that situation.

> All what we have to figure out here is whether the AI should really desire war here.

Well, if this situation represents an opportune action, I'd say the one with the Hoplite and the Settler attacking a Cleric isn't. Nor some others from my previous game, where I often had similar situations and only a turn later the very same civ asked for cease-fire or even peace.

One thing that might help would be to integrate the loss of trade income and bonus from standing trade routes in the evaluation of `Diplomat::DesireWarWith`.

MartinGuehmann:
> > No, the AI clears orders after each turn anyway, maybe even after each process cycle and then re-evaluates the situation.
> 
> Without [bf23463](https://github.com/civctp2/civctp2/commit/bf23463a7b7c165857867773d3880f05fe4839fe) it does not clear the order in that situation.

No, in that situation, you pressed end turn. Then the Italians moved a unit into the path of your Trireme. Then the ended their turn. Then it is your turn again. If the player on screen is a robot. Then the AI clears all orders of all armies, anyway. Then it evaluates the situation, and gives new orders to its armies. Then the orders are executed. Then it re-evaluates the situation again and gives new orders. I also think it clears the orders in between, but that is something I have to look up. It does so five times per turn. That are the process cycles.

> > All what we have to figure out here is whether the AI should really desire war here.
> 
> Well, if this situation represents an opportune action, I'd say the one with the Hoplite and the Settler attacking a Cleric isn't. Nor some others from my previous game, where I often had similar situations and only a turn later the very same civ asked for cease-fire or even peace.

Which situation? The situation with the Coracle? Or the one with the Hopelite, the Settler and the Cleric? These situations are different. The Coracle cannot be expelled, while the Cleric can. Expelling the Cleric is also an opportunity action, but some other opportunity action. The code in unitastar considers this possibility, but not the code in ArmyEvent. In fact, this predicts that the AI should only ask for cease-fire or peace if it did ran over a unit that could have been expelled. 

> One thing that might help would be to integrate the loss of trade income and bonus from standing trade routes in the evaluation of `Diplomat::DesireWarWith`.

I would say this is another issue, but for sure this should be considered.

LynxAbraxas:
> The Coracle cannot be expelled, while the Cleric can.

True, let's regard the above described situation with the fire trime as correctly evaluated as it represents an opportune action and the problem there actually lies in the problem that the AI trade route creation code does not consider `Diplomat::DesireWarWith` nor does the code for evaluating orders considers if precious trade routes exist with the civ it desires war with. I.e.
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/ctp2_code/ai/CityManagement/governor.cpp#L5196-L5214
should consider `Diplomat::DesireWarWith` when creating new trade routes and 
and 
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/ai/diplomacy/diplomat.cpp#L5160-L5170
should get some code that takes the value of standing trade routes into account which should then effect the decision in opportune situation.
I can give it a try to incorporate something suitable in #263.

> Expelling the Cleric is also an opportunity action, but some other opportunity action. The code in unitastar considers this possibility, but not the code in ArmyEvent.

Let's regard this as the actual problem this PR should solve.
Expelling would be the appropriate action while attacking is definitely stupid. This would also fit #224 where the Lawyer got killed instead of expelled.
So do we just lack some code similar to
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/robot/pathing/unitastar.cpp#L254
before
https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/gs/gameobj/armyevent.cpp#L812
?

MartinGuehmann:
> > Expelling the Cleric is also an opportunity action, but some other opportunity action. The code in unitastar considers this possibility, but not the code in ArmyEvent.
> 
> Let's regard this as the actual problem this PR should solve.
> Expelling would be the appropriate action while attacking is definitely stupid. This would also fit #224 where the Lawyer got killed instead of expelled.
> So do we just lack some code similar to
> https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/robot/pathing/unitastar.cpp#L254
> 
> 
> before
> https://github.com/civctp2/civctp2/blob/623e59c962e78a8ff5eed763b02ef43977946fac/ctp2_code/gs/gameobj/armyevent.cpp#L812
> 
> 
> ?

I think so, that it should go there. However, I think that if the AI has war or wants war then it should still kill the special unit. That however also requires to update the code in unitastr.

LynxAbraxas:
> I think so, that it should go there. 

OK, made such a change in a08a2210c7056470b899cb27d7e7e4c10503f8fa and this does prevent the AI (Arabians) to run over a Televangelist (Persian) in this save game [dbgAIUnintionalWar.zip](https://github.com/civctp2/civctp2/files/4395895/dbgAIUnintionalWar.zip) but I feel very unsure about these kind of changes.
@MartinGuehmann your help here would be very much appreciated.

Without the commits of this PR the AI runs over the Televangelist (despite being at peace and even having a trade pact) and asks for cease-fire just in the very next turn.

LynxAbraxas:
Tested this for a while with https://github.com/civctp2/civctp2/pull/264/commits/a08a2210c7056470b899cb27d7e7e4c10503f8fa (or corresponding commits). It has prevented unintentional war sometimes but not always (in which case it might be intentional) however it gives the AI the ability to expel units outside its territory, which a human cannot do (because a human player gets the questions message instead). It would be nicer if the AI instead would re-evaluate its path and change it to go around the unit if it determines that it does not want war, however I probably will not find the time to do so in the near future.

LynxAbraxas:
I won't find the time to work on this in the near future, so closed to serve as a reference. Please feel free to pick it up.

MartinGuehmann:
>however it gives the AI the ability to expel units outside its territory

No, it does not. Unit::CanBeExpelled() checks this, which is called from CellUnitList::CanBeExpelled(). So basically, I arrived at the same conclusion, I guess some extra stuff. However, I missed army->CanExpel().

LynxAbraxas:
Just for the reference: solved with https://github.com/civctp2/civctp2/commit/f4cd3dae3e6a959833e401ee1c3ffa286fe9b1c9 of #437.

PR #273, opened 2020-02-19
[linux] Fix ESC key behavior

### LINKED ISSUE #271: [Linux] ESC key erratic behavior
I expect ESC (Escape) key to close a manager window (i.e. City Manager), if any is opened. And only in case when there are no additional windows opened on screen, ESC should open the Options window.

At the moment only the Great Library window behaves properly, while when there is any other window opened the Options window is shown above that.

Here is a list of affected windows: Empire Manager, National Manager, City Manager, Build Manager, Science Manager, Diplomacy Manager, Unit Manager, Statistics.

Screenshot_20200219_200855
Screenshot_20200219_200942

madand:
Make ESC key close City Manager and similar windows, before popping the game
Options window.

fixes #271

### DIFF
--- a/ctp2_code/ui/aui_common/aui_control.cpp
+++ b/ctp2_code/ui/aui_common/aui_control.cpp
@@ -186,7 +186,7 @@ AUI_ERRCODE aui_Control::InitCommonLdl(
 			else
 				m_actionKey = '^';
 		} else if(stricmp(shortcut, "ESC") == 0) {
-			m_actionKey = VK_ESCAPE;
+			SetActionKeyToESC();
 		} else {
 			m_actionKey = shortcut[0];
 		}
--- a/ctp2_code/ui/aui_common/aui_control.cpp
+++ b/ctp2_code/ui/aui_common/aui_control.cpp
@@ -186,11 +186,7 @@ AUI_ERRCODE aui_Control::InitCommonLdl(
 			else
 				m_actionKey = '^';
 		} else if(stricmp(shortcut, "ESC") == 0) {
-#ifdef WIN32
 			m_actionKey = VK_ESCAPE;
-#else
-			m_actionKey = AUI_KEYBOARD_KEY_ESCAPE;
-#endif
 		} else {
 			m_actionKey = shortcut[0];
 		}
--- a/ctp2_code/ui/aui_common/aui_control.h
+++ b/ctp2_code/ui/aui_common/aui_control.h
@@ -211,6 +211,8 @@ public:
 	virtual aui_Control	*SetKeyboardFocus( void );
 	virtual AUI_ERRCODE	ReleaseKeyboardFocus( void );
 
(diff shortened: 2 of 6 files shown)


MartinGuehmann:
Yeah that is quite bad style to litter the code with these ifdefs. That should be actually how it is supposed to be. Thanks @madand, I just leave it a few more days open if @LynxAbraxas wants to comment on it.

Actually, potentially there might still be more of this kind in.

LynxAbraxas:
Yes, love it. This was something that was annoying me for some while now as well.
Tested this and it works well for me for windows that were opened by a hot-key like F1, F4, F7, W but not for the very same windows if opened with the button of the control panel (unless I hit ESC within about 1 sec after the window appeared (sorry, have no explanation for that behavior so far)). 
It does not close the tile-info window. 
Perhaps some of the code mentioned in #69 is of relevance here.

LynxAbraxas:
Just noticed that the closing does not work if the mouse pointer still is positioned on the control panel (it is not related to the time, it was just that I had not move the mouse pointer in that period).

LynxAbraxas:
> Just noticed that the closing does not work if the mouse pointer still is positioned on the control panel (it is not related to the time, it was just that I had not move the mouse pointer in that period).

Even more precise: The closing does not work for me if a tool-tip is shown.

MartinGuehmann:
> > Just noticed that the closing does not work if the mouse pointer still is positioned on the control panel (it is not related to the time, it was just that I had not move the mouse pointer in that period).
> 
> Even more precise: The closing does not work for me if a tool-tip is shown.

So @LynxAbraxas do you want this included in this PR? I actually tend to have it included. The description is clear. And it seems not too much work.

madand:
Added commits for closing Tile Info window.

madand:
Added new commit:

> Close dialog windows with ESC key
> 
> Easy way to test various dialog windows is to start a new game and settle only 1
> city. Then you can easily get to see 3 types of dialog windows:
> 
> * Informational Dialog - try to disband your only city.
> * Confirmation Dialog - press Clear Queue button in Build Manager.
> * Text Input Dialog - try to rename a city via right click context menu.

madand:
> > Just noticed that the closing does not work if the mouse pointer still is positioned on the control panel (it is not related to the time, it was just that I had not move the mouse pointer in that period).
> 
> Even more precise: The closing does not work for me if a tool-tip is shown.

I pushed the commit which should solve any keyboard-related issues with tooltips. 05afd53

LynxAbraxas:
That's great! Many thanks @madand for those fixes, works like a charm now. If you are fine with this as is, I'd say this can go in. @MartinGuehmann what do you think?

madand:
Yeah, the main pain points has been resolved so it can be merged.

MartinGuehmann:
Well, then we merge.

PR #274, opened 2020-02-21
[linux] Partial fix for permanent 100% CPU usage

### LINKED ISSUE #272: [Linux] Permanent 100% CPU usage, even in main menu
With Linux build I get permanent 100% use of a single CPU core by the game. Even if I just started the game and main menu is on screen. 

In contrast, when I run the windows version (AE 2011-06-11) in [Wine](https://en.wikipedia.org/wiki/Wine_(software)) the CPU usage in main menu is around 46% and around 50% while in-game.

madand:
The event loop throttling for 30 msec implemented by this path reduced the game
CPU load on my laptop (Thinkpad X220 i7 2.7 Ghz) from 100% to around 50% in menu
and 55% while in-game. Not ideal solution, but it prevents the CPU cooler noise
from overheat at 100%.

The proper solution would be to rework the event loop to the blocking mode with
SDL_WaitEvent(). Yet the current hack is still better than nothing.

fixes #272

### DIFF
--- a/ctp2_code/ctp/civ3_main.cpp
+++ b/ctp2_code/ctp/civ3_main.cpp
@@ -1698,12 +1698,16 @@ int WINAPI CivMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR szCmdLine,
 #ifdef __AUI_USE_SDL__
 		SDL_Event event;
 		while (!g_letUIProcess) { // There are breaks, too ;)
-			SDL_PumpEvents();
-			int n = SDL_PeepEvents(&event, 1, SDL_GETEVENT,
+			// Throttle the loop a bit to prevent 100% CPU usage in idle state.
+			// FIXME: implement blocking loop with SDL_WaitEvent().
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
Sounds interesting, will test this and report back. Let's see if @MartinGuehmann has some feedback on this.

MartinGuehmann:
Sure, you don't want the game to use 100% of the CPU if it is not necessary. Maybe that could also be linked to some other issue. My feeling was that the Windows version was faster than the Linux version. That could be linked to it.

Another thing is that we haven't really worked on these SDL related issues, since it is has to be reworked for SDL2, anyway. And that wasn't really our priority. (Related issue #92)

Otherwise, as you have said it is better then noting and I would go for it.

LynxAbraxas:
Seems to work (at least slightly less CPU usage) and not to break anything obvious, so good to go in.

MartinGuehmann:
Then let's tick this off.

PR #276, opened 2020-02-21
Rename files to the same case as in the GOG version

### LINKED ISSUE #268: File names case difference in comparison to version by GOG
Readme's section [Install the Apolyton edition](https://github.com/civctp2/civctp2#install-the-apolyton-edition) claims this repo's `ctp2_data` is compatible with the GOG version of the game, in terms of file name casing. I found a few files which happen to have different case and thus will not be overwritten when copying AE data over the GOG one.

All the files are in the directory `ctp2_data/default/aidata`:
| GOG                    | AE                     |
|------------------------|------------------------|
| AdvanceLists.txt       | advancelists.txt       |
| BuildListSequences.txt | buildlistsequences.txt |
| UnitBuildLists.txt     | unitbuildlists.txt     |

I suggest to rename our files to have the same case as the GOG ones.

madand:
Renamed files:
  advancelists.txt       -> AdvanceLists.txt
  buildlistsequences.txt -> BuildListSequences.txt
  unitbuildlists.txt     -> UnitBuildLists.txt

Also added `gamefile.txt` into the repository and updated its entries to have
exactly the same case as the actual file names in the `ctp2_data` directory.

fixes #268


LynxAbraxas:
I would recommend to split that commit into 2, one just for the rename and the other for the `gamefile.txt`, telling in the commit message what it is good for, perhaps that should also be mentioned in the readme or similar.

madand:
Splat into 2 commits, as you asked, and force-pushed them.

LynxAbraxas:
Thanks, that way it is easier in future to revert some commit, though it is generally hard to foresee if that will be needed at some point or not at all.

madand:
I agree and will try to make PRs with more granular commits in the future. 

It is just mental inertia, sine I frequently contribute stuff to [Spacemacs](https://github.com/syl20bnr/spacemacs), and these guys do not merge commits but rather cherry-pick them, with the policy to squash everything into a single commit before opening PR. Makes sense for high throughput repo like theirs. Sorry for off-topic.

MartinGuehmann:
> Sorry for off-topic.

Don't worry about this. GitHub is also a social network. And for sure these things must also be said somewhere. In principle, we could also write down these politics somewhere. But since we are so small it is not that necessary. Even so it would be nice to have.

MartinGuehmann:
So, I guess we are now happy with it. If nobody objects I will merge it, tomorrow.

LynxAbraxas:
Please, give me some more time, because this PR needs an addition to ctp2CD and ideally a change to
https://github.com/civctp2/civctp2/blob/67954be9d07cfe5944dec129d0a5f21114a9682b/.gitlab-ci.yml#L44
that binds the artifact curl command to the SHA of the commit in ctp2CD (currently it just get the one master points to).

MartinGuehmann:
OK, I wasn't quite aware of that, so let us know, when I can merge.

LynxAbraxas:
Hm, seems getting an artifact by commit SHA is not supported:
https://docs.gitlab.com/ee/api/jobs.html#download-the-artifacts-archive
So let's go for the simpler option and leave things as they are until we need to reference different artifacts.
So I'd think this PR can go in but first https://gitlab.com/civctp2/ctp2CD needs to be updated to b672316fb11cf2f4a10b9d172d606bb093fec0a7 (created a GitLab MR for that) and waited until the new artifact is ready (about 1 Min.) then merge this and we'll see if the smoke tests on GL still work.

MartinGuehmann:
Merged on GitLab, and it passed. So let's merge this.

PR #277, opened 2020-02-21
Update Readme with info on preparing the GOG version data

### LINKED ISSUE #267: [Linux] Build is sucessfull, but segfault when trying to start a new game
It would be very good to have regular or semi-regular (like once a month/quarter/year) releases of builds for Windows (Win7 x86 might be a good baseline). So a user is able to just download a zip archive, replace existing files and start playing right away.

At the moment, the 'latest' prebuilt version of CTP:AE is from 2011-06-11, which is pretty sad considering how much stuff is being here  by @LynxAbraxas and @MartinGuehmann lately.

PS: I am a GNU/Linux user, but all my attempts to play the current code from this repo were futile to date (tried Docker build - segfault; installed real Ubuntu 18.04 and replicated steps from Dockerfile - also got segfault when i press the 'New Game' button). On the other hand, I am successfully playing the mentioned old build of CTP:AE 2011-06-11 in WINE, so I expect windows builds of the current code to work in WINE as well.

madand:
fixes #267 and #268

This is alternative solution for #268 without modifying or renaming anything (except Readme) on our side. @LynxAbraxas  after I realized that #276 will mess with CI build, since `ctp2CD` directory contains files in lower case, I think that giving instructions in Readme is better than messing with file names in the repo. If you agree, you can just close #276 without merging and merge this PR instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,11 @@ First, you need the original game files of CTP2.
 If you have the CD version, it's better to install it on a Windows machine / using Wine and copy over the files. You need the folders `ctp2_data`, `ctp2_program` and `Scenarios`.
 
 If you have the GoG version, use `innoextract` to extract all the required files (you can `sudo apt install innoextract` if you need) from the `setup_call_to_power2_2.0.0.13.exe` windows installer.
-For example `innoextract -m -I app  setup_call_to_power2_2.0.0.13.exe` will extract the `app`folder that contains all the game files, which are in the same subfolders as in the CD version: `ctp2_data`, `ctp2_program` and `Scenarios`.
+For example `innoextract -m -I app  setup_call_to_power2_2.0.0.13.exe` will extract the `app` folder that contains all the game files, which are in the same subfolders as in the CD version: `ctp2_data`, `ctp2_program` and `Scenarios`. GOG' version of `ctp2_data` folder needs some files deleted to be fully compatible with AE:
+```bash
+rm ctp2_data/default/aidata/{AdvanceLists,BuildListSequences,UnitBuildLists}.txt
+rm ctp2_data/english/uidata/layouts/*
+```
 
 Put those folder in you `home` path or anywhere else, for example `~/ctp2`
 


MartinGuehmann:
> since `ctp2CD` directory contains files in lower case, I think that giving instructions in Readme is better than messing with file names in the repo.

Which files? The ones you propose to rename? Or some others?

madand:
@MartinGuehmann, yes the files I listed #268.

LynxAbraxas:
Need some time to think about this. 
Generally, renaming in `ctp2CD` is possible and should not be a big deal. The question is, are more potential CTP2 gamer having an Activision-CD or a GOG version? We should try to make it as easy as possible for newcomers to get started with the current version.

LynxAbraxas:
Just checked that the GOG case is also that which my Activision CD files have. Actually these 3 files were the only ones that were changed by
```
paste -d ' '  <(find ~/ctp2/ctp2_data/ -type f | grep -i -oFf <(find ctp2_data/ -type f) | grep -vFf <(find ctp2_data/ -type f)  | grep -i -oFf - <(find ctp2_data/ -type f) | sort -f) <(find ~/ctp2/ctp2_data/ -type f | grep -i -oFf <(find ctp2_data/ -type f) | grep -vFf <(find ctp2_data/ -type f) | sort -f)  | sed 's/ /\n/' | parallel   -N2 'mv {1} {2}'
```
I would favor this PR, and will try to test the changes of this PR with the GL-CI and adjust the pull of ctp2CD to work with it and report back than.

LynxAbraxas:
In conjunction to e54211a07d05e0e239890df9d8a557c3c09be275, I wonder if there is a way to get an artifact by commit SHA instead of just the newest of master, will investigate the current GL-CI API. The GL issue was moved and apparently now is only meant to be realized for the commercial version of GL, so no change in that artifact/curl workaround.

madand:
I'd like to point out that the part about english layouts directory cleanup:
```
rm ctp2_data/english/uidata/layouts/*
```
is relevant not only for Linux and not only for GOG, but for *all* users of *any* distribution of CTP2.

MartinGuehmann:
> The question is, are more potential CTP2 gamer having an Activision-CD or a GOG version?

I would expect that newcomers would come with a GOG version, today. Additionally, there are at least two kinds of CDs, one that's the original release version, without the patch and another re-release version with the patch.

I have both and from one of them I copied the files onto my computer and that has the same case as the GOG version. However, I've moved to Bristol and the CDs are still at my parents' place in Berlin, so I can't check them.

Anyway, I would also go for renaming.

MartinGuehmann:
> I'd like to point out that the part about english layouts directory cleanup:
> 
> ```
> rm ctp2_data/english/uidata/layouts/*
> ```
> 
> is relevant not only for Linux and not only for GOG, but for _all_ users of _any_ distribution of CTP2.

In fact that is not only for the english/layouts directory, but also applies to the other languages, as well.

In fact, I would prefer something with the original intention of my change. Namely that those wfiles would be ignored. But well, I have to see.

LynxAbraxas:
> > I'd like to point out that the part about english layouts directory cleanup:
> > ```
> > rm ctp2_data/english/uidata/layouts/*
> > ```
> > 
> > 
> > is relevant not only for Linux and not only for GOG, but for _all_ users of _any_ distribution of CTP2.
> 
> In fact that is not only for the english/layouts directory, but also applies to the other languages, as well.
> 
> In fact, I would prefer something with the original intention of my change. Namely that those wfiles would be ignored. But well, I have to see.

OK, then let's note that we want both PRs this and #276. This PR for handling/removing LDLs from GOG/CD and #276 to avoid renaming.

madand:
We certainly want this handled in a way that keeps AE installation for the end-user as simple as possible. Namely, unzip/overwrite files and play, with no further fiddling with the game files. So:
1. #276 should be merged.
2. This PR should be rejected, if @MartinGuehmann figures out a way to just make files inside `ctp2_data/english/uidata/layouts/` (and other language-specific subdirs) be ignored by the game.

LynxAbraxas:
Another way could be to soft-link the current LDLs to `ctp2_data/english/uidata/layouts/` such that they would overwrite the old ones but as a soft-link. That way new users would not have to do the extra step of removing the old ones and only the main LDLs have to be cared about in modifications. However, I am not sure how git handles soft-links on windows checkouts.

LynxAbraxas:
How about merging this for the time being, and revert in a new PR if a better solution (that does not involve deletion of files) is found?
@MartinGuehmann what do you think?

MartinGuehmann:
> 2. This PR should be rejected, if @MartinGuehmann figures out a way to just make files inside `ctp2_data/english/uidata/layouts/` (and other language-specific subdirs) be ignored by the game.

Rejected in favor of #288.

PR #278, opened 2020-03-01
Add Polish text files, probably those are v1.1

MartinGuehmann:
I found a Polish translation of CTP2 on BureauBert's website http://www.ctp2.info/getctp2/language.php.

So should be useful to incorporate this.


MartinGuehmann:
This is now part of #431, so can be closed.

PR #279, opened 2020-03-06
Copy libs in DF first since they are less likely to change

LynxAbraxas:
and therefore represent a more stable docker layer. That way only one layer, that of
https://github.com/civctp2/civctp2/blob/45cd01bb9f54502c2dfa13e8acd40558fb2af7c0/Dockerfile#L97-L98
has to be downloaded and stored individually. So when downloading successive DIs from different builds less than 30MB need to be downloaded and stored additionally, all other layers can be shared between the dockerimages as long as no change happened to the libs or the ctp2CD, which is far less likely to happen.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,13 +92,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+## copy libs first since they are less likely to change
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
I assume this is now ready for merge.

LynxAbraxas:
Yes, ready.

MartinGuehmann:
So then let's merge.

PR #252, opened 2020-01-24
Debug AI route creation

LynxAbraxas:
PR to resolve #155.

PR #280, opened 2020-03-19
Speedup trade route cost evaluation

LynxAbraxas:
The use of astar for the exact cost evaluation of a trade route as in
https://github.com/civctp2/civctp2/blob/b2aaae74c063d8e86b281ffb826d252fe30a8c55/ctp2_code/gs/gameobj/tradeutil.cpp#L80-L93
is significantly more time consuming than the approximation by
https://github.com/civctp2/civctp2/blob/b2aaae74c063d8e86b281ffb826d252fe30a8c55/ctp2_code/gs/gameobj/tradeutil.cpp#L96-L103
which was used in
https://github.com/civctp2/civctp2/blob/ea3274c3d43176670ac71f6bc75703fac18a226e/ctp2_code/ai/CityManagement/governor.cpp#L5137
before db7b3e0861a90f464d1cf502ed81c494f3bdc38f of #252.
This was not noticeable until about 500AD and later (see also https://github.com/civctp2/civctp2/pull/263#issuecomment-595757346), because the AI had not many trade routes to evaluate. In later games after about 300 turns this becomes noticeable as finishing a turn for a single AI player can take up to a minute then when a few 100 possibly a few 1000 trade route combinations are checked for their value.

Just reverting db7b3e0861a90f464d1cf502ed81c494f3bdc38f leads again to the problem that the AI trade route creation gets unstable because then the fast estimate of the cost of a trade route in general differs to that of the exact cost as used during creation so the AI cancels routes in the expectation that it can create better ones that then turn out to be as bad or even worse than the former ones.

So this PR instead deprecates the accurate cost evaluation and only makes use of the estimate. For that to work the accurate cost evaluated during creation must be overwritten by the estimate used everywhere else
https://github.com/LynxAbraxas/civctp2/blob/df0db714a420071439de8bfd81de901ae847cec2/ctp2_code/gs/gameobj/TradeRouteData.cpp#L110

While this leads to consistent trade route costs and stable trade routes from the AI, it is not as nice as a speed up of the accurate trade cost evaluation. However that would need a significant speed up of the astar implementation, which would effect many aspects of the code and could easily lead to the introduction of critical new bugs. So this PR represents a workaround for the time being.

### DIFF
--- a/ctp2_code/ai/CityManagement/governor.cpp
+++ b/ctp2_code/ai/CityManagement/governor.cpp
@@ -5136,10 +5136,8 @@ void Governor::ManageGoodsTradeRoutes()
 
 						const sint32 price = tradeutil_GetTradeValue(m_playerId, destCity, g); // # of gold
 						const sint32 cost = tradeutil_GetTradeDistance(city, destCity); // # of caravans
-						const sint32 cost2 = tradeutil_GetAccurateTradeDistance(city, destCity); // # of caravans
 						const double valuePerCaravan = static_cast<double>(price) / cost; // cost = tradeutil_GetAccurateTradeDistance returns > 1.0
-						fprintf(stderr, "%s L%d: GetAccurateTradeDistance - GetTradeDistance: %d, valuePerCaravan: %f (%f)\n", __FILE__, __LINE__, cost2 - cost, valuePerCaravan, static_cast<double>(price) / cost2);
-						
+
 						if (valuePerCaravan > bestValuePerCaravan) // determine best offer (and its cost) that would be available, to set goal for # of caravans (m_neededFreight)
 						{
 							maxNeededFreight = cost;
--- a/ctp2_code/ai/CityManagement/governor.cpp
+++ b/ctp2_code/ai/CityManagement/governor.cpp
@@ -5089,7 +5089,7 @@ void Governor::ManageGoodsTradeRoutes()
 				{
 					sellingCost = curDestRoute->GetCost();
 					sellingVPC = static_cast<double>(tradeutil_GetTradeValue(m_playerId, curDestRoute->GetDestination(), g))
-					    / sellingCost; // tradeutil_GetAccurateTradeDistance returns > 1.0
+					    / sellingCost; // tradeutil_GetTradeDistance returns > 1.0
 				}
 				else
 				{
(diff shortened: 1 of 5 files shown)


LynxAbraxas:
The use of `tradeutil_GetTradeDistance` in
https://github.com/civctp2/civctp2/blob/df0db714a420071439de8bfd81de901ae847cec2/ctp2_code/ui/interface/trademanager.cpp#L424
also significantly speeds up the setup of the trade route list in the market tab of the trade manager, still this takes about 25 seconds for me when 5 cities per good are selected to be show. About twice the time is needed for the game to be reactive again when a trade route is created, which makes me wonder if the set up of the new list afterwards is executed twice (probably without need).

MartinGuehmann:
> The use of `tradeutil_GetTradeDistance` in
> https://github.com/civctp2/civctp2/blob/df0db714a420071439de8bfd81de901ae847cec2/ctp2_code/ui/interface/trademanager.cpp#L424
> 
> 
> also significantly speeds up the setup of the trade route list in the market tab of the trade manager, still this takes about 25 seconds for me when 5 cities per good are selected to be show and about a 100 cities are known.
> About twice the time is needed for the game to be reactive again when a trade route is created, which makes me wonder if the set up of the new list afterwards is executed twice (probably without need).

If it is about the user interface then it could be that it sorts all the entries whenever one is added. That was the case for the unit manager. And that took quite some time with 800 units around.

LynxAbraxas:
> If it is about the user interface then it could be that it sorts all the entries whenever one is added. That was the case for the unit manager. And that took quite some time with 800 units around.

Thanks for that hint but it seems not to be the sorting that takes the time because re-sorting only takes about a sec or two. It really seems to be the creation, and this seems to be done much more often than necessary.

LynxAbraxas:
Seems the point of this PR is working all right, so created a new PR for the trade manager issue, which is not related to this PR.

LynxAbraxas:
Just a screenshot showing that this PR can lead to long trade routes that cost relatively few trade units, here only 2 caravans (appropriate for the direct distance) for the long route (due to unexplored tiles) from Brighton to Alaca Huyuk:
ss_2020-03-21_02:06:42

MartinGuehmann:
I guess there should some code that may readjust the trade route paths, if the terrain gets explored in the mean time, if there is no such code in place. Anyway, then let's merge.

LynxAbraxas:
> I guess there should some code that may readjust the trade route paths, if the terrain gets explored in the mean time, if there is no such code in place.

No, not that I know of. Once the path of a trade route is generated (during creation) that is fixed and never changed again. If the terrain gets explored an a new trade route is created the new path uses the newly explored tiles:
ss_2020-03-22_00:09:02

PR #282, opened 2020-03-21
Speedup trade manager

LynxAbraxas:
PR to speed up trade manager interaction, see https://github.com/civctp2/civctp2/pull/280#issuecomment-601096981.

### DIFF
--- a/ctp2_code/ai/CityManagement/governor.cpp
+++ b/ctp2_code/ai/CityManagement/governor.cpp
@@ -5118,7 +5118,7 @@ void Governor::ManageGoodsTradeRoutes()
 							continue;
 					}
 
-					for (sint32 d = 0; d < g_player[op]->m_all_cities->Num(); d++) { // check all cities if suitabelfor trading
+					for (sint32 d = 0; d < g_player[op]->m_all_cities->Num(); d++) { // check all cities if suitabel for trading
 						Unit destCity = g_player[op]->m_all_cities->Access(d);
 
 						if(!(destCity.GetVisibility() & (1 << m_playerId))) // city not visible/known
--- a/ctp2_code/gs/gameobj/TradeRoute.cpp
+++ b/ctp2_code/gs/gameobj/TradeRoute.cpp
@@ -158,8 +158,6 @@ void TradeRoute::RemoveAllReferences(CAUSE_KILL_TRADE_ROUTE cause)
 	}
 
 	g_theTradePool->Remove(*this);
-
-	TradeManager::Notify();
 }
 
 Unit TradeRoute::GetSource() const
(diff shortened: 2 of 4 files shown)


MartinGuehmann:
Actually, it does sort the list whenever it adds an item. And that's when it is creating it. To stop this, add in TradeManager::UpdateCreateList after the line that says "m_createList->Clear();" the line: 
m_createList->BuildListStart();

And at the end of TradeManager::UpdateCreateList put: m_createList->BuildListEnd();

TradeManager::UpdateSummaryList has the same problem, which can be fixed in the same way.

LynxAbraxas:
Now I see what you mean, it is sorting the list during creation. Preventing this really makes a difference. Many thanks @MartinGuehmann for that hint and the instructions.
Now what is left for this PR is to find out what causes multiple calls to Update in a row. With cf412ae25443556cb8e00ff7f35ace05e2ed323a this only happens twice for the creation of a route, before cf412ae25443556cb8e00ff7f35ace05e2ed323a it happened even four times in case of cancelling a route.

LynxAbraxas:
Despite 5ed8bd0f4bd1fccb5fedf1f0ab439ffb696d4094 when war is declared there are as many calls to `TradeManager::Update()` as there are trade routes between the two parties. This is due to the calls of `TradeManager::Notify()` in `Player.cpp`, which make sure that the infos in the advice sub-window of the trade manager are updated when creating or cancelling routes from within the trade manager.
So the changes of aafb97ca7f56e9555247964517bb60d28cc4e509 are meant to avoid the various update calls in `TradeManager::Update()` in case the trade manager is not visible, but apparently the condition in the if-statement is true even if the window is not visible. @MartinGuehmann  any idea why or what other check might do the trick?

MartinGuehmann:
> Despite [5ed8bd0](https://github.com/civctp2/civctp2/commit/5ed8bd0f4bd1fccb5fedf1f0ab439ffb696d4094) when war is declared there are as many calls to `TradeManager::Update()` as there are trade routes between the two parties. This is due to the calls of `TradeManager::Notify()` in `Player.cpp`, which make sure that the infos in the advice sub-window of the trade manager are updated when creating or cancelling routes from within the trade manager.
> So the changes of [aafb97c](https://github.com/civctp2/civctp2/commit/aafb97ca7f56e9555247964517bb60d28cc4e509) are meant to avoid the various update calls in `TradeManager::Update()` in case the trade manager is not visible, but apparently the condition in the if-statement is true even if the window is not visible. @MartinGuehmann any idea why or what other check might do the trick?

So far I can only say that the condition in TradeManager::Notify() is already supposed to prevent the updating if the window is not visible.

LynxAbraxas:
> So far I can only say that the condition in TradeManager::Notify() is already supposed to prevent the updating if the window is not visible.

Thanks for the hint. I was understanding `g_c3ui->GetWindow` to check if the window got created/built, not necessarily if it is shown. Perhaps `g_c3ui->GetWindow(s_tradeManager->m_window->Id())->IsHidden()` is needed.

LynxAbraxas:
Tested this for some time now and seems to work fine. @MartinGuehmann what do you think? Ready for merge?

MartinGuehmann:
Looks good.

PR #283, opened 2020-03-22
[linux] Fix empty scenario selection window on New Game

### LINKED ISSUE #270: [Linux] New game 'Scenario Selection' window is empty
Screenshot_20200219_192329

Then I tried to run the game with `strace` and found that the game actually checks the existence of `Scenarios` directory at some point:

```
stat("../../Scenarios", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
close(19)                               = 0
getcwd("/home/madand/games/linux/ctp2/ctp2_program/ctp", 4096) = 47
lstat("/home/madand/games/linux/ctp2/Scenarios", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
openat(AT_FDCWD, "/home/madand/games/linux/ctp2/Scenarios", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 19
fstat(19, {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
getdents64(19, /* 7 entries */, 32768)  = 232
stat("/home/madand/games/linux/ctp2/Scenarios/.", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
stat("/home/madand/games/linux/ctp2/Scenarios/..", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
stat("/home/madand/games/linux/ctp2/Scenarios/WorldMaps", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
stat("/home/madand/games/linux/ctp2/Scenarios/NuclearDetente", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
stat("/home/madand/games/linux/ctp2/Scenarios/AlexanderTheGreat", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
stat("/home/madand/games/linux/ctp2/Scenarios/MagnificentSamurai", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0
stat("/home/madand/games/linux/ctp2/Scenarios/AE_Mod", {st_mode=S_IFDIR|0755, st_size=3488, ...}) = 0

```

madand:
stat(2) returns 0 on success and -1 on error, so the condition was wrong in this
case.

fixes #270

### DIFF
--- a/ctp2_code/gs/fileio/civscenarios.cpp
+++ b/ctp2_code/gs/fileio/civscenarios.cpp
@@ -244,7 +244,7 @@ void CivScenarios::LoadData(void)
 		if (!dent)
 			continue;
 		snprintf(path, sizeof(path), "%s%s%s", rootPath, FILE_SEP, dent->d_name);
-		if (!stat(path, &tmpstat))
+		if (stat(path, &tmpstat) == -1)
 			continue;
 
 		if (S_ISDIR(tmpstat.st_mode)) {


LynxAbraxas:
Many thanks @madand for the debugging, testing now with a fix to #284.

LynxAbraxas:
Yes, your fix does make the scenarios be listed, that's great. Thanks. This was basically the reason why I did not test scenario loading so far and so did not notice #284. Sorry for that, created #286 for fixing that and other issues in conjunction with loading scenarios.
@MartinGuehmann what do you think, read for merge?

MartinGuehmann:
I am still wondering whether it would be better for the condition to read "!= 0" instead of "== -1". This way it is also more general and would also account for changes of the implementation of stat.

Right now, I am a bit undecided about this. So what do you think?

madand:
`stat` is the part of POSIX standard [library](https://en.wikibooks.org/wiki/C_Programming/POSIX_Reference/sys/stat.h) and I don't see why any implementation would dare to change the return value semantics. As for my brain, processing negation `!=` conditionals is always more mentally demanding than simple `==`, [YMMV](https://www.urbandictionary.com/define.php?term=ymmv).

MartinGuehmann:
I remember that when you compare to zero than to another value then you need less machine instructions, to do so. So that should be faster. However, I cannot really find that piece of information. And even if true it is not the part of code that needs to be optimized for speed.

So let's merge.

LynxAbraxas:
Hm, just checked:
```
find  -type f -exec grep -Hn 'stat(path' {} \;
./ctp2_code/gfx/tilesys/tileset.cpp:841:	int rc = stat(path, &st);
./ctp2_code/gs/fileio/civscenarios.cpp:247:		if (stat(path, &tmpstat) == -1)
./ctp2_code/gs/fileio/civscenarios.cpp:465:	if(!_stat(path, &tmpstat)) {
./ctp2_code/gs/fileio/civscenarios.cpp:467:	if(!stat(path, &tmpstat)) {
./ctp2_code/gs/fileio/GameFile.cpp:1746:		if (stat(path, &tmpstat))
./ctp2_code/gs/fileio/GameFile.cpp:1791:				if (stat(path, &tmpstat)) continue;
./ctp2_code/gs/fileio/GameFile.cpp:2327:		if (!stat(path, &tmpstat))
./ctp2_code/gs/fileio/GameFile.cpp:2367:				if (!stat(path, &tmpstat)) continue;
./ctp2_code/libs/anet/src/aeh/analaeh/aehsplit.c:54:	if (-1 == stat(path, &st)) {
./ctp2_code/ctp/ctp2_utils/c3files.cpp:312:	return !_stat(path, &tmpstat);
```
So anet uses `if (-1 == stat(path, &st)) ` in which case I prefer `if (stat(path, &tmpstat) == -1)`.
@madand  would you mind checking which other hits should be changed accordingly?
I guess there is at least one missing for saving maps from the map editor/cheat mode.

LynxAbraxas:
Sorry, page was outdated so I didn't see this was already merged.

MartinGuehmann:
Well, remove the path part from the search string and you find more instance of it. I find the ones in GameFile.cpp a bit suspicious.

madand:
I reviewed the uses of `stat` across the codebase. Here are the seemingly wrong ones:

https://github.com/civctp2/civctp2/blob/bd50117ce68fcf63490bf02a9549390857d55a0c/ctp2_code/gs/fileio/GameFile.cpp#L2327-L2328

https://github.com/civctp2/civctp2/blob/bd50117ce68fcf63490bf02a9549390857d55a0c/ctp2_code/gs/fileio/GameFile.cpp#L2367

https://github.com/civctp2/civctp2/blob/bd50117ce68fcf63490bf02a9549390857d55a0c/ctp2_code/gs/fileio/prjfile.cpp#L418-L419

I believe they all need to be changed to `== -1`.

Also there is quite a few (26 not counting in `anet` lib) uses of `stat` across the codebase, in different variations: `== 0`, `!= 0`, `if (stat(foo, &bar)` and `if (!stat(foo, &bar)`. I would refactor them into explicit `== 0` and `== -1`. Please let me know whether you would agree to merge such a PR.

MartinGuehmann:
For sure, stat does not return a bool and I prefer to have the if statements contain things of type bool. And I also prefer to have the operator ! only be used on bools and not on numbers, since it is supposed to be a boolean operator.

In fact, I would like to see these things be changed. So please go on @madand. Maybe leave the anet library alone. I am not sure whether we have modified it, but we may still use a newer version of it at one day for an unforeseen reason.

Issue #284, opened 2020-03-22
Error when loading AE Mod Scenario


madand:
![Screenshot_20200322_174941](https://user-images.githubusercontent.com/2280844/77253821-a725f200-6c65-11ea-9000-fdceac07675a.png)

I guess @LynxAbraxas forgot to introduce some changes from his `ctp2CD` directory into this repository.

LynxAbraxas:
Seems I forgot the `Const.txt` from the Scenarios in 398f3a29cf555053ea9ba16c6b6a253bc68d8151. Just adding
```
TURNS_FRANCHISED        -1      # How many turns a franchise will be effective, -1: infinit
```
or similar should do the trick.

MartinGuehmann:
Well, @LynxAbraxas, you could have just added a default value for the new flag, then you don't need to edit any const.txts from scenarios or mods.

LynxAbraxas:
I'll check, not sure where I missed that.

MartinGuehmann:
It's in Const.cdb, just scroll to the bottom of that lengthy list to see how it is done.

LynxAbraxas:
I see, kind of obvious, not sure why I have missed that then, sorry for that. #286 contains a fix.

PR #286, opened 2020-03-24
Adjustments for scenarios

LynxAbraxas:
This PR (based on #283) solves #284 and other current issues with loading scenarios (under Linux).


LynxAbraxas:
Is it correct that one should copy the Scenarios folder from the repo over the CD/GOG folder? If so, I guess this should be pointed out in the README.md.

LynxAbraxas:
Has it been tested that the scenarios at least once were fully functional (under windows)? I wonder because I have not seen any code in current ctp2 that would realize a check if "you found a wife" and "have a son" as instructed in the Alexander scenario.

LynxAbraxas:
Loading starts all right with the current state of this PR, but then messages like
```
UNIT_BUILD_LIST_SPECIAL_MISSIONARY not found in UnitBuildList database
```
appear. Continuing with "Yes" generally works, but then the game crashes before the map appears.
@MartinGuehmann are the scenarios compatible to the current save-game format?
@madand how does it work for you so far?

MartinGuehmann:
> Is it correct that one should copy the Scenarios folder from the repo over the CD/GOG folder? If so, I guess this should be pointed out in the README.md.

I guess so, and I wanted to sort out the ldl thing first.

MartinGuehmann:
> Has it been tested that the scenarios at least once were fully functional (under windows)? I wonder because I have not seen any code in current ctp2 that would realize a check if "you found a wife" and "have a son" as instructed in the Alexander scenario.

I don't think so. At least for the Magnificent Samurai scenario something was missing when I played it.

So I guess there are a few issues with these.

MartinGuehmann:
> Loading starts all right with the current state of this PR, but then messages like
> 
> ```
> UNIT_BUILD_LIST_SPECIAL_MISSIONARY not found in UnitBuildList database
> ```
> 
> appear. Continuing with "Yes" generally works, but then the game crashes before the map appears.
> @MartinGuehmann are the scenarios compatible to the current save-game format?
> @madand how does it work for you so far?

No, the savegame format should be compatible, however those scenarios maybe even in a pre-CTP2-release save game format.

I would assume that this scenario has its own build list, and I didn't simply care about the scenarios to update them.

LynxAbraxas:
> > Is it correct that one should copy the Scenarios folder from the repo over the CD/GOG folder? If so, I guess this should be pointed out in the README.md.
> 
> I guess so, and I wanted to sort out the ldl thing first.

That's the way I test it now, and the AE scenario does work (after ignoring `ICON_GOOD_WATER_SHELF_FOUR not found in Icon database`).

> No, the savegame format should be compatible, however those scenarios maybe even in a pre-CTP2-release save game format.

Yes, that could be the problem for those. So I guess we would need a save-game converter for getting those to work with the current CTP2.

MartinGuehmann:
> Yes, that could be the problem for those. So I guess we would need a save-game converter for getting those to work with the current CTP2.

No, the code cares about that. But we have to update the text files.

madand:
> Loading starts all right with the current state of this PR, but then messages like
> 
> ```
> UNIT_BUILD_LIST_SPECIAL_MISSIONARY not found in UnitBuildList database
> ```
> 
> appear. Continuing with "Yes" generally works, but then the game crashes before the map appears.
> @MartinGuehmann are the scenarios compatible to the current save-game format?
> @madand how does it work for you so far?

I've got similar `not found in UnitBuildList` error.

@LynxAbraxas why you put `Scenarios` under the `.civctp2` directory instead of copying over during the `builder` stage in `Dockerfile`? Also this should be mentioned as part of the installation process in the Readme.

LynxAbraxas:
> @LynxAbraxas why you put Scenarios under the .civctp2 directory instead of copying over during the builder stage in Dockerfile? Also this should be mentioned as part of the installation process in the Readme.

I had it in the DF (ef0be5643c0e52de6a7224ab8990174dd4527e47) first, but then figured that this is an unnecessary layer in the DI and mounting it into the DI allows to create new scenarios which will get saved outside the DI, so do not get lost when the docker container is removed. Have to test this a bit more though (it did not work for the maps/ folder so far, but my tests were without your #284).

LynxAbraxas:
> > Yes, that could be the problem for those. So I guess we would need a save-game converter for getting those to work with the current CTP2.
> 
> No, the code cares about that. But we have to update the text files.

OK, which ones?

MartinGuehmann:
> > > Yes, that could be the problem for those. So I guess we would need a save-game converter for getting those to work with the current CTP2.
> > 
> > 
> > No, the code cares about that. But we have to update the text files.
> 
> OK, which ones?

The code that loads the save games, there are different versions of the save game format. And it checks for that.

LynxAbraxas:
With c0448b5f6a3d96a1a457013eba62b226ad9ef920, AE mod starts without issues for me now. MagnificentSamurai probably needs an update to `goods.txt` but I will wait with that until the Activision scenarios do not crash at the end of loading any more.

LynxAbraxas:
Tested this a few times and seems to fix #284 all right. I do not have the time currently to look into the issue of the loading of the original scenarios, so this is good to go in from my side. 
@MartinGuehmann would you open an issue describing what needs to be done for the  original scenarios because so far it is not clear to me.

MartinGuehmann:
> Tested this a few times and seems to fix #284 all right. I do not have the time currently to look into the issue of the loading of the original scenarios, so this is good to go in from my side.
> @MartinGuehmann would you open an issue describing what needs to be done for the original scenarios because so far it is not clear to me.

Let's see whether I can just do it. Otherwise, I'll open an issue, but properly only that this needs to be fixed.

So then let's merge.

Issue #228, opened 2020-01-03
Convert code from DOS to Unix EOL


LynxAbraxas:
Now, since git is used for managing the code, the DOS EOL (`\r\n`) could be changed to Unix EOL (`\n`) because git has the feature to adjust the EOL according to the OS used when cloning the repo. This way we could get rid of the various mixed states that currently exist and that introduce unecessary new lines when viewing in Linux editors or even in GH.

madand:
Wow, I didn't notice there is already open issue regarding EOLs. I was just about to create a new one when @LynxAbraxas have referenced this one. Here are my findings regarding this so far:

At the moment most of the code files (c,cpp,h) already have Unix (LF) terminators and its nice. But I have located some that use Windows (CRLF) terminators and some that have mixed CRLF and LF in one file.

Files with only CRLF:
```bash
bash> file ctp2_code/**/*.{h,c,cpp} | grep 'CRLF[[:space:]]'
ctp2_code/os/nowin32/windows.h:                                                       C source, ASCII text, with CRLF line terminators
ctp2_code/ctp/ctp2_utils/c3debug.cpp:                                                 C source, UTF-8 Unicode text, with CRLF line terminators
ctp2_code/ctp/ctp2_utils/tracklen.cpp:                                                C source, ASCII text, with CRLF line terminators
ctp2_code/ctp/debugtools/debugcallstack.cpp:                                          C source, UTF-8 Unicode text, with CRLF line terminators
ctp2_code/ctp/debugtools/log.cpp:                                                     C source, ASCII text, with CRLF line terminators
ctp2_code/gs/fileio/CivPaths.cpp:                                                     C source, UTF-8 Unicode text, with CRLF line terminators
ctp2_code/os/nowin32/nowin32.cpp:                                                     C source, ASCII text, with CRLF line terminators
ctp2_code/ui/aui_common/aui_mouse.cpp:                                                C source, ASCII text, with CRLF line terminators
ctp2_code/ui/aui_common/aui_movie.cpp:                                                C source, ISO-8859 text, with CRLF line terminators
ctp2_code/ui/aui_common/aui_textfield.cpp:                                            C source, ASCII text, with CRLF line terminators
ctp2_code/ui/interface/chatbox.cpp:                                                   C source, UTF-8 Unicode text, with CRLF line terminators
ctp2_code/ui/netshell/allinonewindow.cpp:                                             C source, ISO-8859 text, with CRLF line terminators
```

[...]

PR #287, opened 2020-03-25
WIP: stat(2) syscall usage fix and refactoring

madand:
Here is a PR resolving remaining issues with regards to `stat` calls, as [discussed](https://github.com/civctp2/civctp2/pull/283#issuecomment-603572247) previously.

However I cannot finish this PR right away, because of the line terminators issues in some of the files I need to edit for this PR. So I will open a separate PR for resolving these issues, so I would be able to finish this one.

### DIFF
--- a/ctp2_code/gs/fileio/civscenarios.cpp
+++ b/ctp2_code/gs/fileio/civscenarios.cpp
@@ -175,7 +175,7 @@ void CivScenarios::LoadScenarioPackData(ScenarioPack *pack, MBCHAR *packPath)
 #else
 		r = stat(scenListName, &tmpstat);
 #endif
-		if (!r) {
+		if (r == 0) {
 			MBCHAR *scenarioPath = new MBCHAR[strlen(scenPath)+1];
 			strcpy(scenarioPath, scenPath);
 			scenList->AddTail(scenarioPath);
@@ -262,7 +262,7 @@ void CivScenarios::LoadData(void)
 #else
 			r = stat(packListName, &tmpstat);
 #endif
-			if (!r) {
+			if (r == 0) {
 				fileListFileName = new MBCHAR[strlen(name)+1];
 
 				strcpy(fileListFileName, name);
@@ -461,11 +461,14 @@ CIV_SCEN_ERR CivScenarios::MakeNewPack(MBCHAR *dirName, MBCHAR *packName, MBCHAR
 	g_civPaths->GetScenarioRootPath(rootPath);
 
 	sprintf(path, "%s%s%s", rootPath, FILE_SEP, dirName);
+
+	int rc;
 #ifdef WIN32
-	if(!_stat(path, &tmpstat)) {
+	rc = _stat(path, &tmpstat);
 #else
-	if(!stat(path, &tmpstat)) {
+	rc = stat(path, &tmpstat);
 #endif
+	if (rc == 0) {
 		return CIV_SCEN_DIR_EXISTS;
 	}
 
@@ -519,12 +522,14 @@ CIV_SCEN_ERR CivScenarios::MakeNewScenario(ScenarioPack *pack, MBCHAR *scenName,
 
 	MBCHAR scenPath[_MAX_PATH];
 	sprintf(scenPath, "%s%sscen%04d", pack->m_path, FILE_SEP, pack->m_numScenarios);
-#ifdef WIN32
-	if(!_stat(scenPath, &tmpstat)) {
-#else
-	if(!stat(scenPath, &tmpstat)) {
-#endif
 
+	int rc;
+#ifdef WIN32
+	rc = _stat(scenPath, &tmpstat);
+#else
+	rc = stat(scenPath, &tmpstat);
+#endif
+	if (rc == 0) {
 		return CIV_SCEN_DIR_EXISTS;
 	}
 
--- a/ctp2_code/ctp/civ3_main.cpp
+++ b/ctp2_code/ctp/civ3_main.cpp
@@ -317,7 +317,7 @@ namespace Os
 		snprintf(szLink, sizeof(szLink) - 1, "/proc/%ld/exe", static_cast<long int>(pid));
 		int rc = lstat(szLink, &st);
 		// szLink must be a link...
-		if (rc != 0)
+		if (rc == -1)
 			return std::basic_string<TCHAR>();
 
 		if (!S_ISLNK(st.st_mode))
@@ -376,7 +376,7 @@ namespace Os
 #elif defined(LINUX)
 		struct stat st = { 0 };
 		int rc = stat(Os::GetExeName().c_str(), &st);
-		if (rc != 0)
+		if (rc == -1)
 			return exeVersion.str();
 
 		struct tm *t = localtime(&st.st_mtime);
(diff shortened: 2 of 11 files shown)


MartinGuehmann:
What line terminator issues?

ctp2_code/gs/fileio/civscenarios.cpp
```diff
@@ -175,7 +175,7 @@ void CivScenarios::LoadScenarioPackData(ScenarioPack *pack, MBCHAR *packPath)
 #else
 		r = stat(scenListName, &tmpstat);
 #endif
-		if (!r) {
+		if (r == 0) {
```
LynxAbraxas: Does the new logic also fit the return value of `_stat(` from windows?

ctp2_code/gs/fileio/civscenarios.cpp
```diff
@@ -461,11 +461,14 @@ CIV_SCEN_ERR CivScenarios::MakeNewPack(MBCHAR *dirName, MBCHAR *packName, MBCHAR
...
+
+	int rc;
 #ifdef WIN32
-	if(!_stat(path, &tmpstat)) {
+	rc = _stat(path, &tmpstat);
 #else
-	if(!stat(path, &tmpstat)) {
+	rc = stat(path, &tmpstat);
 #endif
+	if (rc == 0) {
```
LynxAbraxas: Same here, does the new logic also fit the return value of `_stat(` from windows?

ctp2_code/gs/fileio/civscenarios.cpp
```diff
@@ -519,12 +522,14 @@ CIV_SCEN_ERR CivScenarios::MakeNewScenario(ScenarioPack *pack, MBCHAR *scenName,
...
 	MBCHAR scenPath[_MAX_PATH];
 	sprintf(scenPath, "%s%sscen%04d", pack->m_path, FILE_SEP, pack->m_numScenarios);
+
+	int rc;
 #ifdef WIN32
-	if(!_stat(scenPath, &tmpstat)) {
+	rc = _stat(path, &tmpstat);
 #else
-	if(!stat(scenPath, &tmpstat)) {
+	rc = stat(path, &tmpstat);
```
LynxAbraxas: Getting here
```
civscenarios.cpp:530:12: error: 'path' was not declared in this scope
rc = stat(path, &tmpstat);
```
Guess it shoudl be `scenPath` instead of `path`

LynxAbraxas:
Some files have the linux and the windows EOLs, i.e. `\n` or `\r\n`. This can be ignored but could possibly be very nicely solved with git's abilities: #228.

ctp2_code/gs/fileio/civscenarios.cpp
madand: Yes.

ctp2_code/gs/fileio/civscenarios.cpp
madand: Yes.

ctp2_code/gs/fileio/civscenarios.cpp
madand: Will fix.

madand:
Reference of Windows `_stat()`: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=vs-2019. 
The semantics of return code are exactly the same as POSIX `stat()`.

madand:
I will be able to finish this PR once #289 gets merged.

LynxAbraxas:
> Reference of Windows `_stat()`: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=vs-2019.
> The semantics of return code are exactly the same as POSIX `stat()`.

That's great (some symmetry at least...) many thanks for checking.

LynxAbraxas:
It does compile with the change mentioned above, but sadly saving/loading maps from the cheat-mode still does not work for me, so that might be a different issue.

madand:
@LynxAbraxas do you mean that saving/loading maps from the cheat-mode worked before, but this PR breaks it? I don't know anything about cheat mode, could you please provide steps to reproduce what doesn't work?

LynxAbraxas:
> do you mean that saving/loading maps from the cheat-mode worked before, but this PR breaks it? 

No, didn't work before but sadly still not with the changes of this PR (as I was hoping).

> I don't know anything about cheat mode, could you please provide steps to reproduce what doesn't work?

Well, when you have the screen for launching a new game there is the button "Launch Editor":
ss_2020-03-27_21:07:05
which basically starts the new game in cheat mode (like with the key '!' during a game but without saving the cheat flag) which allows to edit the map and to construct a new scenario. There is a save/load map button, but I never get a map listed, even if I just save one before:
ss_2020-03-27_21:07:41

madand:
Thanks for explanation, I indeed never tried this stuff. Then I guess this should be opened as a separate issue.

LynxAbraxas:
> Then I guess this should be opened as a separate issue.

Yeah, appears to be more than the `stat ` as was the issue for the scenarios.

> I will be able to finish this PR once #289 gets merged.

How does this PR depend on #289? Why not merge this before?

madand:
The following files need `stat` refactoring, but they are CRLF or mixed at the moment:
```
ctp2_code/gs/fileio/CivPaths.cpp
ctp2_code/ctp/ctp2_utils/c3files.cpp
ctp2_code/ctp/ctp2_utils/tracklen.cpp
```

MartinGuehmann:
> The following files need `stat` refactoring, but they are CRLF or mixed at the moment:
> 
> ```
> ctp2_code/gs/fileio/CivPaths.cpp
> ctp2_code/ctp/ctp2_utils/c3files.cpp
> ctp2_code/ctp/ctp2_utils/tracklen.cpp
> ```

And what is the problem with line endings?

LynxAbraxas:
> The following files need `stat` refactoring, but they are CRLF or mixed at the moment:
> 
> ```
> ctp2_code/gs/fileio/CivPaths.cpp
> ctp2_code/ctp/ctp2_utils/c3files.cpp
> ctp2_code/ctp/ctp2_utils/tracklen.cpp
> ```

That's no problem, I think, though not nice it can stay CRLF or mixed here and be fixed with #289.

madand:
Agree, will commit with existing EOLs. The perfectionist part of myself blinded the pragmatic part in this case :smile: 

I also noticed several calls to `lstat` which has exactly the same return code semantics as `stat`, so I will refactor them as well. Will be ready in a few minutes.

madand:
All changes are ready.

LynxAbraxas:
That's great! Testing if that possibly solve the map issue as well.

>Agree, will commit with existing EOLs. The perfectionist part of myself blinded the pragmatic part in this case smile

Yeah, I know what you mean, it took me some time as well to accept the imperfect, but with the code of CTP2 it is really necessary to accept that, otherwise you would basically have to rewrite the whole code base;-)
It's a project where the best we can do is to improve small parts step by step.

LynxAbraxas:
> Testing if that possibly solve the map issue as well.

Sadly not, so that seems to be a different issue.
All looks good so I'd say good to go in.

MartinGuehmann:
Then let's merge.

Issue #262, opened 2020-02-06
it does segfault on rpi4 (retest)


Askmewho:
```
pi@raspberrypi:~/ctp2/ctp2_program/ctp $ gdb ./ctp2 
GNU gdb (Raspbian 8.2.1-2) 8.2.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "arm-linux-gnueabihf".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./ctp2...(no debugging symbols found)...done.
(gdb) run
Starting program: /home/pi/ctp2/ctp2_program/ctp/ctp2 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
[New Thread 0xb48351f0 (LWP 20205)]
[New Thread 0xb3bba1f0 (LWP 20206)]
[New Thread 0xb329b1f0 (LWP 20207)]
[New Thread 0xb19151f0 (LWP 20214)]
[New Thread 0xb11141f0 (LWP 20215)]
Detected Ogg Music track, using that instead of CD Audio
ui/aui_common/aui_movie.cpp L640: SDL mouse code is missing here!

Thread 1 "ctp2" received signal SIGSEGV, Segmentation fault.
0x00265cf0 in spnewgametribescreen_addAllTribes() ()
(gdb) bt
#0  0x00265cf0 in spnewgametribescreen_addAllTribes() ()
#1  0x00265ca0 in spnewgametribescreen_addAllTribes() ()
#2  0x00264f88 in spnewgametribescreen_Initialize(void (*)(aui_Control*, unsigned int, unsigned int, void*)) ()
#3  0x002620ac in spnewgamescreen_displayMyWindow() ()
#4  0x00214ddc in initialplayscreen_newgamePress(aui_Control*, unsigned int, unsigned int, void*) ()
[...]

Issue #269, opened 2020-02-16
[Linux] Build is sucessfull, but segfault when trying to start a new game


madand:
I build from the latest code `67954be9`. Just tried to build the docker image, got segfault.

![Screenshot_20200217_010919](https://user-images.githubusercontent.com/2280844/74614744-4a459200-5123-11ea-8071-6288d0c52f4f.png)

Notice how the main screen layout looks like in vanilla game, unlike CTP:AE 2011-06-11 I play via WINE. 

Contents of `civ3log000.txt`:
```
  civ3_main.cpp@1348: ** BUILD executible :/opt/ctp2/ctp2_program/ctp/ctp2
   unitutil.cpp@118 : Unit 0: Abolitionist
   .... [omitted boring stuff]
   unitutil.cpp@118 : Unit 73: Elephant Warrior
 SlicEngine.cpp@2082: SLIC: control InitPlayWindow.SPButton used
    c3debug.cpp@303 : Assertion (dataBlock) Failed in File:ui/aui_common/aui_ldl.cpp, Line:559
    c3debug.cpp@304 : Stack Trace: '  0x5c955622  [DebugCallStack_Save + 0x24]
  0x5c955a44  [c3debug_StackTrace() + 0x34]
  0x5c9519e6  [c3debug_Assert(char const*, char const*, int) + 0x87]
  0x5c9ec8ca  [aui_Ldl::BuildHierarchyFromRoot(char const*) + 0x108]
  0x5cb7ffcd  [spnewgametribescreen_Initialize(void (*)(aui_Control*, unsigned int, unsigned int, void*)) + 0x262]
  0x5cb7d5df  [spnewgamescreen_displayMyWindow() + 0x59]
  0x5cb36d26  [initialplayscreen_newgamePress(aui_Control*, unsigned int, unsigned int, void*) + 0x34]
  0x5c9d6546  [aui_Button::MouseLDropInside(aui_MouseEvent*) + 0x1ca]
  0x5cac3092  [ctp2_Button::MouseLDropInside(aui_MouseEvent*) + 0x34]
  0x5ca13e5c  [aui_Region::MouseDispatch(aui_MouseEvent*, unsigned int) + 0x2f4a]
  0x5ca0da90  [aui_Region::HandleMouseEvent(aui_MouseEvent*, unsigned int) + 0x212]
  0x5ca0d99f  [aui_Region::HandleMouseEvent(aui_MouseEvent*, unsigned int) + 0x121]
[...]

PR #288, opened 2020-03-25
Make the game only load the ldl files from the default directory

MartinGuehmann:
With this pull request CTP2 only loads the ldl files from the default directory. However, there is now a new keword "importLoc" to load ldl files from the language specific localized directories.

This way we can have have ldl files in the default folder without caring about any ldl files in the localized folders from the original CD.

This pull request replaces pull request #277 and fixes issues #262 and #269.


LynxAbraxas:
I'll test, has been a long time since I had a look at (f)lex code...

LynxAbraxas:
Yes, seems to work nicely:

ss_2020-04-03_23:27:37

ss_2020-04-03_23:28:38

and looks reasonable, as much as I can remember flex code.
Good to go, I'd say.

MartinGuehmann:
So rebased and fixed the merge conflict, then let's merge.

PR #289, opened 2020-03-26
Normalize end of line terminators

### LINKED ISSUE #228: Convert code from DOS to Unix EOL

madand:
resolves #228

At the moment this PR only addresses `C`, `CPP` and `H` files.  Please note that Github does not show line terminators in its diff output.

But I also found out that there are files with mixed EOLs under the `ctp2_data`:
```bash
bash> file ctp2_data/**/*.* | grep CRLF,
ctp2_data/chinese/gamedata/dip2_str.txt:                            ISO-8859 text, with CRLF, CR line terminators
ctp2_data/chinese/gamedata/info_str.txt:                            ISO-8859 text, with CRLF, CR, LF line terminators
ctp2_data/chinese/gamedata/tut_str.txt:                             ISO-8859 text, with CRLF, CR line terminators
ctp2_data/english/gamedata/info_str.txt:                            ISO-8859 text, with very long lines, with CRLF, CR, LF line terminators
ctp2_data/german/gamedata/info_str.txt:                             ISO-8859 text, with very long lines, with CRLF, LF line terminators
ctp2_data/italian/gamedata/info_str.txt:                            ISO-8859 text, with very long lines, with CRLF, CR, LF line terminators
ctp2_data/japanese/gamedata/add_str.txt:                            Non-ISO extended-ASCII text, with very long lines, with CRLF, NEL line terminators
ctp2_data/japanese/gamedata/ldl_str.txt:                            Non-ISO extended-ASCII text, with CRLF, NEL line terminators
ctp2_data/spanish/gamedata/info_str.txt:                            ISO-8859 text, with very long lines, with CRLF, CR, LF line terminators
```
Hell, there are even MacOS (CR) line terminators in there :)

Also, while all `CDB` files under `ctp2_code/gs/newdb/` have CRLF, there is one file that has LF:
```bash
bash> file **/*.{cdb,ldl} | grep -v terminators
ctp2_code/gs/newdb/Map.cdb:                                                     ISO-8859 text
```

And finally, while most of the `TXT` files have CRLF, there are some that have LF:
```
ctp2_data/default/aidata/AdvanceLists.txt:                                ASCII text
ctp2_data/default/aidata/BuildListSequences.txt:                          ISO-8859 text
ctp2_data/default/aidata/UnitBuildLists.txt:                              ASCII text
ctp2_data/default/gamedata/Advance.txt:                                   ASCII text
ctp2_data/default/gamedata/gamefile.txt:                                  ASCII text

Scenarios/AlexanderTheGreat/scen0000/english/gamedata/scen_str.txt:       ASCII text, with very long lines
Scenarios/AlexanderTheGreat/scen0000/english/gamedata/civ_str.txt:        ISO-8859 text
Scenarios/AlexanderTheGreat/scen0000/english/gamedata/gl_str.txt:         ISO-8859 text
```

So I wonder what do you thing about also normalizing all the `TXT`, `LDL` and `CDB` files, with rules for them to always use CRLF (even on Linux)?


LynxAbraxas:
Many thanks @madand for your work here. 

> So I wonder what do you thing about also normalizing all the TXT, LDL and CDB files, with rules for them to always use CRLF (even on Linux)?

I would like all text file to be unified, but I am not sure which version (CRLF or LF) should be the one that gets actually commited because I have never worked with this feature of git so far because I basically don't do windows;-) Perhaps @MartinGuehmann has some experience with this already and can say which way works best.

madand:
I don't do any programming on Windows either, only some gaming.

My opinion is that all text data files should always use CRLF to be more accessible for potential non-programmers who would like to mod the game. Since most non-techie people are on Windows, line terminators should also be Windows'y ones. And all decent editors on Unix/MacOS should work well with CRLF files, while not all Windows text editors can handle Unix or Mac terminators (I am looking at you Notepad, lol).

The way Git handles this stuff is pretty straightforward. When normalization is on, on commit Git converts files to Unix (LF) format, so internally Git always stores files with LF. When user does `checkout` Git might convert terminators to system-specific ones, according to personal user config or `.gitattributes` file as in our case.

MartinGuehmann:
So that are only 30 source files with mixed line endings. That is not much, I guess they used some Linux tool to strip all the comments.

But then I have never really noticed. There were some problems with the line endings, but well.

*.cbs files are only toughed by programmers, so they are basically source files and should be treated like this. You also missed *.y and *.l files.

However, setup TortoiseGit sot that it won't tough line endings. So I would prefer, that the stuff is not converted on checkout, but only on commit.

Then, we would have the text files in Windows format, but we would have to make sure that they are converted on commit.

LynxAbraxas:
> So I would prefer, that the stuff is not converted on checkout, but only on commit.

Hm, as far as I understand that feature, it converts on checkout, because only then it is definite on which OS you are.

MartinGuehmann:
> > So I would prefer, that the stuff is not converted on checkout, but only on commit.
> 
> Hm, as far as I understand that feature, it converts on checkout, because only then it is definite on which OS you are.

But that is wired, if I check it out out on Windows then I get the native windows line endings, at this then I'll commit, and thus goes into the directory.

Well, at least we can have it so that the line endings are the same across all operating systems. And that is the thing I would go for.

LynxAbraxas:
> But that is wired, if I check it out out on Windows then I get the native windows line endings, at this then I'll commit, and thus goes into the directory.

I'm not sure, but I think it is comparable to a git commit hook that converts the EOLs to the specific one chose for the code base and depending on the checkout OS git converts it back and fro.

madand:
Let me clarify things a bit more. I will illustrate with example lines from `.gitattributes` file for each case. Basically Git can handle EOLs in 3 possible ways:

1. `*.cpp -text`. Text attribute is *unset*.  
Git keeps EOLs as-is (no translation at all). This is the default Git behavior.  
That's how we ended up with files having mixed EOLs.

2. `*.cpp text`. Text attribute is *set*.  
On commit: Git will convert EOLs to `LF`.  
On checkout: Git will convert EOLs from `LF` to the native EOLs of the system you are checking out to (e.g. `CRLF` on Win).  
This is what this PR does for `C`,`CPP` and `H` files. I didn't specify rules for `Y` and `L` files, because I don't know whether they are EOL sensitive.

3. `*.ldl text eol=crlf`. Text attribute is *set*, EOL is forced `CRLF`.  
On commit: Git will convert EOLs to `LF`.  
On checkout: Git will convert EOLs from `LF` to `CRLF` on all systems.  
This is the behavior I believe is appropriate for moddable text-data files, to ensure people using different OSes can share modifications without EOL issues. This is a big change which will effectively require all `TXT`, `LDL` files to be recommitted with `LF`. This will be a separate PR, if you agree to this change at all.

MartinGuehmann:
> 2. `*.cpp text`. Text attribute is _set_.
>    On commit: Git will convert EOLs to `LF`.
>    On checkout: Git will convert EOLs from `LF` to the native EOLs of the system you are checking out to (e.g. `CRLF` on Win).
>    This is what this PR does for `C`,`CPP` and `H` files.

And that's what I don't like. Right now most of the source files have UNIX line endings and Visual Studio copes quite well with it. If the line endings are UNIX then it will add new UNIX line endings if they are Windows then it adds new Windows line endings.

The other thing is that I have my working copy on a NTFS partition, which I access on Linux and used for work on Windows and Linux. I think there was also something where I tested whether the code would compile on both systems without problems. And of course checking is something I prefer to do before committing.

That's why I prefer to have line endings all the same on all systems. And if you want them LF then please on Windows, too. They are supposed to be LF anyway, I guess.

For the text files, I just tested the different source files and Win10's Notepad displays the files, correctly. So I wouldn't worry about changing them to LF.

LynxAbraxas:
Just LF would sure be the simplest solution. As I understand the git feature, it can be configured on a user basis, so if LF is fine for the user who checks out than no config is needed. If a user has problems with just LF git can be configured to convert it to CRLF on checkout and commit back just LF. So how about agreeing on LF for the code base. I'll try to test the CRLF on a windows system, but with time, it's not my home;-)

LynxAbraxas:
>  I didn't specify rules for `Y` and `L` files, because I don't know whether they are EOL sensitive.

They should not. They are yacc (yet another compiler compiler, or gnu-bison files for grammar) and lex ( lexical analyzer, or free lex, flex) files, which get translated into c/c++ code.

madand:
Glad to hear that modern Notepad can handle `LF` :) Enforcing `LF` for all program code files seems reasonable for me as well. Will update PR accordingly and will add rules for `Y` and `L` files.

MartinGuehmann:
What's up with the *.cbs files? I would also consider them as program code.

madand:
Do you mean `CDB` files?

MartinGuehmann:
> Do you mean `CDB` files?

Yes.

madand:
Added commit with `CDB` files.

LynxAbraxas:
Looks good to me. `ctp2_code/libs/.gitattributes` wouldn't be needed at the moment but in case the default in `ctp2_code/.gitattributes` is changed, it is good to have the specification for the libs folder.
I generally like to put the editing commands used into the commit message as a reference, as in e.g. 11df8870ee38a35727443798c1930e01541bab35. This came in handy a few time when some months had passed since then.

madand:
@LynxAbraxas added used shell commands into commit messages as you asked.

LynxAbraxas:
That's great, many thanks @madand. 
So do you intend to work on this PR further or is it finished? In that case just remove the WIP from the title so that @MartinGuehmann can consider it for merging.

madand:
If you confirm that you reject the idea of normalizing and enforcing uniform EOL for text-data files in `Scenarios` and `ctp2_data`, I can indeed remove the WIP prefix. Of course, after resolving merge conflicts.

MartinGuehmann:
> If you confirm that you reject the idea of normalizing and enforcing uniform EOL for text-data files in `Scenarios` and `ctp2_data`, I can indeed remove the WIP prefix. Of course, after resolving merge conflicts.

Of course first resolve the merge conflicts.

However, this is up to you if you want to deal with the text files in this pull request or in another one, as you have suggested.

We haven't decided about the EOLs for text files. Since notepad can handle LF, I think we should also go for LF for text files. Even so I don't like it it so much, because it requires much more stuff to change. But well.

madand:
Conflicts resolved, `WIP: ` prefix removed.

MartinGuehmann:
So everything compiles. The text files are another big chunk, so it makes sense to put them into anther pull request. And the linked issue is about code, anyway.

So let's merge.

Issue #290, opened 2020-03-27
Save-directories not automatically created


williamd67:
I built (Ubuntu 18.04; gcc-7.50; g++ 7.50) the source-code and first encountered the same issues as madand #269 with incorrect GoG-data. After reading the thread I was able to start the game but still got assertions on save-directories. I found two things:
1. _fullpath is functional (not returning NULL) on Windows even if the last part does not exist, while on Linux (canonicalize_file_name) returns NULL when the last part does not exist; this causes some assertions to fire (I read in another thread that one was already silenced); I propose to update the GNU-specific implementation of _fullname so that it in case of initial failure, the dirname of the path is tested in stead of the whole path
2. CivPaths::CreateSaveFolders does not create the subfolders; this is caused by the fact that path is given to the c3files_CreateDirectory method in stead of subFolderPath

I have a fix for both. Please let me know if you want me to create a pull-request for it.

BTW: I like civilization games in general and really appreciate all your work.

MartinGuehmann:
Why do you think that you must ask to create a pull request? It's not how it should behave and thus should be fixed.

LynxAbraxas:
Sure, with a PR it is the most precise way to express what should be changed and easiest to discuss. Please go ahead @williamd67.

PR #291, opened 2020-03-28
allow non-complete paths in _fullname (linux)

williamd67:
To minimize the impact I only do the extra work on splitting the path in dirname and basename when canonicalize_file_name returns NULL. It could also be always done. This partly fixes #290. Another PR will be created for the creation of the save-subdirectories.

PR #292, opened 2020-03-28
Create save sub-directories

williamd67:
Together with PR #291 this resolves issue #290.


LynxAbraxas:
Tested #291, #292 and #293. In general I would recommend to combine the commits of #291 and #292, since together they resovle an issue. It is even fine to add a minor change commit like that of #293 to such a PR. It is generally advisable though to always create a separate commit for minor changes such  that they can be handled separately when something needs to be reverted or cherry-picked.

williamd67:
Closed this PR as it is part of #291

PR #293, opened 2020-03-28
Fix incorrect assertion

williamd67:
The assertion tested that m_resourceList was always NULL which is most likely not the case as the member is initialized in the constructor.


williamd67:
Closed this PR as it is part of #291

PR #294, opened 2020-03-30
Make CTP2 compile on MSVS 2017 again

MartinGuehmann:
This pull request makes CTP2 compile again with Microsoft Visual Studio 2017. Additionally, cleans up some white space.


LynxAbraxas:
Many thanks @MartinGuehmann for correcting. Looks good to me and tested. So good to be merged in my opinion.

For these cases it would be really helpful to have a CI for building on windows. @ptitSeb once had considered to introduce it but I guess forgot about it or lacks time.

MartinGuehmann:
Well, I think there is even an issue about it. Indeed it would be nice to have it. But that would also require me to figure out how to do it.

So let's merge.

MartinGuehmann:
The other thing might be to compile the Linux version with stricter error checking. Unfortunately, we had the problem the diplomacy window.

LynxAbraxas:
> Unfortunately, we had the problem the diplomacy window.

Which problem was that?

MartinGuehmann:
It is covered by pull request #35. If don't use -fpermissive, it gives you much more warnings. I would assume that there should also be some warnings about these implicit const casting stuff. Probably the cookie in these windows should be of type uint32. Or it should be a union type with all the different counts of integers superimposed. But that requires to change a lot of course. I tried once, but gave up since it was too much and it broke too much.

LynxAbraxas:
Ah, I see what you mean. Yes, the compilation with modern compilers does need some trade-offs. I agree, that it would be great to have code that would not need this and compile with much stricter checking, but to get there, it seems a lot of work would be needed that currently could be spent on more pressing matters. The question is, whether getting a windows CI build to work, is faster. This would have the additional benefit that this would yield a ready to use windows executable, that could be packed into a regular install package.

PR #296, opened 2020-03-31
Import english files for MagnificentSamurai and NuclearDetente scenarios

madand:


madand:
According to [StackOverflow](https://stackoverflow.com/questions/3253632/what-are-vssver-scc-files-and-can-i-delete-them), these are some generated crap by Visual SourceSafe or something. Will delete them from the PR.

MartinGuehmann:
> According to [StackOverflow](https://stackoverflow.com/questions/3253632/what-are-vssver-scc-files-and-can-i-delete-them), these are some generated crap by Visual SourceSafe or something. Will delete them from the PR.

Well, it was for a source control system, which even Microsoft didn't use very much, when Microsoft abandoned it. I assume it is not really was to extract any information from it. It might be even scattered anyway.

madand:
Force-pushed data without `SCC` files. 

Please note that there is already one such file in the repo: `Scenarios/AlexanderTheGreat/scen0000/chinese/uidata/layouts/vssver.scc`.

MartinGuehmann:
> Force-pushed data without `SCC` files.
> 
> Please note that there is already one such file in the repo: `Scenarios/AlexanderTheGreat/scen0000/chinese/uidata/layouts/vssver.scc`.

But I guess that is the only one. It would only make sense, if we had all of them and if we could get the information into git, which I guess would be some pain. So I guess that should also go.

LynxAbraxas:
Fantastic! Does make the English version of the Samurai scenario work.

MartinGuehmann:
That we can just merge this.

Issue #42, opened 2018-12-01
mouse pointer back ground not always updated with SDL


LynxAbraxas:
If the back ground changes (e.g. new window, change of map position) it sometimes happens that the background of the mouse pointer still has the old back ground, see image below where the start screen was already cleared but the bg of the mouse pointer is still showing some part of it. I observed this with the master and the linux branch (under linux) and expect that this is an SDL problem.

![ss_2018-12-01_10 12 13](https://user-images.githubusercontent.com/21012234/49326501-59a6e280-f553-11e8-8f2b-4b9478cb3832.png)

LynxAbraxas:
Seems solved by #298, thanks @williamd67!

PR #298, opened 2020-04-06
Migrate to sdl2

williamd67:
Initial migration to get the game working with SDL2 without using SDL2 specific features.


williamd67:
Hi I want to flag this PR as WIP as I saw other PRs with that label. It would allow you to look into the changes that I made. It seems that I cannot add the label myself.

MartinGuehmann:
> Hi I want to flag this PR as WIP as I saw other PRs with that label. It would allow you to look into the changes that I made. It seems that I cannot add the label myself.

You mean that nice red label. I don't know how to make it so that everybody can use it. But you can still rename the title to include WIP.

williamd67:
Yes indeed I liked that nice little label ;). I will change the title and look into the failing check.

williamd67:
Current state is that ctp2 runs with SDL2 with some limitations:
* no movie playback;
* no CD-rom;
* no debug-memory as it causes a segmentation fault during SDL_init;
* no unicode input yet (as far as I can determine); SDL2 uses SDL_TEXTEDITING, and SDL_TEXTINPUT events;
* scrolling with the mouse on the edge of the screen is too fast;
* ...

ctp2_code/ctp/ctp2_utils/c3cmdline.cpp
```diff
@@ -34,8 +34,8 @@
 //
 // - Removed non-standard include file <iostream.h>.
 // - Standardised min/max usage.
-// - Replaced old civilisation database by new one. (Aug 20th 2005 Martin G�hmann)
-// - Replaced old const database by new one. (5-Aug-2007 Martin G�hmann)
+// - Replaced old civilisation database by new one. (Aug 20th 2005 Martin G�hmann)
```
williamd67, line 37: My editor keeps corrupting your name. I will fix it.

MartinGuehmann:
> Yes indeed I liked that nice little label ;). I will change the title.
> 
> The failing check is logical as I need SDL2 libraries now. We probably need to configure a SDL2 specific build.

Yes, but that is part of this pull request. You have to tell Travis what to install, and you tell it in file .travis.yml, which is in the project root directory.

LynxAbraxas:
This is great and looks very promising. I will try to look into this as soon as I find enough time.

> Yes, but that is part of this pull request. You have to tell Travis what to install, and you tell it in file .travis.yml, which is in the project root directory.

Yes, and it is similar to the `.gitlab-ci.yml`. One travis build uses the `Dockerfile` for building
https://github.com/civctp2/civctp2/blob/f8f96b23b0ab7c30ae248bec04bb3ad33f04f6cc/.travis.yml#L42-L45
which probably is the easier way to test your changes because docker gives a well-defined build environment and you only need docker to test it locally with e.g. `docker build -t ctp2 --target builder .` If you want, you can even set up GL-CI to build and test your changes within a well-defined docker environment. You only need to create a private ctp2CD repo with the files from GOG/CD and fork/push civctp2 on GL and pass it the token to access the ctp2CD artifact
https://github.com/civctp2/civctp2/blob/f8f96b23b0ab7c30ae248bec04bb3ad33f04f6cc/.gitlab-ci.yml#L44
needed for the merge stage of the DF
https://github.com/civctp2/civctp2/blob/f8f96b23b0ab7c30ae248bec04bb3ad33f04f6cc/Dockerfile#L100-L101

.travis.yml
```diff
@@ -6,9 +6,10 @@ addons:
     sources:
     - ubuntu-toolchain-r-test
     packages:
-    - libsdl1.2-dev
-    - libsdl-mixer1.2-dev
-    - libsdl-image1.2-dev
+    - libsdl2-dev
+    - libsdl2-mixer-dev
+    - libsdl2-image-dev
+    - libtiff-dev
```
LynxAbraxas, line 12: Do we need libtiff-dev? If so I guess it was pulled in by some former *-dev package.

Dockerfile
```diff
@@ -89,18 +62,16 @@ FROM system as install
...
-## copy libs first since they are less likely to change
-COPY --from=builder /usr/local/lib /usr/local/lib
-ENV LD_LIBRARY_PATH "${LD_LIBRARY_PATH}:/usr/local/lib"
-
 ## ctp2CD/ copy done in install stage such that stages before are compatible with travis docker build, results in one additional layer in the final DI (incr. DI download size)
 COPY ctp2CD/ /opt/ctp2/
 ## ctp2/ copy has to be after ctp2CD/ to overwrite with newer versions from civctp2
 COPY --from=builder /opt/ctp2/ /opt/ctp2/
+## Disable English layout files
+RUN mv /opt/ctp2/ctp2_data/english/uidata/layouts /opt/ctp2/ctp2_data/english/uidata/layouts.org
```
LynxAbraxas, line 74: This is something I do not understand. Why is it needed?

LynxAbraxas:
If you look at the travis builds, you can see that the docker build passes (do you  have a GitLab user account? williamd67 did not match any user name on GL). 
The others fail due to `error: unknown type name ‘_IO_cookie_io_functions_t’`, which needs some investigation.

Dockerfile
williamd67: It is probably temporarily I added this based on #267 which caused crashes during startup of the game. I did not try recently if it is still needed. It moves the original ./ctp2/ctp2_data/english/uidata/layouts out of the way (ctp2 will not be able to read it anymore) without deleting it.

.travis.yml
williamd67: I expect so as I did not change anything related to tiff-files, still the compilation complaint about missing tiffio.h.

williamd67:
No I do not have a GitLab account. I am looking at the issue `error: unknown type name ‘_IO_cookie_io_functions_t’` and have no idea yet. I found an old report that pointed to gawk. I tried it but it was not successful.

LynxAbraxas:
> I found out that using bionic (18.04) as base for the Travis build, solves the error. 

This is great! 

>  I do not know the policy for the Linux version that you want to support. So please advice if this change is okay?

There is no policy for this so far and I would say newest stable would probably be best. Since the DF already uses 18.04
https://github.com/civctp2/civctp2/blob/f8f96b23b0ab7c30ae248bec04bb3ad33f04f6cc/Dockerfile#L4
which is the current stable (https://hub.docker.com/_/ubuntu), I would suggest to change to not using a fixed version but `ubuntu:latest` for travis and the DF. At least for GL it would be possible to create a monthly schedule, that runs a GL-CI pipeline for current ctp2 master (independent of PR merges). The GL-CI is setup such that it checks if a newer base image is available (`--pull`)
 https://github.com/civctp2/civctp2/blob/f8f96b23b0ab7c30ae248bec04bb3ad33f04f6cc/.gitlab-ci.yml#L26
an if so uses that as the base for re-building all following levels/stages. This way we would know if changes to `ubuntu:latest` would lead to failures due to updated/outdated/renamed packages etc.

LynxAbraxas:
> No I do not have a GitLab account. 

If you get one, I can invite you such that you can see the setup of a ctp2CD repo, if you want.

williamd67:
> Would there be an alternative to sdl-ffmpeg for the in-game video playing? Before sdl-ffmpeg I had implemented to call mplayer externally, which would be an alternative, but that is really very hacky (but would work with the original video format).
I am currently investigating movie playback. My current plan is to use ffmpeg and SDL2 directly from the ctp2 code, following [this tutorial](http://www.dranger.com/ffmpeg/tutorial01.html). First I need to figure out how all movie classes in the code are connected to be able to add the new movie playback code in the correct place.

LynxAbraxas:
> I am currently investigating movie playback. My current plan is to use ffmpeg and SDL2 directly from the ctp2 code, following [this tutorial](http://www.dranger.com/ffmpeg/tutorial01.html). 

That's great! The tutorial seems very fitting, also the two following ones (SDL, audio).

> First I need to figure out how all movie classes in the code are connected to be able to add the new movie playback code in the correct place.

Perhaps the changes in the commits 
ce577097c945293ce7993627125d631873757615
47dcad3d17aedd1822c65dcc0232c85ad366cf44
818b8795b6ab559cdbf3619499b418467a72ff7b
can help you there (i.e. search for commits containing `ffmpeg` with `gitk`).

LynxAbraxas:
If possible, it would be best (and probably easiest) to use (a modern) ffmpeg as provided by `ubuntu:latest`/`ubuntu:18.04`: https://packages.ubuntu.com/bionic/ffmpeg

LynxAbraxas:
I skimmed you commits and they look really good. You have done a big job already. 
I wonder though if e8b0a52ecce02f531ea992c8ea5a109bf68c2140 should go into its own PR, since it is not directly connected to SDL2 changes. @MartinGuehmann what do you think?

MartinGuehmann:
> I skimmed you commits and they look really good. You have done a big job already.
> I wonder though if [e8b0a52](https://github.com/civctp2/civctp2/commit/e8b0a52ecce02f531ea992c8ea5a109bf68c2140) should go into its own PR, since it is not directly connected to SDL2 changes. @MartinGuehmann what do you think?

Sure, it is better to have it alone. This way it is better to understand. But it has been created during this PR, and probably also in response to something that only showed up in the context of this PR, otherwise I am sure you would have already noticed. So for now it can stay in this PR, unless it also fixes something in the SDL 1.2 version. Then it would make sense, especially since this PR will properly take some time.

williamd67:
I found the same tutorial with a [more recent implementation](https://github.com/rambodrahmani/ffmpeg-video-player) (ffmpeg and SDL2) which is much easier to migrate.

williamd67:
> Sure, it is better to have it alone. This way it is better to understand. But it has been created during this PR, and probably also in response to something that only showed up in the context of this PR, otherwise I am sure you would have already noticed. So for now it can stay in this PR, unless it also fixes something in the SDL 1.2 version. Then it would make sense, especially since this PR will properly take some time.

In my opinion is it not related to SDL2 or SDL 1.2. I think it is caused by my build-tools. I use gcc (g++) 7.5 combined with the -Og flag. I tested it and none of the other -Ox flags causes this. Now in general I think that references should never be NULL so I would be in favor of this change any way but it is definitely not urgent.

As I have to debug a lot especially the segmentation fault in SDL_init was a tough one, I needed this change.

williamd67:
> I found the same tutorial with a more recent implementation (ffmpeg and SDL2) which is much easier to migrate.

I checked the license of the [more recent implementation](https://github.com/rambodrahmani/ffmpeg-video-player) and it is GNU General Public License v3.0 which I think is too strict for ctp2. There is no license file yet in this repository so I am not 100% sure.

To mitigate this I will look at ffplay.c (part of ffmpeg project and licensed under lgpl v 2.1) and the original tutorial (no license).

LynxAbraxas:
> I checked the license of the more recent implementation and it is GNU General Public License v3.0 which I think is too strict for ctp2.

Not sure. In which points do you think it does not fit? ctp2 is open-source, you just need to purchase extra files to start a game.

LynxAbraxas:
> In my opinion is it not related to SDL2 or SDL 1.2. I think it is caused by my build-tools. I use gcc (g++) 7.5 combined with the -Og flag. I tested it and none of the other -Ox flags causes this. Now in general I think that references should never be NULL so I would be in favor of this change any way but it is definitely not urgent.

That's why I think it should get its own PR, `-Og` should be supported. You can then rebase this PR on that so you would have the support for continuing with debugging. We just have to make sure then that the `-Og`  PR gets merged first.
The GL-CI compiles the debug version without `-O`:
https://github.com/civctp2/civctp2/blob/f8f96b23b0ab7c30ae248bec04bb3ad33f04f6cc/Dockerfile#L73-L74

williamd67:
> Not sure. In which points do you think it does not fit? ctp2 is open-source, you just need to purchase extra files to start a game.

Sorry I was not clear. ctp2 can nicely be delivered with the gpl v3 license. What I meant is that that is not my preference. I prefer lgpl, MIT or Apache as they give the developer a little bit more freedom than gpl v3.

LynxAbraxas:
> > Not sure. In which points do you think it does not fit? ctp2 is open-source, you just need to purchase extra files to start a game.
> 
> Sorry I was not clear. ctp2 can nicely be delivered with the gpl v3 license. What I meant is that that is not my preference. I prefer lgpl, MIT or Apache as they give other developers a little bit more freedom than gpl v3.

Yes, me too, but here it will be more helpful if the final code works with the default ffmpeg from ubuntu, since new users then do not have to compile an ffmpeg in addition. On the other hand, so far no linux packages for ctp2 are generated. However, that could be changed (GL-CI generally offers this option, it would just need configuration for that) and it would be a pity if we had to get into ffmpeg compatibility changes again then.

williamd67:
I had some issues to run ctp2 with the new movie player within the docker-container on my laptop (nvidia).
> libGL error: No matching fbConfigs or visuals found
> libGL error: failed to load driver: swrast
> The program 'ctp2' received an X Window System error.

I fixed it by installing the exact same driver in the container as on my host machine. I added the following command to the Dockerfile:
`RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install nvidia-driver-435 && rm -rf /var/lib/apt/lists/*`
and by adding `--privileged` to the command to run ctp2 in `run.sh`.

LynxAbraxas:
Sounds very exciting, will test this as soon as I find the time and I think I have seen the nvidia swrast error somewhere before, thought I can't remember at the moment where. As far as I can remember it depends on software GL (swrast) or hardware accelerated GL, which is possible to pass to the docker environment with nvidia-docker. Anyway, while the DF can serve for building locally, its main purpose is to create a DI that allows to run and test ctp2 on GL-CI, which does not provide nvidia hardware (so far).

williamd67:
Next step is to use SDL_Cursor to get rid of the cursor-related blitting issues (#42) (only on Linux).

As there exists quiet some logic to keep the background of the cursor correct I also hope to gain some performance.

Dockerfile
LynxAbraxas: Just tested, this causes the GL-CI build to fail because /opt/ctp2/ctp2_data/english/uidata/layouts does not exist there. Please test if it works without these lines for you with current master, i.e. including #288.
You can "amend" this change by doing a `git rebase -i bba27706^' and choosing `e` instead of `pick`, then when rebase stops, just remove the lines and do a `git add -u && git commit --amend` followed by a `git rebase --continue`, then force push.

LynxAbraxas:
> Next step is to use SDL_Cursor to get rid of the cursor-related blitting issues on Linux (#42).
> 
> As there exists quiet some logic to keep the background of the cursor correct I also hope to gain some performance.

That would be cool! I never got the grip on that issue.

LynxAbraxas:
> Sounds very exciting, will test this as soon as I find the time 

Works like a charm, I love it! Great work @williamd67! 
Videos play nicely (even at full screen) and even the sound is fine (which was quite delayed with the SDL_ffpmeg hack). It also loads and renders much faster for me. We might have to consider an increased delay for the sprite animations now, at least they appear very "hasty" to me now;-)
The OGG music does not play for me any more (which it did so far), so #300 might be related to the SDL2 changes.
I generally play from the (GL-CI) DF with something like
```
~/ctp2/run.sh -v ~/civctp2-music/:/opt/ctp2/ctp2_program/ctp/music/:ro -v ~/civctp2-videos/:/opt/ctp2/ctp2_data/default/videos/:ro --env="ALSA_CARD=1" registry.gitlab.com/lynxabraxas/civctp2/migrate-to-sdl2:test-plain ./ctp2
```

> swrast error 

none for me on AMD/Radeon, will try on a PC with Nvidia card.

Dockerfile
LynxAbraxas: Without these lines it compiles all right on GL-CI and even passes all tests. Sadly, some tests still need multiple tries (like e.g. https://gitlab.com/LynxAbraxas/civctp2/-/jobs/511036310), a flaw I was hoping to vanish as well.

LynxAbraxas:
> The OGG music does not play for me any more (which it did so far), so #300 might be related to the SDL2 changes.

Sorry, false alarm. Music also works for me, just was disabled on start up (by default?).

Dockerfile
williamd67: Sorry I should not have committed those lines. Still I will check the master first and if still needed keep the change local.

MartinGuehmann:
> > The OGG music does not play for me any more (which it did so far), so #300 might be related to the SDL2 changes.
> 
> Have to correct that: Music also works for me, but only plays one audio file, then stops until I open the music window and close it again. It then plays another audio file but again stops after that. Could this be related to the mixture of audio of videos and that of OGGs?

That's actually also how it works right now on Windows.

.travis.yml
```diff
@@ -1,16 +1,20 @@
 language: cpp
 sudo: true
-dist: trusty
+dist: bionic
```
LynxAbraxas, line 3: Is there also a dist tag `latest` as used by ubuntu docker pointing to the latest stable?

Dockerfile
```diff
@@ -89,14 +62,10 @@ FROM system as install
 ARG BTYP
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    libsdl1.2debian libsdl-mixer1.2 libsdl-image1.2 libgtk2.0-0 && \
+    libsdl2-2.0 libsdl2-mixer-2.0 libsdl2-image-2.0 ffmpeg libgtk2.0-0 && \
```
LynxAbraxas, line 65: Is the whole ffmpeg package needed or just some of its libraries? If so, I would prefer to name those (even if multiple ones) instead of one big package, to avoid unnecessary bloating the DF size, which needs to be pushed and pull multiple times during the GL-CI build and tests. Even though the net-traffic IO is not limited in GL-CI, the additional time needed for up/down loading is lost from the GL-CI minutes (2000 per month for the free plan).

Dockerfile
williamd67: Checked with the latest master and ctp2 runs without modifications of the file-/directory-names.

Dockerfile
williamd67: I can install the libraries directly which may cause some version mismatch in the future. The libraries are libavcodec57, libavformat57, and libswscale4 which together with there dependencies are still 70Mb where ffmpeg is 130Mb.

In case this will help the automatic build and test system we can implement this as I do not think the versions will change very often and the docker-container is specific for ctp2.

.travis.yml
williamd67: There does not seem to be, reading their documentation there are 4 options for Ubuntu Linux: `dist: bionic` or `dist: xenial` or `dist: trusty` or `dist: precise`

LynxAbraxas: Thanks for checking. Odd they do not have a floating version, but no problem, the DF build will use that and the travis setting will have to be adjusted when needed.

Dockerfile
LynxAbraxas: Well it is nearly only half the size and it would correspond 1:1 to the build deps in the builder stage, so I would go with just the libs and fix the package names in case they change.

williamd67:
As SDL is now responsible for rendering the cursor I expect that #42 is solved for Linux.

Dockerfile
williamd67: Replaced ffmpeg with the needed packages.

williamd67:
Current state is that ctp2 runs with SDL2 with some limitations:
    **done**: ~~no~~ movie playback;
    no CD-rom;
    no debug-memory as it causes a segmentation fault during SDL_init;
    no unicode input yet (as far as I can determine); SDL2 uses SDL_TEXTEDITING, and SDL_TEXTINPUT events which are not handled yet;
    **fixed**: scrolling with the mouse on the edge of the screen is too fast;
    **not related to SDL2 migration**: ~~quitting the game takes a very long time;~~
    ... (all feedback is welcome)

williamd67:
> Music also works for me, but only plays one audio file, then stops until I open the music window and close it again.

This is probably caused by the disabling of the CD-rom code which is a little mixed with ogg-files playback. I will fix it when I process the disabling of the CD-rom code.

LynxAbraxas:
> **not related to SDL2 migration**: ~quitting the game takes a very long time;~

Hm, how long does it take for you? For me it quits in about a second or two.

williamd67:
> Hm, how long does it take for you? For me it quits in about a second or two.

It only happens once in a while and I think it is related to debug builds. For now it is not an issue but I will keep an eye on it if it happens more frequently or consistently.

williamd67:
Current state is that ctp2 runs with SDL2 with some limitations:
**done**: ~~no~~ movie playback;
**removed**: CD-rom support for Linux;
todo: fix debug-memory as it causes a segmentation fault during SDL_init;
todo: movie playback in window (wonder-movie)
**move to separate task**: unicode text input
**fixed**: scrolling with the mouse on the edge of the screen is too fast;
**not related to SDL2 migration**: ~~quitting the game takes a very long time;~~
... (all feedback is welcome)

LynxAbraxas:
https://github.com/civctp2/civctp2/pull/298/commits/3f0c4a152f8565a069bb85f405b5b6619fd07d66 seems to have solved the audio playing issue, thanks.

ctp2_code/ui/aui_common/aui_mouse.cpp
```diff
@@ -110,55 +91,44 @@ AUI_ERRCODE aui_Mouse::InitCommon( void )
...
 	memset( &m_data, 0, sizeof( m_data ) );
 	memset( m_inputs, 0, sizeof( m_inputs ) );
 
 	SetClip( NULL );
+
 #ifdef __AUI_USE_DIRECTX__
 	GetCursorPos( &m_data.position );
-#elif defined(__AUI_USE_SDL__)
-        int x = 0, y = 0;
-        Uint8 state = SDL_GetMouseState(&x, &y);
```
williamd67, line 139: The initialization of those members have been moved to the constructor of aui_SDLMouse.

LynxAbraxas:
The force-push looks fine, I'm testing the new branch state.

williamd67:
> The force-push looks fine, I'm testing the new branch state.

Thanks for testing the changes. Would it be possible to introduce a Windows build in travis so it is guaranteerd at least that it will build.

williamd67:
Current state is that ctp2 runs with SDL2 with some limitations:
**done**: ~~no~~ movie playback;
**removed**: CD-rom support for Linux;
todo: fix debug-memory as it causes a segmentation fault during SDL_init;
**done**: movie playback in window (wonder-movie)
**move to separate task**: unicode text input
**fixed**: scrolling with the mouse on the edge of the screen is too fast;
verify: UI-sound late 
**not related to SDL2 migration**: ~~quitting the game takes a very long time;~~
... (all feedback is welcome)

LynxAbraxas:
> Would it be possible to introduce a Windows build in travis so it is guaranteed at least that it will build?

I'm not sure about the current status and currently do not have the time to investigate this. The travis build was introduced by @ptitSeb and there was the intention to test building on AppVeyor, see e.g.  https://github.com/ptitSeb/ctp2/issues/3#issuecomment-429133884, #6, #294. I have seen some info on GL-CI for windows builds, but as far as I can remember, those need a GL-runner on a windows machine, and I am not sure if such a runner is offered by gitlab.com in the free plan.

MartinGuehmann:
Travis is also supposed to support Windows: https://docs.travis-ci.com/user/reference/windows/

However, I find it a bit messy to get trough this. So in the end I will just try it on Windows with VS2017.

LynxAbraxas:
Thanks @williamd67 for the new commits. I'll test. They might solve the issue I have had in some cases where the back ground of the main mouse pointer was black (instead of being rendered transparent).

williamd67:
Current state is that ctp2 runs with SDL2 ~~with some limitations~~:
**done**: ~~no~~ movie playback;
**removed**: CD-rom support for Linux;
**fixed**: fix debug-memory as it causes a segmentation fault during SDL_init;
**done**: movie playback in window (wonder-movie)
**move to separate task**: unicode text input
**fixed**: scrolling with the mouse on the edge of the screen is too fast;
**improved**: audio latency in UI
**fixed**: fix rendering of battle (only renders when mouse is moving)
**not related to SDL2 migration**: ~~quitting the game takes a very long time;~~

There are no open tasks for this PR any more. Please inform me if you find any issue/glitch/malfunction with this version so I can fix it. I will also test it further.

Let me know what still needs to be done/tested/verified before we can decide to merge this into master.

LynxAbraxas:
Thanks @williamd67 for all your work on this.

> fixed: fix rendering of battle (only renders when mouse is moving)

So far I did not observe this. However, rendering independent of mouse movement might be a good thing, I'm not sure about this at the moment.

williamd67:
Hi @LynxAbraxas, I agree that rendering independent of mouse movements would be a good thing. Unfortunately the drawing and mouse-handling is quiet connected, so I did not dare to touch it is I expected a lot of small side-effects which I cannot oversee by just reading the code.

The commit [Create virtual mouse-events to force a redraw](https://github.com/civctp2/civctp2/pull/298/commits/ee85d1637ee795e5115f65cde71ee7b9fef27ee5) fixed it as I re-introduced virtual mouse events to ensure draw events.

Initially it was caused by this commit [Handle mouse-events in main thread](https://github.com/civctp2/civctp2/pull/298/commits/834003e234e3ed8929d5f9e56d94e268b0c5337c) as I removed the mouse-event thread that injected those virtual mouse events.

To be clear currently the rendering is also done when there is no mouse movement as the issue is fixed.

MartinGuehmann:
> Let me know what still needs to be done/tested/verified before we can decide to merge this into master.

You still have to update the readme. These instructions don't allow me to compile the game: "On debian and friends, use sudo apt install libsdl1.2-dev libsdl-mixer1.2-dev libsdl-image1.2-dev byacc gtk2.0-dev flex to install them all."

Also we should add information about getting the videos from the CD and ripping the CD audio.

williamd67:
> Also we should add information about getting the videos from the CD and ripping the CD audio.

Good point. Unfortunately I neither have a CD nor a CD-rom player so I have to kindly ask some help to write those instructions.

MartinGuehmann:
Well, I have a DVD-Rom drive and two CDs but those are not in Bristol, but in Berlin. So @LynxAbraxas would need us to tell where those are on the CD. For the music you could rip it from youtube as he suggested in #92. But  guess the first track should still have the number 2 as it is with the GOG files. Actually, I still would prefer some ripping instructions for the CD.

williamd67:
Updated readme with SDL2 information. How to rip assets-file from the CD and put them in the working-directories is still missing.

LynxAbraxas:
Hm, I have the CD but no CD-ROM any more... Will try to find a way and report back. As far as I can remember, the CD has an audio part and a data part. The audio part was possible to be ripped with `cdparanoia` into OGGs.

williamd67:
SDL2 has a method SDL_ShowSimpleMessageBox which can be used to show messages such as errors and information.

As far as I can see is gtk only used for such messages, so if the current message-box based on gtk can be replaced with the SDL_ShowSimpleMessageBox we can remove the dependency on gtk.

I will verify if the current functionality can be implemented by this method.

Please let me know if gtk is used in other ways as message-boxes or if it is not meant to be removed.

MartinGuehmann:
> SDL2 has a method SDL_ShowSimpleMessageBox which can be used to show messages such as errors and information.
> 
> As far as I can see is gtk only used for such messages, so if the current message-box based on gtk can be replaced with the SDL_ShowSimpleMessageBox we can remove the dependency on gtk.
> 
> I will verify if the current functionality can be implemented by this method.
> 
> Please let me know if gtk is used in other ways as message-boxes or if it is not meant to be removed.

Not that I know, but I guess you will figure it if it is. :D

But actually, it is just a simple message box or one that can have three buttons? Actually, we need message boxes with one, two, and three buttons.

I think there was something with it that was not so nice. I just don't remember what it was.

LynxAbraxas:
As far as I saw, gtk was only used for those massage boxes. I could imagine that gtk was chosen at a time when SDL integration was not sufficiently working to be use for error messages. It would be great if gkt could be avoided, so please go ahead as you see fit and we can then test with a debug build where many assertions etc happen often.

LynxAbraxas:
> Hm, I have the CD but no CD-ROM any more... Will try to find a way and report back. As far as I can remember, the CD has an audio part and a data part. The audio part was possible to be ripped with `cdparanoia` into OGGs.

Sorry, I'm lacking time and can at the moment only provide the old info of mine which includes the path `Setup/data/Max/ctp2_data/default/videos/` to the video files on my German CD:
https://github.com/civctp2/civctp2/blob/fdcc6083c6caa1aa5edca56fe48d236f226ba787/doc/README.linux#L263

LynxAbraxas:
It seems, since the mouse pointer is drawn by SDL af084f2498662befe6ba86c4929876aa2db46897 (which is great), it does not appear in screen shots any more. Is there a way to somehow include it if it matters (like for #305)?

williamd67:
> It seems, since the mouse pointer is drawn by SDL af084f2 (which is great), it does not appear in screen shots any more. Is there a way to somehow include it if it matters (like for #305)?

I just made the screenshots for the 'Jumping unit-animation' issue and the cursor is shown. I used `alt-PrtScr` on Ubuntu to make those screenshots.

williamd67:
As SDL_WaitEventTimeout is only available in SDL2 I committed this change in this branche.

LynxAbraxas:
> As SDL_WaitEventTimeout is only available in SDL2 I committed this change in this branche.

So that would correspond to the SDL1 compliant additions of #274?

LynxAbraxas:
> > It seems, since the mouse pointer is drawn by SDL [af084f2](https://github.com/civctp2/civctp2/commit/af084f2498662befe6ba86c4929876aa2db46897) (which is great), it does not appear in screen shots any more. Is there a way to somehow include it if it matters (like for #305)?
> 
> I just made the screenshots for the 'Jumping unit-animation' issue and the cursor is shown. I used `alt-PrtScr` on Ubuntu to make those screenshots.
> 
> EDIT: probably there is a way to take a screenshot in the game which I was not aware of. I will check if there is a way to draw the cursor to the surface in that case.

Thanks for replying. I might have an unusual setup (Gentoo with fluxbox) which uses ImageMagick's `import` to take a screen shots with `import -window root `. I just noticed that it included the cursor before and not any more. In general it should not be important (#305 was quite a special case) so no problem.

williamd67:
> So that would correspond to the SDL1 compliant additions of #274?

Indeed it is and when there are no events the application will sleep roughly the same amount of time.

LynxAbraxas:
> > So that would correspond to the SDL1 compliant additions of #274?
> 
> Indeed it is and when there are no events the application will sleep roughly the same amount of time.

That's great. 
@madand how does it work for your setup?

LynxAbraxas:
https://github.com/civctp2/civctp2/pull/298/commits/44885f579717453b0703a5115749705fc099bb09 works well for me, at least for info-messages with an OK-button, also the DI is 32MB smaller. Many thanks, I very much like this PR.

MartinGuehmann:
Visual Studio doesn't like it.

BOOL ShouldTerminateThread(void);

That needs to be public in aui_mouse.h. Just moving it up to "LPCRITICAL_SECTION LPCS( void ) const { return m_lpcs; }" should do it.

And I get this when I quit the game:

```
    c3debug.cpp@304 : Assertion (((aui_DirectSurface *)destSurf)->BUFFER()->IsLost() == DD_OK) Failed in File:c:\activision\civctp2\ctp2_code\ui\aui_directx\aui_directblitter.cpp, Line:27
    c3debug.cpp@305 : Stack Trace: '  0x0149fabc  [char * __cdecl c3debug_StackTrace(void) + 0x3c]
  0x014978b2  [void __cdecl c3debug_Assert(char const *,char const *,int) + 0x72]
  0x0122dbe5  [public: virtual enum AUI_ERRCODE __thiscall aui_DirectBlitter::Blt16To16(class aui_Surface *,struct tagRECT *,class aui_Surface *,struct tagRECT *,unsigned int) + 0x135]
  0x0118a08c  [public: virtual enum AUI_ERRCODE __thiscall C3Blitter::Blt16To16(class aui_Surface *,struct tagRECT *,class aui_Surface *,struct tagRECT *,unsigned int) + 0x7c]
  0x010f9629  [public: virtual enum AUI_ERRCODE __thiscall aui_Blitter::Blt(class aui_Surface *,int,int,class aui_Surface *,struct tagRECT *,unsigned int) + 0x429]
  0x01130b9f  [public: enum AUI_ERRCODE __thiscall aui_UI::BltToSecondary(int,int,class aui_Surface *,struct tagRECT *,unsigned int) + 0x7f]
  0x0113205f  [protected: enum AUI_ERRCODE __thiscall aui_Mouse::Erase(void) + 0xef]
  0x01131de1  [public: enum AUI_ERRCODE __thiscall aui_Mouse::End(void) + 0x161]
  0x0112edd5  [public: virtual __thiscall aui_Mouse::~aui_Mouse(void) + 0x85]
  0x01230ab1  [public: virtual __thiscall aui_DirectMouse::~aui_DirectMouse(void) + 0x81]
  0x01230b1b  [public: void __thiscall aui_DirectMouse::`vbase destructor'(void) + 0x2b]
  0x01230b9b  [public: virtual void * __thiscall aui_DirectMouse::`vector deleting destructor'(unsigned int) + 0x2b]
  0x0117136b  [public: virtual __thiscall aui_UI::~aui_UI(void) + 0x9b]
  0x012382da  [public: virtual __thiscall aui_DirectUI::~aui_DirectUI(void) + 0xfa]
  0x01197e01  [public: virtual __thiscall C3UI::~C3UI(void) + 0x1e1]
  0x0119859b  [public: virtual void * __thiscall C3UI::`vector deleting destructor'(unsigned int) + 0x2b]
  0x0146eb54  [void __cdecl allocated::clear<class C3UI>(class C3UI * &) + 0x54]
  0x01471b64  [private: void __thiscall CivApp::CleanupAppUI(void) + 0x84]
  0x014716b6  [public: void __thiscall CivApp::CleanupApp(void) + 0xa6]
  0x0147afc0  [public: void __thiscall CivApp::QuitGame(void) + 0x40]
  0x01467639  [void __cdecl DoFinalCleanup(int) + 0x69]
  0x014687d0  [long __stdcall WndProc(struct HWND__ *,unsigned int,unsigned int,long) + 0x2d0]
  0x76e0444b  [$R002030 + 0x756200bb]
  0x76de4f8c  [$R002030 + 0x75600bfc]
  0x76de4b2a  [$R002030 + 0x7560079a]
  0x76dee16f  [$R002030 + 0x75609ddf]
  0x76fe427d  [$R002030 + 0x757ffeed]
  0x76dfa4b0  [$R002030 + 0x75616120]
  0x724198eb  [$R002030 + 0x70c3555b]
'
```

It is not so obvious what is wrong here so that we could put it into another pull request. In fact, there were some other asserts on quitting, which are now gone. And quite some memory leaks seem also be gone.

madand:
Just built at d2d03dbe and playtested on my system (it's Arch Linux based, thus latest stable GCC, SDL and other libs). Played for about 30 turns with a new game and a few turns with a pre-existing save. In both cases there were no apparent issues, only good stuff so far:

* Music is working OK.
* Movies are working OK.
* CPU utilization is lower than previously (~40% in comparison to ~50% with the previous loop throttle hack).
* Game loads much faster both at start-up and when loading a save or generating a new map.

The PR feels very solid, good job @williamd67! I will give it a more thorough playtest after finishing my current Wasteland 2 playthrough :)

williamd67:
I removed `WIP` from the title of this PR as I am not actively working on this any more. Only main open issue is: the documentation how to rip assets-files from the CD.

LynxAbraxas:
> Only main open issue is: the documentation how to rip assets-files from the CD.

I'd say we can live without that for the time being. It would be great to have this in. The docs can be adjusted with a new PR when the info or need arises.

MartinGuehmann:
> https://github.com/civctp2/civctp2/blob/fdcc6083c6caa1aa5edca56fe48d236f226ba787/doc/README.linux#L263

I don't understand this line. It seems that it would modify the videos on the CD. But the CD is read only, so that shouldn't work.

Basically, I have now a version of the readme including the information how to get the stuff, as best we know, for cherry-picking, but that still leaves a question mark.

madand:
@MartinGuehmann the re-encoded videos will be output into the *current directory*, not back into CD. The `cd` command on the previous line does exactly that, changes the *current directory*.

https://github.com/civctp2/civctp2/blob/fdcc6083c6caa1aa5edca56fe48d236f226ba787/doc/README.linux#L262-L263

But I believe it would be more correct to put encoded videos under `ctp2_data`: 
```bash
cd /home/foo/CTP2/ctp2_data/default/videos
```

LynxAbraxas:
> But I believe it would be more correct to put encoded videos under ctp2_data:

Yes, I fully agree. Possibly, at that time it might have been the right place for the then experimental search path of ctp2 of the linux branch, can't remember though.

LynxAbraxas:
One thing left: The changes to @MartinGuehmann name clutter this and other PRs and distract from the actual changes made. I would either recommend to remove those from the commits (with e.g. `git filter-branch`) or to add a PR changing that to unicode for all files of the repo and then rebasing all other standing PRs on that.

MartinGuehmann:
@williamd67 if you could cherry-pick ebd724c, then we should have the documentation done. At least as best as possible we can do it right now.

MartinGuehmann:
I get quite some of these assertions:

```
    c3debug.cpp@304 : Assertion (m_primary) Failed in File:ui/aui_sdl/aui_sdlui.cpp, Line:360
    c3debug.cpp@305 : Stack Trace: '  0xdd30aa80  [DebugCallStack_Save + 0xd]
  0xdd30adaa  [c3debug_StackTrace() + 0x52]
  0xdd308e47  [c3debug_Assert(char const*, char const*, int) + 0x79]
  0xdd3b1fd8  [aui_SDLUI::SDLDrawScreen() + 0x9e]
  0xdd39ae4b  [aui_UI::Draw() + 0xb1]
  0xdd493175  [ProgressWindow::StartCountingTo(int, char const*) + 0xa7]
  0xdd493701  [ProgressWindow::BeginProgress(ProgressWindow*&, char*, int, int, int) + 0xf9]
  0xdd3159d7  [CivApp::InitializeApp(hinstance_t*, int) + 0x117]
  0xdd310dd0  [CivMain(int, char**) + 0x230]
  0xdd310a78  [main + 0x26]
  0x9beee1e3  [(unknown) + 0x0]
  0xdd2fccce  [start + 0x2e]
'
```

That's not good.

williamd67:
That is indeed not good.

Do you have a reproduction scenario? Or do you start with specific parameters? I normally start with a combination of: fullscreen, nointromovie, noassertdialogs, eventlog. As this assert never fired for me.

I will have a look at the code and see if I can find a path to SDLDrawScreen while m_primary is not set yet.

LynxAbraxas:
Just checked with the debug version (e.g. DI of a1621e4446fd155c61bf000a861590775a113dd4), but I'm not getting these asserts.
I do though observe the same odd behavior as with SDL1 that under some circumstances a running firefox or thunderbird prevents CPT2 to start properly, basically I only get the window frame but it shows firefox within and never becomes responsive. Closing firefox and then starting CTP2 solves this issue. It is even possible to first start firefox again and then CTP2. I've no idea what is happening there.

MartinGuehmann:
> I do though observe the same odd behavior as with SDL1 that under some circumstances a running firefox or thunderbird prevents CPT2 to start properly, basically I only get the window frame but it shows firefox within and never becomes responsive. Closing firefox and then starting CTP2 solves this issue. It is even possible to first start firefox again and then CTP2. I've no idea what is happening there.

Well, I closed Waterfox Classic, but that had no effect.

> Questions:
> 
> * are you sure that you have a clean build?

I think so, in fact I had to do something to get it to compile.

> * what are your optimization parameters, or maybe asked differently: how do you build the application?

Well, I just used the instructions from the readme:

```
./autogen.sh
CFLAGS="$CFLAGS -w -Og -fno-omit-frame-pointer -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -fpermissive -w -Og -fno-omit-frame-pointer -fuse-ld=gold" ./configure --enable-silent-rules --enable-debug
make
```

In fact, I added these DPRINTFs to the code:

```
AUI_ERRCODE aui_SDLUI::SDLDrawScreen( void ) {
	DPRINTF(k_DBG_FIX, ("m_primary (%x)\n", m_primary));
	Assert(m_primary);
	DPRINTF(k_DBG_FIX, ("m_SDLTexture (%x)\n", m_SDLTexture));
	Assert(m_SDLTexture);
	DPRINTF(k_DBG_FIX, ("m_SDLRenderer (%x)\n", m_SDLRenderer));
	Assert(m_SDLRenderer);
	SDL_UpdateTexture(m_SDLTexture, NULL, m_primary->Buffer(), m_primary->Pitch());
	SDL_RenderClear(m_SDLRenderer);
	SDL_RenderCopy(m_SDLRenderer, m_SDLTexture, NULL, NULL);
	SDL_RenderPresent(m_SDLRenderer);
}

```

And m_primary was not 0. I also outcommented each Assert. When I did the next fired. When I outcommented the last it just crashed.

Looks like some wired optimization stuff so I replaced -Og with -O0 and then it was fine. 

Either we have to add a flag so that gcc does not do this wired optimization with -Og, find that code that should not be optimized, or change the readme so that it says -O0.

williamd67:
> Looks like some wired optimization stuff so I replaced -Og with -O0 and then it was fine.

I had some similar behavior with the CivArchive reference where the compare with NULL was optimized away. I could only reproduce that with -Og and not with any of the other optimization settings. I did not like the reference comparison with NULL and refactored that (see  https://github.com/civctp2/civctp2/commit/e8b0a52ecce02f531ea992c8ea5a109bf68c2140) which at that moment solved that case.

As you encounter the same kind of issue I propose to change the debug optimization to -O0 that is what I currently use and a positive side-effect is that it does not optimize local variables away.

MartinGuehmann:
> > Looks like some wired optimization stuff so I replaced -Og with -O0 and then it was fine.
> 
> I had some similar behavior with the CivArchive reference where the compare with NULL was optimized away. I could only reproduce that with -Og and not with any of the other optimization settings. I did not like the reference comparison with NULL and refactored that (see [e8b0a52](https://github.com/civctp2/civctp2/commit/e8b0a52ecce02f531ea992c8ea5a109bf68c2140)) which at that moment solved that case.

If I look at the original code, I must say it is more complicated than the changed code. Which is itself already a bad thing.
 
> As you encounter the same kind of issue I propose to change the debug optimization to -O0 that is what I currently use and a positive side-effect is that it does not optimize local variables away.

I introduced -Og to accelerating debugging, I think it that this PR makes it faster anyway. And probably I am also using a faster computer. But well, it seems to optimize things it should not tough, which is just buggy behavior. If you just change this information in the readme we should be done, unless @LynxAbraxas still has something.

LynxAbraxas:
The DF uses `-O3` for the non-debug version and no spec in case of the debug version:
https://github.com/civctp2/civctp2/blob/fc6d5cb4309adb6449b8a09b63fba4448fb336ec/Dockerfile#L73-L74
So far I did not encounter any optimization problems with those settings, however I have been wondering if optimization could be the cause for https://github.com/civctp2/civctp2/pull/312#issuecomment-631556937.

MartinGuehmann:
> The DF uses `-O3` for the non-debug version and no spec in case of the debug version:
> https://github.com/civctp2/civctp2/blob/fc6d5cb4309adb6449b8a09b63fba4448fb336ec/Dockerfile#L73-L74
> 

`-O0` is the default value, we just specified it.

> So far I did not encounter any optimization problems with those settings, however I have been wondering if optimization could be the cause for [#312 (comment)](https://github.com/civctp2/civctp2/pull/312#issuecomment-631556937).

Then try it out. ;)

Anyway, I assume we are then happy with this.

MartinGuehmann:
Actually, is there a way to give the message box buttons key board shortcuts? For instance if you get the assert message and you press 'i' it presses the ignore button like it is on Windows.

williamd67:
> Actually, is there a way to give the message box buttons key board shortcuts?

I do not find any way to connect keys to the buttons in SDL_ShowMessageBox  except for `esc` and `return` which I did in the following way (already in the code):

OK-dialog: 'Ok' ~ `return`
OK-Cancel-dialog: 'Ok' ~ `return`, 'Cancel' ~ `esc`
Abort-Retry-Ignore-dialog: 'Abort' ~ `return`, 'Retry' ~ none, 'Abort' ~ `esc`
Yes-No-dialog: 'Yes' ~ `return`, 'No' ~ `esc`
Yes-No-Cancel-dialog: 'Yes' ~ `return`, 'No' ~ none, 'Cancel' ~ `esc`
Retry-Cancel-dialog: 'Retry' ~ `return`, 'Cancel' ~ `esc`
Continue-Try-Cancel-dialog: 'Continue' ~ `return`, 'Try' ~ none, 'Cancel' ~ `esc`

MartinGuehmann:
> > Actually, is there a way to give the message box buttons key board shortcuts?
> 
> I do not find any way to connect keys to the buttons in SDL_ShowMessageBox except for `esc` and `return` which I did in the following way (already in the code):

OK, that's not that obvious, but I guess that's then the best way we can do it.

So finally, we can merge, that was quite long. :partying_face:

LynxAbraxas:
Just noticed that the tool tips appear to be gone with SDL2. Before https://github.com/civctp2/civctp2/commit/28548417da351831daa1732b5ff34863947b73ab just moving the mouse pointer on e.g. one of the action buttons showed a tool tip like pillage here:
ss_2020-06-06_23:23:50
Some were quite helpful, e.g. in the cheat and word building window.
@williamd67 Do you see a chance to get those also with SDL2?

MartinGuehmann:
If they are missing then it is a bug, and should be fixed. Obviously, migrating to SDL2 is complex so that something can be overlooked.

williamd67:
Indeed a regression. Personally I would not miss it as I use the status-bar to see the meaning of the buttons. I will try to fix it (most likely dependent on the mouse-event-handling change).

LynxAbraxas:
Fixed with #317, many thanks @williamd67 for fixing.

LynxAbraxas:
I just noticed that when changing the player in the cheat window, it now sometimes moves in steps of 2, so it can become quite challenging to change to a specific player for debugging. Currently I cannot say which commit introduced this odd behavior. @williamd67 do you have an idea where that might come from?

williamd67:
> do you have an idea where that might come from?

No I do not have an idea. I use this player selection quiet often for debugging, force a specific scenario, as well, and do not encounter this issue. I just tried to reproduce it (on refactor-director branch) and failed.
@LynxAbraxas do you encounter this on a specific branch or on all branches? In case it is on a specific branch I can try to reproduce it.

LynxAbraxas:
> @LynxAbraxas do you encounter this on a specific branch or on all branches? In case it is on a specific branch I can try to reproduce it.

I do not observe this for fc6d5cb4309adb6449b8a09b63fba4448fb336ec (last merge commit on master without SDL2) but for 28548417 (merge of this PR) and following ones, so I would expect that it came from one of the commits of this PR.

williamd67:
As I do not see this behavior and I only use SDL2 branches, we have to find the differences in our configuration.

I use `Ubuntu 18.04 LTS`, with gcc `gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)`. The libraries are standard installed via `sudo apt`.
I also tried the docker-build and also there I could use the spinner with one player up and down.
Just to ensure we talk about the same thing: I use the 'Cheat Mode', after that the 'Empire' tab, and click on the arrows of the player-spinner.

LynxAbraxas:
@williamd67: Do you remember which version of FFmpeg you used for `ffplay.c` as a base for [80965a5](https://github.com/civctp2/civctp2/pull/298/commits/80965a59f2e43bcf8f7da5d30c185ceff53d4f97)

PR #302, opened 2020-04-17
Fixes a division by zero

williamd67:
Player without any units would cause a division by zero in calculation of ratio of done/busy units.

### DIFF
--- a/ctp2_code/ui/aui_ctp2/SelItem.cpp
+++ b/ctp2_code/ui/aui_ctp2/SelItem.cpp
@@ -29,15 +29,15 @@
 //   correct stop player is set.
 // - #01 Standardization of city selection and focus handling
 //   (L. Hirth 6/2004)
-// - Entrenching units are treated like Entrenched units. (Oct 16th 2005 Martin Gühmann)
-// - Added select city instead of army option. (Oct 16th 2005 Martin Gühmann)
+// - Entrenching units are treated like Entrenched units. (Oct 16th 2005 Martin GÃ¼hmann)
+// - Added select city instead of army option. (Oct 16th 2005 Martin GÃ¼hmann)
 // - Added option to avoid an end turn if there are cities with empty build
-//   queues. (Oct. 22nd 2005 Martin Gühmann) Doesn't really work.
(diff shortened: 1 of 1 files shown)


williamd67:
> Did you check if the return values you chose in case there are no units is fitting the logic that calls SelectedItem::UnitsDoneRatio or SelectedItem::UnitsBusyRatio?

Yes in case there are no units... all units are done by default -> 1.0 and UnitsBusyRatio is  the opposite calculation so -> 0.0.
It would have been better (and cheaper) if they would have calculated UnitsBusyRatio by subtracting UnitsDoneRatio from 1.0. Both are only used to draw the progress bar during 'turn end'.

LynxAbraxas:
> It would have been better (and cheaper) if they would have calculated UnitsBusyRatio by subtracting UnitsDoneRatio from 1.0. Both are only used to draw the progress bar during 'turn end'.

Thanks for the pointer, thought I had come across that code once... Now figured, I created it in the scope of #202. How time passes...
Please go ahead and adjust as you see fit, but if I remember correctly `CanAutoSelect` of 
https://github.com/civctp2/civctp2/blob/1fb807386cfe47364d86e70db61635c489f82e12/ctp2_code/ui/aui_ctp2/SelItem.cpp#L671
and `HasContinuingOrder` of 
https://github.com/civctp2/civctp2/blob/1fb807386cfe47364d86e70db61635c489f82e12/ctp2_code/ui/aui_ctp2/SelItem.cpp#L688
have different results.

williamd67:
> It would have been better (and cheaper) if they would have calculated UnitsBusyRatio by subtracting UnitsDoneRatio from 1.0. Both are only used to draw the progress bar during 'turn end'.

Oh, I was too hasty. After reading the thread #202 I realize that there are three groups: not-done, done, and busy of which two have to be calculated. Apologies.

The fix is still valid.

LynxAbraxas:
No problem, looks good to me. @MartinGuehmann ready for merge?

MartinGuehmann:
> No problem, looks good to me. @MartinGuehmann ready for merge?

Looks good, so let's merge.

Issue #111, opened 2019-03-10
Cities can be made visible in unexplored regions with special action mouse pointer


LynxAbraxas:
"Scanning" black regions on the map with a mouse pointer for a special action reveals cities even in unexplored terrain.

PR #305, opened 2020-04-25
Hide unexplored cities for special cursor

williamd67:
It was possible to find cities on unexplored tiles by using an action (cursor) of a special unit.


williamd67:
I expect that this will solve #111 which describes a scenario where actions (cursors) of special units may show unexplored cities.

LynxAbraxas:
Sounds great! I'll check.

LynxAbraxas:
Yes, I can confirm that this PR solves #111. Many thanks @williamd67 for the fix.
@MartinGuehmann What do you think, ready for merge?

MartinGuehmann:
> Yes, I can confirm that this PR solves #111. Many thanks @williamd67 for the fix.
> @MartinGuehmann What do you think, ready for merge?

Looks good. So we can merge.

Issue #303, opened 2020-04-17
Shieldstore of city can become negative


williamd67:
In case a city has an empty build-queue. It does not earn any shields.

The issue is that the number of shields that are subtracted are the 'raw' production of the city while to number of shields that is added is the reduced number of shields after the paid materials. Per example a city earns 20 shields but has to pay 4 for materials. This will mean that in case the queue is empty it will gain 16 but it will loose 20 -> -4 shields are added. In case the city still has shields left it will get a penalty for switching production as well.

There is no protection in the code for a negative shieldstore which causes an assertion when the number of shields will be drawn.

Easiest solution is to prevent the shieldstore to become negative. Still it is not fair that more shields are reduced than earned.

PR #306, opened 2020-04-25
Ensure city's shield-store cannot become negative

williamd67:
Changed functionality of 'removing added shields when the queue is empty' to not 'adding shields in case queue is empty'. The number of shields that were removed was bigger than the number of shields added which may result in negative number of shields in the store. It was bigger as it was not corrected for military and material shields that were already removed.

### DIFF
--- a/ctp2_code/gs/gameobj/bldque.cpp
+++ b/ctp2_code/gs/gameobj/bldque.cpp
@@ -625,8 +625,6 @@ bool BuildQueue::BuildFront(sint32 &shieldstore, CityData *cd, const MapPoint &p
 			// Cities with empty queues lose the same switch penalty % of their shield store
 			// each turn they are empty.
 
-			shieldstore -= cd->GetNetCityProduction();// Remove shields given at beginturn.
-
 			if (shieldstore > 0)
 			{
 				sint32 s = 0;
@@ -643,6 +641,7 @@ bool BuildQueue::BuildFront(sint32 &shieldstore, CityData *cd, const MapPoint &p
 				DPRINTF(k_DBG_GAMESTATE, ("Deducting %i shields for empty queue in city of %lx\n", s, m_city.m_id));
 				shieldstore = s;
 			}
+			assert(shieldstore >= 0);
 		}
 		return false;
 	}
--- a/ctp2_code/gs/gameobj/bldque.cpp
+++ b/ctp2_code/gs/gameobj/bldque.cpp
@@ -616,33 +616,19 @@ bool BuildQueue::BuildFront(sint32 &shieldstore, CityData *cd, const MapPoint &p
 
 		return true;
 	} else {
-
-		DPRINTF(k_DBG_GAMESTATE, ("BuildFront: City %lx building nothing\n",
-								  m_city.m_id));
-
-		if (!Player::IsThisPlayerARobot(m_owner) ||
-			(g_network.IsClient() && g_network.IsLocalPlayer(m_owner))) {
-			// Cities with empty queues lose the same switch penalty % of their shield store
-			// each turn they are empty.
-
-			if (shieldstore > 0)
-			{
-				sint32 s = 0;
-				double penalty =
+		// Cities with empty queues lose the same switch penalty % of their shield store each turn they are empty.
+		DPRINTF(k_DBG_GAMESTATE, ("BuildFront: City %lx building nothing\n", m_city.m_id));
+		if (shieldstore > 0)
+		{
+			double penalty =
 					static_cast<double>(g_theConstDB->Get(0)->GetChangeCurrentlyBuildingItemPenalty());
+			penalty = std::min<double>(1.0, std::max<double>(0.0, 1.0 - penalty * 0.01));
+			sint32 s = static_cast<sint32>(static_cast<double>(shieldstore) * penalty);
+			DPRINTF(k_DBG_GAMESTATE, ("Decreased shields to %d for empty queue in city of %lx\n", s, m_city.m_id));
 
-				penalty = std::min<double>
-								  (
-								   1.0,
-								   std::max<double>(0.0, 1.0 - penalty * 0.01)
-								  );
-
-				s = static_cast<sint32>(static_cast<double>(shieldstore) * penalty);
-				DPRINTF(k_DBG_GAMESTATE, ("Deducting %i shields for empty queue in city of %lx\n", s, m_city.m_id));
-				shieldstore = s;
-			}
-			Assert(shieldstore >= 0);
+			shieldstore = s;
 		}
+		Assert(shieldstore >= 0);
 		return false;
 	}
 }
--- a/ctp2_code/gs/gameobj/bldque.cpp
+++ b/ctp2_code/gs/gameobj/bldque.cpp
@@ -641,7 +641,7 @@ bool BuildQueue::BuildFront(sint32 &shieldstore, CityData *cd, const MapPoint &p
 				DPRINTF(k_DBG_GAMESTATE, ("Deducting %i shields for empty queue in city of %lx\n", s, m_city.m_id));
 				shieldstore = s;
 			}
-			assert(shieldstore >= 0);
+			Assert(shieldstore >= 0);
 		}
 		return false;
 	}
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -2373,7 +2373,10 @@ void CityData::AddShieldsToBuilding()
 		m_net_production = 0;
 	}
 
-	SetShieldstore(m_shieldstore + m_net_production);
+	// Do not add shields to the store when the build-queue is empty.
+	if (m_build_queue.GetHead() != NULL) {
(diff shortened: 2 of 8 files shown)


williamd67:
I expect this to solve #303

LynxAbraxas:
Only had the time to test this indirectly, i.e. found no negative side effects;-)

ctp2_code/gs/gameobj/bldque.cpp
```diff
@@ -643,6 +641,7 @@ bool BuildQueue::BuildFront(sint32 &shieldstore, CityData *cd, const MapPoint &p
 				DPRINTF(k_DBG_GAMESTATE, ("Deducting %i shields for empty queue in city of %lx\n", s, m_city.m_id));
 				shieldstore = s;
 			}
+			assert(shieldstore >= 0);
```
MartinGuehmann, line 644: This should be Assert with capital A. That's the one we use here in the code. It raises that assert message box and makes also sure that the assert is logged in the debug logs with the stack trace.

MartinGuehmann:
I think, I have to look at it a bit more carefully, to check whether there is a point to give add the production at the beginning of the turn.

ctp2_code/gs/gameobj/bldque.cpp
williamd67: Good point... assert (lowercase) is still used in a couple of other places. I will fix them all at once.

williamd67:
> I think, I have to look at it a bit more carefully, to check whether there is a point to give add the production at the beginning of the turn.

Is that to make a more fundamental change?

FYI: this change only prevent adding shields in case the build-queue is empty to not to have them  removed later. It did not change the location of adding shields.

MartinGuehmann:
> > I think, I have to look at it a bit more carefully, to check whether there is a point to give add the production at the beginning of the turn.
> 
> Is that to make a more fundamental change?

I wasn't sure. I thought that empty build queues were supposed to accumulate production, even so at a reduced rate, but that is not the case so that is fine. I would say just some cleanup of Activision's code while you are already at it:

You could also remove the condition that the code is only executed for the human player. Even so practically it won't make a difference since the AI fills it build queues.

And could you also adjust the debug message, it claims to give the production that is removed, but in fact it gives the production that is remaining.

And since you are already at it, the variable 's' is declared and zeroed at the top of that code block, even so it is not used with that value. In fact, it is later overwritten with its final value, even so it could be declared with its final value at that line. That is something Activision likes to do, but it just adds lines without value.

williamd67:
> And since you are already at it, the variable 's' is declared and zeroed at the top of that code block, even so it is not used with that value. In fact, it is later overwritten with its final value, even so it could be declared with its final value at that line. That is something Activision likes to do, but it just adds lines without value.

General question: should we cleanup code when we touch some areas? As till now I made the smallest change possible (except the refactoring of Directory where the code is completely 'shuffled'). While I see a lot of small technical issues that could be improved, such as empty lines, variables declared before usage, even testing inside a loop without any relationship with the loop so performance gain to do it outside the loop, etc.

MartinGuehmann:
> > And since you are already at it, the variable 's' is declared and zeroed at the top of that code block, even so it is not used with that value. In fact, it is later overwritten with its final value, even so it could be declared with its final value at that line. That is something Activision likes to do, but it just adds lines without value.
> 
> General question: should we cleanup code when we touch some areas? As till now I made the smallest change possible (except the refactoring of Director where the code is completely 'shuffled'). While I see a lot of small technical issues that could be improved, such as empty lines, variables declared before usage, even testing inside a loop without any relationship with the loop so performance gain to do it outside the loop, etc.

I would say, yes. When else would you clean it up? Probably, never, unless you go for some big refactoring.

Usually, I address also these little issues when I am toughing code. Actually, I also end up modifying code that is quite unrelated or far away from the actual code I need to change. That is because I just looked at it and thought that this can be fixed or improved. If this code is even in unrelated source files then these changes could be committed first without any extra works.

Usually, these things are also minor to fix and since you remember these easily, it means they also bother you and distract you from the code, even so if it is just slightly.

Sometimes I also clean up white space before any edit, simply because the code is then easier to read and understand.

LynxAbraxas:
>General question: should we cleanup code when we touch some areas? As till now I made the smallest change possible (except the refactoring of Director where the code is completely 'shuffled').

Sure, just put it in its own commit as much as it is possible. Combining commits later with a `git rebase -i` is hardly any problem while splitting a commit can be quite tedious work. If you later find that you started to follow another topic with those commits than with those associated with the PR, you can `git cherry-pick` them into a new branch and create a PR with that. If these commits do not really make up a topic, just leave them in the original branch/PR and distinguish their content in the commit message. I does not always have to the the smallest change possible, it just is very helpful for others if the commit message describes the idea behind the changes as good as possible.

In general, before ideas for improvement get forgotten or lost, it is better to have the changes/improvements in even if not in a perfect manner.

>Sometimes I also clean up white space before any edit, simply because the code is then easier to read and understand.

Yes, me too. I would recommend to make that its own commit and check if no change except white space is shown with e.g. gitk using word-diff and ignoring space change.

MartinGuehmann:
> it just is very helpful for others if the commit message describes the idea behind the changes as good as possible.

Talking about this: "Fixed review comments" is not the best commit message. Even so it where about addressing my comments, it rather suggests that you have edited my comments for spellings, grammar, style, etc.

The idea behind that edit was rather that the shieldstore penalty for empty build queues is also applied to the AI. Just keep in mind that in the I will have only the commit message to tell me what the change was about. This discussion here will be quite remote. In principle, you could also think about scenarios where GitHub just disappears or chances its webpage in a way that these discussions are not even accessible anymore.

LynxAbraxas:
@DEVoytas  once posted these as a pointer:
https://chris.beams.io/posts/git-commit/
https://github.com/erlang/otp/wiki/writing-good-commit-messages
though I must admit that it can be quite hard to meet those standards and it can take longer to write the commit massage than it took to do the actual code change;-)

MartinGuehmann:
> @DEVoytas once posted these as a pointer:
> https://chris.beams.io/posts/git-commit/
> https://github.com/erlang/otp/wiki/writing-good-commit-messages
> though I must admit that it can be quite hard to meet those standards and it can take longer to write the commit massage than it took to do the actual code change;-)

Right, and therefore we should not worry if we do not manage, after all we do this in our free time and aren't paid for doing it.

williamd67:
> Talking about this: "Fixed review comments" is not the best commit message.

Good point and I know. To be honest I did not know exactly what to put there and was expecting a reply that I should rebase and collect the different commits in a single one but was also not 100% sure so I awaited the reaction.

BTW: I like and try to use this rule for commits:
`If applied, this commit will <<your subject line here>>`

MartinGuehmann:
> > Talking about this: "Fixed review comments" is not the best commit message.
> 
> Good point and I know. To be honest I did not know exactly what to put there and was expecting a reply that I should rebase and collect the different commits in a single one but was also not 100% sure so I awaited the reaction.

Well, if at all then only the first and third commit. The assert -> Assert commit could even be a PR of its own, but well, it came out of this PR, so I don't mind to have it here. The third commit may be combined with the first, but to remove an AI cheat is worth to have it on its own. Obviously, it is a bit tricky to find here the right balance. And keep in mind none of us is paid here, so we can't demand a highly polished history and demand that every commit compiles, of course we can demand that every merge commit compiles at least on Linux. And we can demand that the history isn't cluttered too much.

> BTW: I like and try to use this rule for commits:
> `If applied, this commit will <<your subject line here>>`

The first commit in this PR is in this format:

If applied, this commit will
Ensure city's shield-store cannot become negative

On the other hand "Ensure" itself is an imperative and so the line makes perfect sense on its own. Actually, I also think that this is a good commit message.

MartinGuehmann:
I think you should also fix the commit message of the last commit, something like:

Apply the shieldstore penalty for empty build queues to the AI too

Actually, that should be quite simple to do:

https://help.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

LynxAbraxas:
> If applied, this commit will

I would not put that into the commit message title (first line) because:
1. many commits would start with the very same phase
2. the title should be as short as possible and as meaningful as possible

Anyway if a commit is applied or not is basically discussed in the PR and _done_ when the PR is merged into master.

The message of https://github.com/civctp2/civctp2/pull/306/commits/93eabefe086351776293ca91547a7238d20eefe8 is a good example for a good commit message.

Changing a commit message before the PR is merged is easy. If it is the last commit just use `git commit --amend` if it is one or more before use `git rebase -i` and `r` (reword) for those commits, then just do a force push.

williamd67:
> I would not put that into the commit message title (first line) because:

I totally agree. It is an imaginary addition that you put before your `<<your subject line here>>` so that you can verify if the commit-message is saying the right thing in the right syntax.

williamd67:
Done. Thanks, @MartinGuehmann for the suggested commit-message.

MartinGuehmann:
If you could clean up the merge conflicts then we should be done here, too.

MartinGuehmann:
The merge conflicts have been fixed, so we can merge.

Issue #307, opened 2020-04-26
Incorrect unit-animations


williamd67:
Sometimes unit animations start to jump forward and then continue their normal animation.

This is caused by the fact that the unit-animations are started from two different places (HandleNextAction) which sometimes invalidate the correct sequence.

williamd67:
I found two other cases where the unit-animation shows some glitches. In both cases (boat and soldier) does an unit disappear (see images).
![boat-shown](https://user-images.githubusercontent.com/10560119/80620204-f8d05200-8a45-11ea-89e8-b474a6c59b7f.jpg)
![boat-hidden](https://user-images.githubusercontent.com/10560119/80620219-fec63300-8a45-11ea-9927-d1fe20146b27.jpg)
![soldier-shown](https://user-images.githubusercontent.com/10560119/80620239-0554aa80-8a46-11ea-9996-7d9b5c6a85e8.jpg)
![soldier-hidden](https://user-images.githubusercontent.com/10560119/80620249-084f9b00-8a46-11ea-908f-06c78643f1ff.jpg)

williamd67:
I found out that during end-of-turn (when show enemy moves is enabled) there is a split in visual (UnitActor) and real (UnitData) position of the unit which are not aligned at that time. I assume that this causes the above mentioned glitches.

This may also cause the issue mentioned in #260 `Unit/army stays invisible even if in vision of stealth units` as the UnitData-position of those units is correct but the UnitActor-position is incorrect and the reason that they are not drawn.

I do not see an easy solution yet. 

Although I think the solution for the jumping unit is a good solution, together with the above described issue it has a side-effect that during some frames the unit is not rendered which causes a flicker (empty frame) during the unit animation. I propose not to merge this PR till the above mentioned issue has been solved as well.

LynxAbraxas:
Hm, possibly this is even connected to #210, and that stealth units, which get revealed due to bumping into them, appear at the wrong position (I thought that we have an issue already for that but can't find it).

> I propose not to merge this PR till the above mentioned issue has been solved as well.

Just put WIP: in front of the subject/title, that is the convention to mark it that it is not yet ready for merge but open for discussion.

LynxAbraxas:
> I found three other cases where the unit-animation shows some glitches. In two cases, a boat (see image 1 and 2) and a soldier (see image 3 and 4) does a unit disappear and in one case a unit appears: settler (see image 3 and 4).

[...]

PR #308, opened 2020-04-26
WIP: Incorrect unit animations

williamd67:
I expect that this fixes issue #307 'Jumping unit animations'.

### DIFF
--- a/ctp2_code/ctp/civ3_main.cpp
+++ b/ctp2_code/ctp/civ3_main.cpp
@@ -45,17 +45,17 @@
 // - Merged GNU and MSVC code (DoFinalCleanup, CivMain).
 // - Option added to include multiple data directories.
 // - Display the main thread function name in the debugger.
-// - Removed references to CivilisationDB. (Aug 20th 2005 Martin G?hmann)
-// - Removed references to old SpriteStateDBs. (Aug 29th 2005 Martin G?hmann)
-// - Initialized local variables. (Sep 9th 2005 Martin G?hmann)
-// - Removed unused local variables. (Sep 9th 2005 Martin G?hmann)
-// - Removed some unreachable code. (Sep 9th 2005 Martin G?hmann)
+// - Removed references to CivilisationDB. (Aug 20th 2005 Martin Gühmann)
+// - Removed references to old SpriteStateDBs. (Aug 29th 2005 Martin Gühmann)
+// - Initialized local variables. (Sep 9th 2005 Martin Gühmann)
+// - Removed unused local variables. (Sep 9th 2005 Martin Gühmann)
+// - Removed some unreachable code. (Sep 9th 2005 Martin Gühmann)
 // - Moved debug tools handling to c3.h, so that the leak reporter doesn't
-//   report leaks that aren't leaks. (Oct 3rd 2005 Matzin G?hmann)
+//   report leaks that aren't leaks. (Oct 3rd 2005 Matzin Gühmann)
 // - Added version to crash.txt
-// - USE_LOGGING now works in a final version. (30-Jun-2008 Martin G?hmann)
+// - USE_LOGGING now works in a final version. (30-Jun-2008 Martin Gühmann)
 // - The log files are now only opened and closed once, this speeds up
-//   debugging significantly. (09-Aug-2008 Martin G?hmann)
+//   debugging significantly. (09-Aug-2008 Martin Gühmann)
 //
 //----------------------------------------------------------------------------
 //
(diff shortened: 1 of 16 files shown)


LynxAbraxas:
Yeah, I have observed that too sometimes (not sure under which conditions it happens for me) and it is very obvious in a screen cast of @Askmewho: https://github.com/civctp2/civctp2/issues/262#issuecomment-592913847.
Will try to test this.

LynxAbraxas:
Only very rarely observed this effect on my PC (even though it is quite old) so cannot really say if it got resolved but tested the game together with this PR and could not observe any unintended side effects, so I've no objections concerning a merge.

williamd67:
The Director as interface (abstract class) has been introduced to get grip on the animation-issues. I kept the changes to other classes than Director as minimal as possible. I prefer to hide implementation details and decouple as much as possible. This will allow local changes without impacting other classes. 

@MartinGuehmann, @LynxAbraxas what do you think of this change?

williamd67:
I expect that the latest commit fixes the (dis-)appearing units when enemy moves are shown.

I still see one (or more) units where only the **shield** is animated/moved. Is this expected behavior, per example for stealth units or is this another glitch I can look into? When stealth units are hidden, do you see something like the shield/flag or are they really unknown for the viewer?

This does not fix the issue that UnitData and UnitActor are misaligned. I will create a separate pull-request for that.

LynxAbraxas:
Will test this new commits. I noticed with the old one that the enemy moves were only giving sound for the move from the first tile but not from following ones.

williamd67:
I noticed after slowing down the animations that there is a clear glitch where the facing of the unit is incorrect. I will look into that.

LynxAbraxas:
> Will test this new commits. I noticed with the old one that the enemy moves were only giving sound for the move from the first tile but not from following ones.

This is still present with the new commits. Every odd unit move seems to have sound and every even seems to have none.

> I noticed after slowing down the animations that there is a clear glitch where the facing of the unit is incorrect.

I can remember @MartinGuehmann mentioning that the sprites seem to not keep their facing direction, cannot remember in which context that was though.

williamd67:
Found the following issues with unit-animations:
* no sound played for units that ends in transporter
* unit sound does not stop and plays through other unit sounds
* televangelist not facing city when executing special attack
* unit animation jumping

LynxAbraxas:
Will test the new commits. There was one before you force pushed where units were returning to their default position after every move, which is now gone (which is good).

williamd67:
I am currently refactoring Director and friends. The aim is to make the code more straight-forward and make dependencies one-way.

LynxAbraxas:
> I am currently refactoring Director and friends. The aim is to make the code more straight-forward and make dependencies one-way.

That's great. I guess you mean your williamd67/refactor-director branch, which has no PR (yet), since it is based on #298 and this PR, which need to be merged first. When it becomes a PR I would recommend a rebase before its final merge to avoid the multiple merge commits of williamd67/sdl2-master.

williamd67:
> When it becomes a PR I would recommend a rebase before its final merge...

I will definitely do that and will rebase with the master at that moment in time.

I have a general request, do any of you have some save files that are far in the game which I will use to run a turn and save before the refactoring and after so that I can verify that the functionality of the code did not change. Thanks.

LynxAbraxas:
> some save files that are far in the game 

Sure: 
[close2GaiaControllerOnline.zip](https://github.com/civctp2/civctp2/files/4652857/close2GaiaControllerOnline.zip)

williamd67:
When using the save-games to compare the results before and after my changes I saw a strange difference between doing an end-turn with show-enemy-moves and without.
Test-case:
1. load a previously stored game
2. press end-turn
3. wait till it is done
4. exit CTP2
5. use the automatically generated AUTOSAVE for inspection.

The save-game contains two differences when I do this with and without show-enemy-moves enabled:
1. the facing of the actors differs (which may be caused by the animation of the actors);
2. city walls are all unset (0) in the save-game of 'show-enemy-moves' disabled while they are set (1) for the save-game of 'show-enemy-moves' enabled'.

My first guess is that the render code 'fixes' the state of the city-walls. As there is in the RepaintLayerSprites a call to SetHasCityWalls. This seems a patch to me.

I have no idea if this causes any issues still it seems strange.

BTW: I expected that running 'end-turn' with and without show-enemy-moves should result in the same end-state.

LynxAbraxas:
Just as a ref concerning city wall drawing: #244.

williamd67:
This PR is replaced with #314.

PR #310, opened 2020-05-16
Remove uninitialized data from save-file

williamd67:
This will ensure that save-files are binary equal and can be used for a simple regression-test. Equal start point and actions will create a binary equal save-file.


LynxAbraxas:
Thanks for this PR @williamd67. Makes sens to me, but should definitely be looked at by @MartinGuehmann for he knows the save game format at the very best.

MartinGuehmann:
> Thanks for this PR @williamd67. Makes sens to me, but should definitely be looked at by @MartinGuehmann for he knows the save game format at the very best.

It doesn't change the save game format in any way. It just changes the data that is saved. Which is in that case irrelevant for the game.

Of course, it is good practice to zero the array anyway.

I am just surprised that there not more those uninitialized things.

MartinGuehmann:
I guess then we are happy with this. So let's merge.

PR #175, opened 2019-07-22
WIP: Allow AI to break no-piracy agreements

LynxAbraxas:
This PR allows the AI to break no-piracy agreements.

PR #311, opened 2020-05-16
For pull fixes only

henshall828:
Fix a number of issues that were causing compilation or crash problems for me.
The AI changes seem to make sense, but there was some guess-work about intent involved.

This series contains the following changes:

    - fix some uses of delete where delete[] is required

    - fix a compile error on Ubuntu 20.04 caused by comparing an int8 to 256

    - fix a buggy implementation of Match_Resources when removing goals

    - fix use of this != something or this != something else, which will always be true. && was presumably intended

    - fix some assert statements that were performing an assign of zero to size

    - fix some DEBUG print statements that would crash when there was no best unit

### DIFF
--- a/ctp2_code/ai/diplomacy/nproposalevent.cpp
+++ b/ctp2_code/ai/diplomacy/nproposalevent.cpp
@@ -713,7 +713,7 @@ STDEHANDLER(TradePact_NewProposalEvent)
 	if (!args->GetInt(0, desireType))
 		return GEV_HD_Continue;
 
-	if ((MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_MAKE_FRIEND ||
+	if ((MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_MAKE_FRIEND &&
 		(MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_GOLD)
 		return GEV_HD_Continue;
 
@@ -870,7 +870,7 @@ STDEHANDLER(MilitaryPact_NewProposalEvent)
 		if (!args->GetInt(0, desireType))
 			return GEV_HD_Continue;
 
-		if ((MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_MAKE_FRIEND ||
+		if ((MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_MAKE_FRIEND &&
 			(MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_ENLIST_FRIEND)
 			return GEV_HD_Continue;
 	}
--- a/ctp2_code/ai/strategy/scheduler/scheduler.cpp
+++ b/ctp2_code/ai/strategy/scheduler/scheduler.cpp
@@ -632,12 +632,9 @@ void Scheduler::Match_Resources(const bool move_armies)
 	sint32 committed_agents = 0;
 	sint32 total_agents     = m_agents.size();
 
-	for
-	(
-	    Goal_List::iterator goal_iter  = m_goals.begin();
-	                        goal_iter != m_goals.end();
-	                      ++goal_iter
-	)
+	//It is best not to iterate using a for loop, as items may be erased as we go through
+	Goal_List::iterator goal_iter = m_goals.begin();
+	while(goal_iter != m_goals.end())
 	{
 		if(committed_agents >= total_agents)
 		{
@@ -686,7 +683,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 
 			// Actually should be checked in the next cycle, but there still seems to be something wrong.
 			Goal_List::iterator tmp_goal_iter = goal_iter;
-			--goal_iter;
+			goal_iter++;
 			m_goals.erase(tmp_goal_iter);
 
 			/*
@@ -723,6 +720,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 					--goal_iter;
 					goal_ptr->Rollback_All_Agents(); // Just roll back but don't report to the build list
 					Reprioritize_Goal(tmp_goal_iter);
+					goal_iter++;
 					continue;
 				}
 			}
@@ -745,6 +743,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 					("\t\tGOAL_FAILED Not enough transporters (goal: %x)\n", goal_ptr));
 
 				Rollback_Matches_For_Goal(goal_ptr);
+				goal_iter++;
 				continue;
 			}
 		}
@@ -756,6 +755,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 			AI_DPRINTF(k_DBG_SCHEDULER, m_playerId, goal_ptr->Get_Goal_Type(), -1,
 					("\t\tGOAL (goal: %x) -- No agents were committed, maybe next time. Continuing...\n",
 						goal_ptr));
+			goal_iter++;
 			continue;
 		}
 
@@ -865,6 +865,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 				break;
 			}
 		}
+		goal_iter++;
 	}
 
 #if defined(_DEBUG)
--- a/ctp2_code/ai/strategy/scheduler/scheduler.cpp
+++ b/ctp2_code/ai/strategy/scheduler/scheduler.cpp
@@ -632,10 +632,18 @@ void Scheduler::Match_Resources(const bool move_armies)
 	sint32 committed_agents = 0;
 	sint32 total_agents     = m_agents.size();
 
+#if defined(_DEBUG)
+	int loopCount = 0;
+	int loopCountBadUtility = -1;
+#endif
+
 	//It is best not to iterate using a for loop, as items may be erased as we go through
 	Goal_List::iterator goal_iter = m_goals.begin();
 	while(goal_iter != m_goals.end())
 	{
+#if defined(_DEBUG)
+		loopCount++;
+#endif
 		if(committed_agents >= total_agents)
 		{
 			Assert(committed_agents == total_agents);
@@ -664,6 +672,9 @@ void Scheduler::Match_Resources(const bool move_armies)
 
 			AI_DPRINTF(k_DBG_SCHEDULER, m_playerId, goal_ptr->Get_Goal_Type(), -1, ("\n"));
 			AI_DPRINTF(k_DBG_SCHEDULER, m_playerId, goal_ptr->Get_Goal_Type(), -1, ("\n"));
+#if defined(_DEBUG)
+			loopCountBadUtility = loopCount;
+#endif
 			// Assuming that the list is still sorted,
 			// and the following has only Goal::BAD_UTILITY
 			break;
@@ -864,7 +875,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 
 #if defined(_DEBUG)
 	sint32 committed_agents_test = 0;
-
+      loopCount = 0;
 	for
 	(
 	    Goal_List::iterator goal_iter2  = m_goals.begin();
@@ -872,6 +883,9 @@ void Scheduler::Match_Resources(const bool move_armies)
 	                      ++goal_iter2
 	)
 	{
+		loopCount++;
+		if (loopCount == loopCountBadUtility)
+			break;
 		Goal_ptr goal_ptr        = static_cast<Goal_ptr>(*goal_iter2);
 		committed_agents_test   += goal_ptr->Get_Agent_Count();
 	}
--- a/ctp2_code/ai/strategy/scheduler/scheduler.cpp
+++ b/ctp2_code/ai/strategy/scheduler/scheduler.cpp
@@ -682,9 +682,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 			goal_ptr->Rollback_All_Agents(); // Just roll back but don't report to the build list
 
 			// Actually should be checked in the next cycle, but there still seems to be something wrong.
-			Goal_List::iterator tmp_goal_iter = goal_iter;
-			goal_iter++;
-			m_goals.erase(tmp_goal_iter);
+			m_goals.erase(goal_iter);
 
 			/*
 			// Move to the end
@@ -716,11 +714,8 @@ void Scheduler::Match_Resources(const bool move_armies)
 					// http://www.cplusplus.com/reference/stl/list/splice.html
 					//or use a decrement
 					// Sort the goal list, move iterator increment herein back
-					tmp_goal_iter = goal_iter;
-					--goal_iter;
 					goal_ptr->Rollback_All_Agents(); // Just roll back but don't report to the build list
-					Reprioritize_Goal(tmp_goal_iter);
-					goal_iter++;
+					Reprioritize_Goal(goal_iter);
 					continue;
 				}
 			}
@@ -822,10 +817,9 @@ void Scheduler::Match_Resources(const bool move_armies)
 					{
 						AI_DPRINTF(k_DBG_SCHEDULER, m_playerId, goal_ptr->Get_Goal_Type(), -1,
 							("\t\t Transporters found.\n"));
-						--goal_iter;
 						committed_agents -= goal_ptr->Get_Agent_Count();
 						goal_ptr->Rollback_All_Agents(); // No we don't want to report this to the build list
-						break;
+						continue;
 					}
 				} // if out_of_transports is true, we just fail like in the original code
 			}
--- a/ctp2_code/gfx/spritesys/FacedSpriteWshadow.cpp
+++ b/ctp2_code/gfx/spritesys/FacedSpriteWshadow.cpp
@@ -599,7 +599,7 @@ void FacedSpriteWshadow::SetFrameData(uint16 facing, uint16 frame, Pixel16 *data
 	Assert(m_frames[facing] != NULL);
 	Assert(m_framesSizes[facing] != NULL);
 #ifdef _WINDOWS
-	Assert((((data == NULL) && (size = 0)) || ((data != NULL) && (_msize(data) == size))));
+	Assert((((data == NULL) && (size == 0)) || ((data != NULL) && (_msize(data) == size))));
 #endif
 	m_frames[facing][frame] = data;
 	m_framesSizes[facing][frame] = size;
@@ -612,7 +612,7 @@ void FacedSpriteWshadow::SetMiniFrameData(uint16 facing, uint16 frame, Pixel16 *
 	Assert(m_miniframes[facing] != NULL);
 	Assert(m_miniframesSizes[facing] != NULL);
 #ifdef _WINDOWS
-	Assert((((data == NULL) && (size = 0)) || ((data != NULL) && (_msize(data) == size))));
+	Assert((((data == NULL) && (size == 0)) || ((data != NULL) && (_msize(data) == size))));
 #endif
 
 	m_miniframes[facing][frame] = data;
@@ -672,7 +672,7 @@ void FacedSpriteWshadow::SetShadowFrameData(uint16 facing, uint16 frame, Pixel16
 	Assert(m_shadowFrames[facing] != NULL);
 	Assert(m_shadowFramesSizes[facing] != NULL);
 #ifdef _WINDOWS
-	Assert((((data == NULL) && (size = 0)) || ((data != NULL) && (_msize(data) == size))));
+	Assert((((data == NULL) && (size == 0)) || ((data != NULL) && (_msize(data) == size))));
 #endif
 	m_shadowFrames[facing][frame] = data;
 	m_shadowFramesSizes[facing][frame] = size;
@@ -685,7 +685,7 @@ void FacedSpriteWshadow::SetMiniShadowFrameData(uint16 facing, uint16 frame, Pix
 	Assert(m_miniShadowFrames[facing] != NULL);
 	Assert(m_miniShadowFramesSizes[facing] != NULL);
 #ifdef _WINDOWS
-	Assert((((data == NULL) && (size = 0)) || ((data != NULL) && (_msize(data) == size))));
+	Assert((((data == NULL) && (size == 0)) || ((data != NULL) && (_msize(data) == size))));
 #endif
 	m_miniShadowFrames[facing][frame] = data;
 	m_miniShadowFramesSizes[facing][frame] = size;
--- a/ctp2_code/ui/aui_common/aui_pixel.cpp
+++ b/ctp2_code/ui/aui_common/aui_pixel.cpp
@@ -34,62 +34,6 @@ uint16 aui_Pixel::Get16BitRGB( uint8 red, uint8 green, uint8 blue )
     }
 }
 
-uint8 aui_Pixel::GetPaletteIndexedColor(uint8 red, uint8 green, uint8 blue, HPALETTE *hpal)
-{
-#ifdef __AUI_USE_DIRECTX__
-	PALETTEENTRY *  pe              = new PALETTEENTRY[256];
-	size_t const    paletteCount    =
-        static_cast<size_t>(GetPaletteEntries(*hpal, 0, 256, pe));
-    Assert(paletteCount == 256);
-
-	uint16          valMain         = ColorCode555(red, green, blue);
-	uint8           color           = 0;
-	sint32          diff            = INT_MAX;
-
-	for (uint8 i = 0; i < static_cast<uint8>(paletteCount); ++i)
-    {
-		uint16  valCompare = ColorCode555(pe[i].peRed, pe[i].peGreen, pe[i].peBlue);
-
-		if (abs(valCompare - valMain) < diff )
-        {
-			color   = i;
-			diff    = abs(valCompare - valMain);
-			if (diff == 0) break;
-		}
-	}
-
-	delete [] pe;
-	return color;
-#else
-	// TODO: 8bit modes wanted?
-	Assert(0);
-	return 0;
-#endif
-}
-
-uint8 aui_Pixel::GetPaletteIndexedColor( uint8 red, uint8 green, uint8 blue, RGBQUAD *rgbq )
-{
-	uint8   color   = 0;
-	sint32  diff    = INT_MAX;
-	uint16  valMain = ColorCode555(red, green, blue);
-
-	for (uint8 i = 0; i < 256; i++)
-    {
-		uint16  valCompare = ColorCode555(rgbq[i].rgbRed, rgbq[i].rgbGreen, rgbq[i].rgbBlue);
-
-		if (abs(valCompare - valMain) < diff)
-        {
-			color   = i;
-			diff    = abs(valCompare - valMain);
-
-			if (diff == 0) break;
-		}
-	}
-
-	return color;
-
-}
-
 AUI_ERRCODE aui_Pixel::Convert24To16Dither(
 							 uint16 *buf16,
 							 uint8 *buf24,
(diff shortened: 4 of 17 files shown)


ctp2_code/ai/diplomacy/dstateevent.cpp
```diff
@@ -34,12 +34,12 @@ STDEHANDLER(InitDStateEvent)
 		return GEV_HD_Continue;
 
 	Diplomat & diplomat = Diplomat::GetDiplomat(playerId);
-	AiState state;
+	AiState thisstate;
```
williamd67, line 37: Why is state changed to 'thisstate'? Is it needed from a technical point of view? Personally I do not like names prefixed with 'the' or 'this' or something else that is not functionally describing the variable.

This also makes the changelist much bigger than maybe needed.

ctp2_code/ai/diplomacy/sstateevent.cpp
```diff
@@ -701,7 +701,7 @@ STDEHANDLER(DefenseLevel_NextSStateEvent)
 	Diplomat & diplomat = Diplomat::GetDiplomat(playerId);
 	diplomat.ClearEffectiveRegardCache();
 
-	AiState state;
+	AiState thisState;
```
williamd67, line 704: Inconsistent naming previously it was called 'thisstate' with small 's'. I prefer this with capital 'S'. Still I prefer not to prefix it with 'the' or 'this'. In other places in the code 'the' is used.

ctp2_code/ai/diplomacy/nproposalevent.cpp
```diff
@@ -713,7 +713,7 @@ STDEHANDLER(TradePact_NewProposalEvent)
 	if (!args->GetInt(0, desireType))
 		return GEV_HD_Continue;
 
-	if ((MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_MAKE_FRIEND ||
+	if ((MOTIVATION_TYPE) desireType != MOTIVATION_DESIRE_MAKE_FRIEND &&
```
williamd67, line 716: Although I agree with you that this code make no sense. Did you test the change or reasoned that it is an improvement?

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -631,6 +631,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 
 	sint32 committed_agents = 0;
 	sint32 total_agents     = m_agents.size();
+	bool needToGoBackOne = false;
```
williamd67, line 634: Can this be solved without adding an extra variable? It is quiet hard to keep track of all these cases and in most cases it is not needed when the algorithm is changed a little.

williamd67:
Thanks a lot for your contribution. I will review your changes and add some comments/questions. The aim of the comments is to increase the quality of the code and is in no way a comment on your contribution. All comments should be related to technical elements which should improve the quality of the code. In case you disagree just say so.

My experience with review changes are that they almost take no time and in the end make the code much better.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -867,6 +883,231 @@ void Scheduler::Match_Resources(const bool move_armies)
 		}
 	}
 
+	if (needToGoBackOne)
```
williamd67, line 886: Wow, did you add all this code?  I am not familiar with the AI so I cannot review this.

In case it is a copy, please try not to copy code but use methods if possible.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -867,6 +883,231 @@ void Scheduler::Match_Resources(const bool move_armies)
...
+			//		Assert(goal_ptr->Get_Agent_Count() == 0); // Is still ok
+					// City garrison problem
+			goal_ptr->Rollback_All_Agents(); // Just roll back but don't report to the build list
+
+			// Actually should be checked in the next cycle, but there still seems to be something wrong.
+			Goal_List::iterator tmp_goal_iter = goal_iter;
+			m_goals.erase(tmp_goal_iter);
+
+			/*
+			// Move to the end
```
williamd67, line 939: Please do not add 'new' commented code. We have our source-code repository which we can inspect when we need this code again. In the future it will be hard to understand why this code is commented out or will just make the reading harder/code dirty.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -879,6 +1120,8 @@ void Scheduler::Match_Resources(const bool move_armies)
 	{
 		Goal_ptr goal_ptr        = static_cast<Goal_ptr>(*goal_iter2);
 		committed_agents_test   += goal_ptr->Get_Agent_Count();
+		if (goal_ptr->Get_Agent_Count() > 0)
+			committed_agents_test = committed_agents_test;
```
williamd67, line 1124: This seems an odd statement as it says a = a; Please check what is intended.

ctp2_code/gfx/spritesys/FacedSpriteWshadow.cpp
```diff
@@ -599,7 +599,7 @@ void FacedSpriteWshadow::SetFrameData(uint16 facing, uint16 frame, Pixel16 *data
 	Assert(m_frames[facing] != NULL);
 	Assert(m_framesSizes[facing] != NULL);
 #ifdef _WINDOWS
-	Assert((((data == NULL) && (size = 0)) || ((data != NULL) && (_msize(data) == size))));
+	Assert((((data == NULL) && (size == 0)) || ((data != NULL) && (_msize(data) == size))));
```
williamd67, line 602: Well found

ctp2_code/gfx/spritesys/FacedSprite.cpp
```diff
@@ -431,7 +431,7 @@ void FacedSprite::DrawDirect(aui_Surface *surf, sint32 drawX, sint32 drawY, sint
 
 
 
-void FacedSprite::DirectionalDraw(sint32 drawX, sint32 drawY, sint32 facing,
+void FacedSprite::DirectionalDraw(sint32 drawX, sint32 drawY, uint32 facing,
```
williamd67, line 434: Should this change not more general be executed? The caller may still use a signed integer. The method shown in this review 'Draw' still uses a sint32 for facing. I would not do this so local as it will confuse new contributors if it sometimes 'sint32' and sometimes 'uint32'. In case it was really needed for a compiler issue, can a comment be added why this has been changed here but not in other places.

ctp2_code/ui/aui_common/aui_pixel.cpp
```diff
@@ -111,12 +111,12 @@ AUI_ERRCODE aui_Pixel::Convert24To16Dither(
 	if (!thisrerr || !thisgerr || !thisberr ||
 		!nextrerr || !nextgerr || !nextberr)
     {
-		delete thisrerr;
-		delete thisgerr;
-		delete thisberr;
-		delete nextrerr;
-		delete nextgerr;
-		delete nextberr;
+		delete[] thisrerr;
```
williamd67, line 114: Ah... here is 'this' used. I did not see this code before. I still think that in general we should not use it.

ctp2_code/ui/aui_common/aui_pixel.cpp
```diff
@@ -73,7 +73,7 @@ uint8 aui_Pixel::GetPaletteIndexedColor( uint8 red, uint8 green, uint8 blue, RGB
 	sint32  diff    = INT_MAX;
 	uint16  valMain = ColorCode555(red, green, blue);
 
-	for (uint8 i = 0; i < 256; i++)
```
williamd67, line 76: Would this case an endless loop or did it accidentally work?

ctp2_code/ui/aui_ctp2/c3_listitem.cpp
```diff
@@ -30,7 +30,7 @@ c3_ListItem::~c3_ListItem()
 		delete m_childList->GetNext( position );
 }
 
-sint32 c3_ListItem::Compare(c3_ListItem *item2, uint32 column)
+sint32 c3_ListItem::Compare(c3_ListItem *item2, sint32 column)
```
williamd67, line 33: Can you explain why we should compare against -2? Is that not a bug in another place which we should fix in stead of making the column unsigned? Although it is indeed strange that the code contains a test on smaller than 0 on an unsigned variable.

I checked the callers and they use singed integers so I think this is a good improvement. Thanks.

henshall828:
> 
> 
> Thanks a lot for your contribution. I will review your changes and add some comments/questions. The aim of the comments is to increase the quality of the code and is in no way a comment on your contribution. All comments should be related to technical elements which should improve the quality of the code. In case you disagree just say so.
> 
> My experience with review changes are that they almost take no time and in the end make the code much better.
> 
> BTW: this is my first review on github ;) In my day-to-day work I do a lot of reviews.

Thanks for your quick response and comprehensive review :) I agree with most of your points, and will add respond and revise accordingingly.

Many thanks

ctp2_code/ai/strategy/scheduler/scheduler.cpp
henshall828: I meant to remove this it was just somewhere for me to put a breakpoint when trying to find why the error in the assert was occuring. I will fix this.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
henshall828: I think the approach can be changed by iterating through the list in a different fashion - this will hopefully mean any extra variables will be local to the loop.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
henshall828: This is copied from the loop above. The problem this changeset is trying to solve is that the code can remove goals from the list, but would crash if the first goal was removed as it would try to set the iterator to the previous goal, then remove the goal - which doesn't work if the goal to be removed is the first goal.

The fix as implemented catches this case and instead moves the iterator to the next goal and removes the previous one. The loop then moves the iterator on to the item following the one that needs to be processed, so the variable added is used tp . The problem is if there are only 1 or 2 goals. If there is only one goal, this cannot be removed in the iterator loop, otherwise the loop will then be working with an iterator that points to nothing - causing a crash. If there are 2 goals then the loop will terminate without processing the second goal. I therefore added the variable check for this case and repeat the loop to either remove the only item in the list or process it as necessary.

There should be a better way to do this that is overall more elegant and doesn't require copying the loop code (which you are right is a bad idea - it means that any code fixes now need to be made in two places). I contemplate this further.

ctp2_code/ai/diplomacy/nproposalevent.cpp
henshall828: This is a reasoned improvement. Unfortunately I am only just beginning to look at the code, so am not yet familiar enough to be certain that this is what was intended (and should happen).
I am certain the previous behaviour of always executing code in the if statement was unintentional, though.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
henshall828: This was copy+pasted from the existing code - I will remove this and the original when I improve this implementation.

ctp2_code/ui/aui_ctp2/c3_listitem.cpp
henshall828: I think the use of a negative number was meant to show that we are not sorting on a column but instead doing a special type of sort - however this is only implemented in loadsavewindow.cpp, where k_LOADSAVE_AUTOSORT_COL is defined as -2. The change for this file and header was mistakenly left out of this branch when cherry-picking the commits.

This doesn't seem to be used anywhere, so perhaps a better fix might be to remove the code from loadsavewindow and change all the callers to use unsigned int. This would also mean redundant code that checks whether column is < 0 (which can never be true for an unsigned int) could be removed from many of the compare functions.
The current implementation causes a debug compile error for me on Fedora 32 linux (and presumably other distributions with the same version of gcc), as the compiler detects that -2 is outside the defined range for an unsigned int.

ctp2_code/ui/aui_common/aui_pixel.cpp
henshall828: As a uint8 i will always be less than 256, but this will not loop endlessly as it will find a colour match, at which point it exits the loop. This change only fixes an issue because the compiler detects that checking against 256 is incorrect.

I guess the only execution problem that could even potentially occur would be if the compiler tried to demote 256 to unit8 do the check, rather than promoting i to int, in which case 256 would become 0 and the loop would never execute. I don't think any compiler would actually be allowed to do that, though.

ctp2_code/ai/diplomacy/dstateevent.cpp
henshall828: Agreed - this is an uneccesary change and was motivated by misreading a different ld error on Fedora 32 when configured with --enable-debug due to int state being defined in both ldl.l and slic.l

I will revert this change.

ctp2_code/ai/diplomacy/sstateevent.cpp
henshall828: Agreed - this is an unecessary change and was motivated by misreading a different ld error on Fedora 32 when configured with --enable-debug due to int state being defined in both ldl.l and slic.l

I will revert this change.

williamd67:
Looks good to me. I have no experience with the AI (goal) code so maybe someone else may have a look as well.

MartinGuehmann:
The commit history is hard to understand. You added changes and then removed them and added then them back again. I am not really a fan of rebuilding history, but here it is just too complicated. I would suggest to start a new branch from master and cherry-pick all the commits you want to keep as they are. And then add any other commits, which need modification. Once done you can delete the old branch and rename the new branch to the old one and then force push the new branch to GitHub replacing the old one there.

Additionally, you added a new file called Scheduler.cpp even so the file you wanted to edit was scheduler.cpp with lower case 's'.

henshall828:
Apologies - one of the downsides of working in Windows is the lack of case sensitivity. Currently the code has scheduler.cpp and Scheduler.h - I must have automatically changed the filename to match.
I will re-create the branch with just the fixes and without the reverts - this is my first time pull request and review, so I wasn't sure if it was preferred to show the response to the review in the git history or not - it will be certainly be cleaner without it.

henshall828:
The history and patch descriptions have now been cleaned up. This is hopefully easier to follow, and gets rid of the strange issue where I ended up with Scheduler.cpp and scheduler.cpp due to Windows lack of case sensitivity, while still allowing different cases in filenames. This caused OneDrive to be very confused!

ctp2_code/gfx/spritesys/FacedSprite.cpp
henshall828: This change was incorrectly including in this series as it was part of wider changes I was making attempting to reduce uninteresting warnings so that I could find actual issues - this patch has now been removed.

LynxAbraxas:
Welcome @henshall828 to ctp2 and many thanks for your PR. You caught quite a few inconsistencies and mistakes in the code. Also many thanks to @williamd67 for the very thorough review and for @MartinGuehmann recommendations, which I fully support. That's the concept of PRs with WIP (Work In Progress) as a prefix to the PR title. That way it can be discussed and the history rewritten as needed as long as the PR was not merged.
I'll try to take a close look and to test, but my time for CTP2 is currently very limited, so please go ahead without a report of mine if I do not respond within 10 days or so.

henshall828:
Thanks @LynxAbraxas for agreeing to try to test and to @williamd67 and @MartinGuehmann for their through review and helpful advice. As my first time using GitHub and submitting a pull request, this has been a very positive experience for me - leading to a nice, focused series of patches.

MartinGuehmann:
@henshall828, you missed this one:

https://github.com/civctp2/civctp2/blob/7857aefacd75cd08acbd5f20ab3fef2531df34c1/ctp2_code/ai/strategy/scheduler/scheduler.cpp#L823-L828

In case, the goal needs transporters these are added agent lists, and then the goal is tried to be executed again.

With the new loop design the iterator decrement has to go and break has to be replaced with a continue.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -686,7 +683,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 
 			// Actually should be checked in the next cycle, but there still seems to be something wrong.
 			Goal_List::iterator tmp_goal_iter = goal_iter;
-			--goal_iter;
+			goal_iter++;
 			m_goals.erase(tmp_goal_iter);
```
MartinGuehmann, line 684: That could be simplified, list::erase returns the iterator after the one to be erased. The code should then be a one liner, with eliminating the temporary iterator: goal_iter = m_goals.erase(goal_iter);

The commented lines after this can also go.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -723,6 +720,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 					--goal_iter;
```
MartinGuehmann, line 717: That isn't right either. And my feeling is that this code around has some more problems.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -664,9 +664,6 @@ void Scheduler::Match_Resources(const bool move_armies)
 
 			AI_DPRINTF(k_DBG_SCHEDULER, m_playerId, goal_ptr->Get_Goal_Type(), -1, ("\n"));
 			AI_DPRINTF(k_DBG_SCHEDULER, m_playerId, goal_ptr->Get_Goal_Type(), -1, ("\n"));
-			// Assuming that the list is still sorted,
-			// and the following has only Goal::BAD_UTILITY
-			break;
```
MartinGuehmann, line 672: That isn't really a good idea to remove it without testing. If the assumption is false and the list isn't sorted anymore than we have a bug here, since the AI is supposed to consider the goals with higher priority first.

Anyway, if we leave the loop here, then committed_agents_test == committed_agents should be true, actually it should be true at any point in the loops, which doesn't seem to be true. In fact, it should updated whenever agents are committed or rolled back.

In fact, it is another optimization and the behavior is correct as long as committed_agents <= committed_agents_test.

Actual, you can have thousands of goals in m_goals, and it is quite some code that is executed during one iteration. Therefore, you want to optimize this as much as you can.

henshall828:
> 
> 
> @henshall828, you missed this one:
> 
> https://github.com/civctp2/civctp2/blob/7857aefacd75cd08acbd5f20ab3fef2531df34c1/ctp2_code/ai/strategy/scheduler/scheduler.cpp#L823-L828
> 
> In case, the goal needs transporters these are added agent lists, and then the goal is tried to be executed again.
> 
> With the new loop design the iterator decrement has to go and break has to be replaced with a continue.

This works as previously, since it then falls through to the end of the loop and is then re-incremented - however you are correct that this can now be replaced with a simple continue. I will include this as a separate patch as it is not a code fix, but a minor code improvement that is now made possible.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
henshall828: I agree that this change is incorrect. Would be better if committed_agents_test could be calculated in such a way that the Assert could remain the same, as I can imagine there are valid and invalid ways for committed_agents to be < committed_agents_test?

MartinGuehmann:
> > @henshall828, you missed this one:
> > https://github.com/civctp2/civctp2/blob/7857aefacd75cd08acbd5f20ab3fef2531df34c1/ctp2_code/ai/strategy/scheduler/scheduler.cpp#L823-L828
> > 
> > In case, the goal needs transporters these are added agent lists, and then the goal is tried to be executed again.
> > With the new loop design the iterator decrement has to go and break has to be replaced with a continue.
> 
> This works as previously, since it then falls through to the end of the loop and is then re-incremented - however you are correct that this can now be replaced with a simple continue. I will include this as a separate patch as it is not a code fix, but a minor code improvement that is now made possible.

No that is not a minor code improvement, it is the same thing you wanna fix. Again you may decrement before m_goals.begin(). In that case you get undefined behavior and are on the mercy of your compiler to get the intended behavior. I just googled for it that are the first two links:

https://stackoverflow.com/questions/42237716/decrementing-stdvectoriterator-before-stdvectorbegin
https://stackoverflow.com/questions/19417171/stl-iterator-before-stdmapbegin

LynxAbraxas:
> > This works as previously, since it then falls through to the end of the loop and is then re-incremented - however you are correct that this can now be replaced with a simple continue. I will include this as a separate patch as it is not a code fix, but a minor code improvement that is now made possible.
> 
> No that is not a minor code improvement, it is the same thing you wanna fix. Again you may decrement before m_goals.begin(). In that case you get undefined behavior and are on the mercy of your compiler to get the intended behavior. I just googled for it that are the first two links:
> 
> https://stackoverflow.com/questions/42237716/decrementing-stdvectoriterator-before-stdvectorbegin
> https://stackoverflow.com/questions/19417171/stl-iterator-before-stdmapbegin

So, is this the only issue left to get this ready for merging?

MartinGuehmann:
> > > This works as previously, since it then falls through to the end of the loop and is then re-incremented - however you are correct that this can now be replaced with a simple continue. I will include this as a separate patch as it is not a code fix, but a minor code improvement that is now made possible.
> > 
> > 
> > No that is not a minor code improvement, it is the same thing you wanna fix. Again you may decrement before m_goals.begin(). In that case you get undefined behavior and are on the mercy of your compiler to get the intended behavior. I just googled for it that are the first two links:
> > https://stackoverflow.com/questions/42237716/decrementing-stdvectoriterator-before-stdvectorbegin
> > https://stackoverflow.com/questions/19417171/stl-iterator-before-stdmapbegin
> 
> So, is this the only issue left to get this ready for merging?

I think so.

LynxAbraxas:
> > So, is this the only issue left to get this ready for merging?
> 
> I think so.

@henshall828 could you do this? The PR would be ready for a merge then.

henshall828:
> 
> 
> > > So, is this the only issue left to get this ready for merging?
> > 
> > 
> > I think so.
> 
> @henshall828 could you do this? The PR would be ready for a merge then.

Thanks for testing this @LynxAbraxas.
 
I have now changed the code so that the iterator isn't decremented when it could be at the beginning of the list. I did find one more instance of this pattern of an iterator in a for loop being decremented in Goal.cpp, so I have also changed this to a while loop in the same way.

LynxAbraxas:
Many thanks @henshall828 for adjusting, I'll test.

ctp2_code/ui/aui_common/aui_pixel.cpp
```diff
@@ -45,8 +45,8 @@ uint8 aui_Pixel::GetPaletteIndexedColor(uint8 red, uint8 green, uint8 blue, HPAL
 	uint16          valMain         = ColorCode555(red, green, blue);
 	uint8           color           = 0;
 	sint32          diff            = INT_MAX;
-
-	for (uint8 i = 0; i < static_cast<uint8>(paletteCount); ++i)
+	// palleteCount must be 256
+	for (uint8 i = 0; i <= 255; ++i)
```
MartinGuehmann, line 49: This piece still smells. In principle, `i` could still become 255 and then be incremented which would lead to an overrun and `i` becomes 0. So either turn `i` into a uint16 or sint16, or add a break line at the end of the loop like `if(i == 255) break;`.

henshall828: You're right - I was assuming this routine would always find a match, and therefore that overflow would never happen (as otherwise these routines would never have worked).

I have now searched through the history of these routines and they have never actually been used in all of the git history (though they have been refactored) - this is why the fact they were incorrect didn't seem to break anything. Should I just do a patch removing them and the supporting routine entirely?

MartinGuehmann:
I tested it. I am not really happy with the AI right now, but it does not seem to be related to this. So I would say it can go in. Any other comments, before I merge?

LynxAbraxas:
> I am not really happy with the AI right now, but it does not seem to be related to this.

Yes, I think so too. Problems with the AI I noticed (#224, #259, #175, #172) were already there before this PR.

MartinGuehmann:
> > I am not really happy with the AI right now, but it does not seem to be related to this.
> 
> Yes, I think so too. Problems with the AI I noticed (#224, #259, #175, #172) were already there before this PR.

It's not really these. My impression is that the AI is not taking so many cities, even so it has war and the means to take the city. There was a Barbarian city for quite some time that it did not take back, but it tried to establish an embassy there.

Anyway, let't merge this.

LynxAbraxas:
> There was a Barbarian city for quite some time that it did not take back, but it tried to establish an embassy there.

Yes, observed that too, see e.g. https://github.com/civctp2/civctp2/pull/259#issuecomment-595787920.

LynxAbraxas:
Hm, just figured that this PR causes a
```
double free or corruption (fasttop)
```
when ending the turn for this save-game [test.zip](https://github.com/civctp2/civctp2/files/4854042/test.zip). It also causes a segfault if opening a scenario if #312 is included. I had tested this PR with save-games at an earlier stage, which did not lead to the above error, so my impression is, that some rebase between force push has introduced an unintentional change.
@henshall828 would you take a look?

williamd67:
I got a segfault while playing and was able to reproduce and debug it.

@henshall828, I think that line 696 of scheduler.cpp needs a slight change as erase will invalidate the iterator, you have to use the returned iterator to continue the loop.
`			goal_iter = m_goals.erase(goal_iter);`

henshall828:
> 
> 
> I got a segfault while playing and was able to reproduce and debug it. I checked that my segfault was gone with below change but did not test it thoroughly.
> 
> @henshall828, I think that line 696 of scheduler.cpp needs a slight change as erase will invalidate the iterator, you have to use the returned iterator to continue the loop.
> ` goal_iter = m_goals.erase(goal_iter);`

Yes, you are completely correct - apologies, this must have been accidentally lost when re-doing the patch.

henshall828:
> 
> 
> Hm, just figured that this PR causes a
> 
> ```
> double free or corruption (fasttop)
> ```
> 
> when ending the turn for this save-game [test.zip](https://github.com/civctp2/civctp2/files/4854042/test.zip). It also causes a segfault if opening a scenario if #312 is included. I had tested this PR with save-games at an earlier stage, which did not lead to the above error, so my impression is, that some rebase between force push has introduced an unintentional change.
> @henshall828 would you take a look?

Does changing the code in scheduler.cpp at line 721 to:
`			if(tmp_goal_iter != m_goals.end())
			{
				Utility nextMatchValue = static_cast<Goal_ptr>(*tmp_goal_iter)->Get_Matching_Value();
				if(newMatchValue < nextMatchValue)
				{
					// http://www.cplusplus.com/reference/stl/list/splice.html
					//or use a decrement
					// Sort the goal list, move iterator increment herein back
					tmp_goal_iter = goal_iter;
					goal_ptr->Rollback_All_Agents(); // Just roll back but don't report to the build list
					Reprioritize_Goal(tmp_goal_iter);
					continue;
				}
			}`
fix this error? Reprioritize_Goal seems to be able to make the input iterator invalid somehow.

LynxAbraxas:
Changes of https://github.com/LynxAbraxas/civctp2/commit/ee1ca54ebb7fd495b344ba969b64bd8c6fd3abb8 did not solve it for me, still got a segfault during the change of turn.

henshall828:
> 
> 
> Changes of [LynxAbraxas@ee1ca54](https://github.com/LynxAbraxas/civctp2/commit/ee1ca54ebb7fd495b344ba969b64bd8c6fd3abb8) did not solve it for me, still got a segfault during the change of turn.

I should have explained that the fix from @williamd67 is also required at line 696
`goal_iter = m_goals.erase(goal_iter);`

The fix for Reporitize_Goal isn't quite right - I will change it so that Reprioritize_Goal returns the iterator to use, which will make this bit of code correct and cleaner.

LynxAbraxas:
Yes, combination of both (https://github.com/LynxAbraxas/civctp2/commit/c1a58d546010abbf4c74444fef0374f00da89ff2 https://github.com/LynxAbraxas/civctp2/commit/d4d23bf947c1a5cc493461d13b28a75538a5ba62) seems to do the trick.

PR #313, opened 2020-05-21
Correct trade route deletion

LynxAbraxas:
Changes missing from #256: In case a city is destroyed its trade routes must be killed before the CityData destructor is called because then it is not possible any more to determine the owner to remove the SeenByBit.


MartinGuehmann:
Looks fine as far as I can see. So let's merge.

Issue #299, opened 2020-04-08
Attacks with special effects in FOW not fully hidden


LynxAbraxas:
It seems attacks with special effects (like conversion, franchising, stealing advances) if happening in FOW (e.g. done by the AI) are not fully hidden: The map focuses on the attack position and the sound of the special attack is played, giving away the attack to human players, and can become an annoyance if happening many times between turns.

MartinGuehmann:
Actually, I like it when the sounds are played, this way you know that something is going on. But of course it shouldn't focus there.

LynxAbraxas:
> this way you know that something is going on

True, but that is a kind of "cheating" since it is some info only humans can make use of, i.e. the AI so far cannot, and being in FOW should be covered to be concise with the game concept, otherwise one could argue that even the animation should be shown even in FOW. And it can become annoying when you have shower of special attack sounds each turn.

LynxAbraxas:
I can also observe special effects for units that are invisible to me like conversion, franchising, stealing advances. This gives away the unit in a way that the AI cannot make use of, so in a way it is also an unfair advantage to human player.

williamd67:
My thoughts on this topic.

Currently the End-turn behavior is influenced by two settings:
1. center map
2. show enemy moves
* If both are disabled sounds of the special effects are played and in case the special effect is on screen an animation; for me that is hard to follow and not very informative;
* In case center maps is enabled, and show enemy moves is disabled; the map jumps to the place were the special effect is happening, and before the animation or the sound is finished, jumps to the next special effect; for me this is also hard to follow and in most cases before I can interpret the effect there is a jump to another place;
* In case both center maps and show enemy moves is enabled: I can, in most cases, follow the actions and effects; this has a negative side-effect that it takes quiet some time to see all enemy moves, especially with larger maps and further in the game (can be sped up by making the animations shorter but still);

So next to @LynxAbraxas remark that it is unfair to the AI-player, we could, in my opinion, improve the End-turn process for the human player as well, by:
* only play animations and sounds that involve the human player so no animations and sounds for the actions between AI-players (or other human-players if applicable)
* or, make it clear (slow enough) for the human-player so (s)he can both understand and enjoy the special effects; although this will slow down his/her process.

Do you have any other ideas, suggestions, or improvements?

LynxAbraxas:
[...]

Issue #247, opened 2020-01-16
Work map reactivation for city window


LynxAbraxas:
https://github.com/civctp2/civctp2/pull/195#issuecomment-570683690 revealed that not being able to see which tiles/rings are worked on can cause odd behavior (like the loss of a trade route) hard to understand. While the trade route issue can be solved with #195, it would still be interesting to have a visualization of which tiles/rings are worked on, similar to CIV2 and CTP1. In CTP1 such a visualization popped up with a right click on a selected city (which always had 20 tiles of influence no matter of the city population):
![ss_2020-01-16_09:43:25](https://user-images.githubusercontent.com/21012234/72511964-ccece000-384b-11ea-8d34-a4aa36f66811.png)
![ss_2020-01-16_09:50:09](https://user-images.githubusercontent.com/21012234/72512135-dd04bf80-384b-11ea-8d3b-f2f977beba7c.png)
![ss_2020-01-16_09:52:10](https://user-images.githubusercontent.com/21012234/72512185-e1c97380-384b-11ea-8354-b488d69db006.png)

It was also possible to move the workers to other tiles:
![ss_2020-01-16_09:44:23](https://user-images.githubusercontent.com/21012234/72512014-d1b19400-384b-11ea-838e-c6edf56c700e.png)
![ss_2020-01-16_09:44:58](https://user-images.githubusercontent.com/21012234/72512082-d7a77500-384b-11ea-9c5c-66c55af7c6a6.png)

LynxAbraxas:
It might be possible to reactivate the CTP1 workmap
https://github.com/civctp2/civctp2/blob/e404a9abe18ce22ddfc4c12ed7f6b0c24bd0ba5f/ctp2_data/default/uidata/layouts/ctp2.ldl#L294-L315
and workwin
https://github.com/civctp2/civctp2/blob/e404a9abe18ce22ddfc4c12ed7f6b0c24bd0ba5f/ctp2_code/ui/interface/workwin.cpp#L144-L155

MartinGuehmann:
Actually, I wouldn't put in the workers. First, it is quite some work, and seconds it adds too much micromanagement.

I just would have it so that if a good is within the collection radius, it is collected and if you have for each tile a worker or a slave then all the resources are collected. And if less then only the fraction of resources according to the fraction of workers/slaves you have for tiles.

I thing that is easy to do and easy to understand for the player, at least if it is a bit better explained.

Issue #229, opened 2020-01-03
Launched space plane newer seems to reach target


LynxAbraxas:
It seems that units loaded into a space plane and launched never reach their destination, at least not even after more than 20 turns, neither for land nor water cities.
Space lifts would become interesting for far-off cities created for building obelisks for the science victory.

MartinGuehmann:
Space lifts only work between existing cities, so why would they be interesting? You can already build obelisks in cities you own, even so they are on the other side of the world.

LynxAbraxas:
> Space lifts only work between existing cities

That's what I was expecting and found not to work.

> so why would they be interesting? You can already build obelisks in cities you own, even so they are on the other side of the world.

Yes, but those cities might need a garrison which in some cases could be provided faster with a space lift than by other means.
While I tested the space lift between land (and water) cities, it would be most interesting if it also worked for sea colonies.

LynxAbraxas:
I now had my very first successful space transport, will test this further.

LynxAbraxas:
It seems `m_reentryTurn` and `m_reentryPos` are not saved, so the problem arises if a space plane is in space when a game is saved and reloaded. Then

https://github.com/civctp2/civctp2/blob/11b9a024f4e0b5ace7011f6e53a49f5f7bb83651/ctp2_code/gs/gameobj/ArmyData.cpp#L1735
is true because `m_reentryTurn` is initialized to `-1` which leads to
https://github.com/civctp2/civctp2/blob/11b9a024f4e0b5ace7011f6e53a49f5f7bb83651/ctp2_code/gs/gameobj/ArmyData.cpp#L4685-L4696
where the not stored `m_reentryPos` leads to an invalid city and the destruction of the space plane and its units.
So to solve this, where do we have some unused space in the current save game format where `m_reentryTurn` and `m_reentryPos` could be saved? @MartinGuehmann any ideas?

LynxAbraxas:
Solved with #321.

PR #314, opened 2020-05-23
Refactor director

williamd67:
Director will be refactored where the goal is the make the code more straight-forward and dependencies one-way. I also try to limit the number of methods that are publicly available and
 that may interfere with the internal state of Director.

This should solve #307 (Incorrect unit-animations).


ctp2_code/gfx/spritesys/director.cpp
```diff
@@ -2037,43 +2004,23 @@ class DQActionBeginScheduler : public DQActionExternal
...
+	m_lastTickCount				(0),
+	m_timeLogIndex				(0),
+	m_averageElapsed			(0),
+	m_averageFPS				(k_DEFAULT_FPS),
+	m_paused					(FALSE),
+	m_processingActiveEffects	(FALSE),
+	m_actionQueue				(new PointerList<DQAction>),
+	m_actionWalker				(new PointerList<DQAction>::Walker),
+	m_pendingGameActions		(0),
+	m_endTurnRequested			(false)
```
MartinGuehmann, line 2245: Now this is harder to read. Please keep these things aligned. Keep in mind that tabs lengths is a setting in your editor and does not need to be at four spaces. In general uses tabs for things that are not screwed up by any tab setting. That are the things on the left. And use spaces for things that are affected by the tab settings like those on the right.

williamd67: > In general uses tabs for things that are not screwed up by any tab setting.  That are the things on the left. And use spaces for things that are affected by the tab settings like those on the right.

Thanks I will follow this guideline and update Director classes.

May I propose to not using `tabs` as most editors now support inject `spaces` for `tab`. I do understand that would be another change in style so a huge change. We may plan to update all files once (UTF-8, no-tabs, maybe other changes).

williamd67:
Found the following issues with unit-animations:
- **fixed**: no sound played for units that ends in transporter
- **fixed**: unit sound does not stop and plays through other unit sounds
- **not an issue**: televangelist not facing city when executing special attack (happened by co-incidence, televangelist was not part of the reform)
- **fixed**: unit animation jumping
- **fixed** : unit invisible (only shield is moving) - unit is deleted (per example settler) at end of move
- **fixed**: unit is not rendered till it is deleted
- **fixed**: death of units is not rendered
- death of unit inside city is not rendered

ctp2_code/gfx/spritesys/director.cpp
```diff
@@ -667,126 +669,126 @@ class DirectorImpl : public Director {
...
-	virtual void AddCombatFlash(MapPoint const & pos);
+	virtual void AddCombatFlash(const MapPoint &position);
 	virtual void AddCopyVision();
-	virtual void AddCenterMap(const MapPoint &pos);
+	virtual void AddCenterMap(const MapPoint &position);
 	virtual void AddSelectUnit(uint32 flags);
 	virtual void AddEndTurn();
 	virtual void AddBattle(Battle *battle);
-	virtual void AddPlaySound(sint32 soundID, MapPoint const & pos);
+	virtual void AddPlaySound(sint32 soundID, const MapPoint &position);
```
MartinGuehmann, line 760: Why is it now 'position' instead of 'pos'? 'pos' is the standard name across the code, whenever you are referring to a position on a map. Basically, this change is like renaming the loop variable 'i' to 'index'.

ctp2_code/gfx/spritesys/director.cpp
```diff
@@ -667,126 +669,126 @@ class DirectorImpl : public Director {
...
+		sint32      projectilePath
 	);
-	virtual void AddSpecialEffect(MapPoint &pos, sint32 spriteID, sint32 soundID);
-	virtual void AddMorphUnit(UnitActor *morphingActor, SpriteState *ss, sint32 type,  Unit id);
-	virtual void AddHide(Unit hider);
-	virtual void AddShow(Unit hider);
-	virtual void AddWork(Unit worker);
+	virtual void AddSpecialEffect(const MapPoint &position, sint32 spriteID, sint32 soundID);
+	virtual void AddMorphUnit(UnitActor *morphingActor, SpriteState *spriteState, sint32 type, const Unit &id);
+	virtual void AddHide(const Unit &hider);
```
MartinGuehmann, line 745: Why is this now a reference? This way, it should be now a 64 bit address on Linux and a 32 bit address on Windows. Even so the original object is just 32 bits.

In fact that argument can also be make for MapPoint, which are now 32 bits but were in CTP1 48 bits.

ctp2_code/gfx/spritesys/director.cpp
williamd67: This is a kind of fundamental question related to how I see code quality so the answer is a little longer than usual. I am happy to discuss when you disagree with any of the reasoning.

**Summary**: As I overvalued the `do not use abbreviations`-rule against the `be consistent`-rule. I will revert the change and replace `position` back to `pos`.

**Reasoning**
As developer I like the source code to be as readable as possible. Abbrevations distract from that. Abbreviations need to be interpreted and translated to the real message while reading. This costs energy. One of my statements is that if a book contains as many abbreviations as source code nobody would read it.

With code-completion there is also no need for short variable names. Although I think line length should be limited as well so there is a need for balance.

Abbreviations are not the only element for readability, consistency is another. Consistent code is easier to understand. So as in CTP2 `pos` is used for position everywhere the `Director` code is not consistent any more and violates the `be consistent`-rule which means that I need to revert the change.

BTW: I do not buy your statement that it is the same thing as renaming `i` to `index`. Positions have a functional meaning in CTP2 and therefor have more value than a loop variable. In case `i` would be used for an element that has a functional meaning in the application I would also have renamed it to `index`. I did it a couple of times with `p` -> `player`.

ctp2_code/gfx/spritesys/director.cpp
williamd67: Sorry for the long reply I want to explain and be concrete why I think this is a good change.

**Summary**: As in my opinion this is a technical improvement I propose to keep this change.

**Reasoning**
In general, when I am not optimizing, I do not want to know the exact size of all classes or the amount of time a copy-constructor takes when it is technically not needed. The exact size of an instance of a class may not be so obvious. The data-size may be small but in case a class has virtual methods also the virtual-table pointer will be part of the data and in case it is a derived class it becomes even more complicated. I am not 100% sure but depending on the situation there also may be a need for the `this` pointer to be stored.

In general, computers with registers have less work when using their native pointer-size as the values will be in registers that can be used immediately. In case a 16-bit value is stored in a 64-bit register it needs to be masked out (by shifting or by and-ing). As the compiler takes care of the pointer-size (ensuring that it fits the host) I, as developer, do not need to care about that which will ensure that I make less mistakes and be more efficient. 

So in case I need a class that is only used for reading some data without modifying the class and there is also no need for a copy (it will not be stored), I use the `const reference` as that gives me exactly what I need without any overhead. When an instance will be used in stead of a `const reference` the program will allocate space for the object (stack or heap) and execute a copy-constructor both are not needed in this situation.

BTW: Technically as the class already exists it may even be that the reference is given to the next method via a register which is not possible for the class itself which means less overhead. I am not sure but I also think that inlining a method becomes harder when the class itself is copied. While by using a `const reference` it is obvious for the compiler that it is the same class.

LynxAbraxas:
The long list of what is fixed with this is amazing. Will test this.
> death of unit's while enemy-moves is enabled is not shown (visibly - not in view; maybe add center-map)

@williamd67 what do you mean with this?

ctp2_code/gfx/spritesys/director.cpp
MartinGuehmann: I agree that this stuff is nothing we have to optimize, extensively. I just happened to have optimized some classes for size and also looked at the sizes of Unit and others to make sure that the Linux 64 bit executable can load game created by the Windows 32 bit executable.

So then we go here for consistency and ignore the implementation details, since they do not matter here. So we keep the change.

ctp2_code/gfx/spritesys/director.cpp
MartinGuehmann: It depends on the abbreviation, I even have a "Pos 1" on my keyboard so that should be quite common. In principle the more common it is the easier it is to remember and to understand.

And that is the thing about the variable i, it is so common that it does not need to be introduced. And that's what it has in common with pos. Of course, it is a metaphor, so not every aspects carries over. In fact renaming it to index would indicate that it is not a loop variable. In fact, that it is a one letter variable already indicates that it is used in a loop. So that meaning carries over to others of these kinds. So 'p' is most likely in the code an index into the g_player array. But I think it is quite rare. At least I couldn't find when I was looking for it.

Of course, the variable p may stand for position, player, or point, which is not very specific. And may be confusing. So if 'p' is a pointer on a Player object then it is more appropriately named player. You may also player_ptr.

williamd67: I agree and `pos` is definitely on the edge (here on the good side of the edge). The question is what is the edge. My experience is that it is easier to say no abbreviations than this one is allowed and that one is not.

williamd67:
> what do you mean with this?

Sorry for the cryptic description. What I meant is that when an enemy unit dies and 'show enemy moves' is enabled and the enemy is currently not on the visible map, a dying sound is played but the animation is not shown.

This could be fixed by adding a center-map on the position of the dying unit so that the unit is visible while the death animation is played.

ctp2_code/gs/slic/slicfunc.cpp
```diff
@@ -4243,7 +4241,7 @@ SFN_ERROR Slic_GetNearestWater::Call(SlicArgList *args)
 	}
 
 	SlicSymbolData *sym = args->m_argValue[1].m_symbol;
-	//Outcommented by Martin G�hmann
```
MartinGuehmann, line 4246: These comments and the out-commented code can also go. When we started working on the code we did it without source control due to licence concerns, and the changes were documented with #if defined(ACTIVISON_ORIGINAL) guards. Eventual, we stripped that. But these comments are still in, in without the original code they don't really make sense anymore. And they actually, also clutter the code.

ctp2_code/gs/slic/slicfunc.cpp
```diff
@@ -6721,15 +6719,15 @@ SFN_ERROR Slic_ClearBattleFlag::Call(SlicArgList *args)
...
-	//Wrong number of arguments added by Martin G�hmann
+	//Wrong number of arguments added by Martin Gühmann
 	if(args->Count() != 0)
 		return SFN_ERROR_NUM_ARGS;
 
 	open_ScenarioEditor();
 	return SFN_ERROR_OK;
 }
 
-//New Slic functions of CTP2.1 readded by Martin G�hmann
```
MartinGuehmann, line 6732: That might still be a useful comment.

ctp2_code/gfx/spritesys/director.cpp
```diff
@@ -3195,7 +3155,7 @@ void DirectorImpl::AddTerminateSound(const Unit &unit)
 	m_actionQueue->AddTail(action);
 }
 
-void DirectorImpl::AddInvokeThroneRoom(void)
```
MartinGuehmann: Just a note for the future: Removing these `void`s, should not change anything and is ideally done in a separate commit. This way it can be separated from the other stuff in case something goes wrong there.

williamd67: I totally agree and will in the future do `cleanup` actions in separate commits.

ctp2_code/gs/slic/slicfunc.cpp
williamd67: Done in a separate commit.

LynxAbraxas:
Makes a very good impression so far, need to test for a few more turns tough. One thing I noticed is that the sound of the trumpets (e.g. when pillaging) is not played any more. Is that intentional?

williamd67:
> One thing I noticed is that the sound of the trumpets (e.g. when pillaging) is not played any more. Is that intentional?

No that is not intentional. I need to investigate that.

There is one fix that may be related to this. Till now the success animations were played even if the action failed. Now it is only played when the action is successful.

LynxAbraxas:
> Till now the success animations were played even if the action failed.

In conjunction to that, I have the impression that now some failure sounds are played, that I have never heard before, really cool.

williamd67:
> One thing I noticed is that the sound of the trumpets (e.g. when pillaging) is not played any more.

This should fix the pillaging (and others) success and failure-sounds not being played. I also changed the `illegal` sound to the official `illegal` sound in stead of `failed` sound. Let me know if this change is not wanted or annoying.

BTW: I found a special pillaging-sound that could be specified in `Orders.txt`. Unfortunately `Orders.txt` expects an integer as sound-identifier which is very hard to determine. In other places a string-identifier is used. I will change that in a separate pull-request.

LynxAbraxas:
>This should fix the pillaging (and others) success and failure-sounds not being played.

Yes, fixed, plays nicely now.

> I also changed the illegal sound to the official illegal sound in stead of failed sound. 

I'll test this for some while, it's a bit unusual at the moment but possibly better. Odd how one can get accustomed to some specific sounds over time.

I noticed that the sound of  sinking ships is not played (tested with https://github.com/civctp2/civctp2/pull/314/commits/5d2894388256d9b2d78ac17bf0afed5c7dfacaef).

> BTW: I found a special pillaging-sound that could be specified in Orders.txt. Unfortunately Orders.txt expects an integer as sound-identifier which is very hard to determine. In other places a string-identifier is used. I will change that in a separate pull-request.

That sounds very interesting, looking forward to that.

williamd67:
> I noticed that the sound of sinking ships is not played.

Can you give me the context when this happened? Was it in battle-view? What kind of ship was it? As I encountered a situation in battle-view where the sound was played; the ship was an aircraft-carrier. Can a ship sink in another way than in battle?

LynxAbraxas:
Just as a note: WorkerActor (https://github.com/civctp2/civctp2/pull/314/commits/c4e6c0f489e2002f29083f7ad031a8f7042b9827) might be a remainder of CTP1, see e.g. #247.

LynxAbraxas:
I also noticed that sprites now keep their last facing direction instead of returning to a default facing direction. It feels also a bit unusual but fine and would resolve what @MartinGuehmann once considered as a bug: https://github.com/civctp2/civctp2/pull/244#issuecomment-573459550.

williamd67:
> Just as a note: WorkerActor (c4e6c0f) might be a remainder of CTP1, see e.g. #247.

I saw it was not used but still kept it alive. It may be revived once.

williamd67:
> Yes, when bombarded (sorry for missing to mention that). There is no battle-view then.

Thanks. I tried with master and indeed there a sinking sound is played. I noticed that both in this branch as in master the defender is not rendered and the `death` animation is also not played. I will investigate that and if possible, fix it.

williamd67:
> I also noticed that sprites now keep their last facing direction instead of returning to a default facing direction.

Yes I had read the comment on #244 and (planned to) fix it in this branch (almost done; still cleaning up some code after refactoring).

Personally I think the picture comes more alive when units seem to go somewhere and the movement of the next turn is more natural.

williamd67:
During debugging the health-bar rendering (UnitActor) I noticed that the bottom of the shield and medal did not have the correct color. After investigation I found that the last line of a TGA-pictures were not translated from RGB555 to RGB565 (off by one error). Impact seems to be low as I did not see any other distortion.

Screenshot from 2020-06-09 20-49-09
Screenshot from 2020-06-09 21-02-34

LynxAbraxas:
Many thanks @williamd67 for this great work. I've tested the commits upto 3f76fd799deb154fc8ef5179d2afa271ecf25bdd, and it is working very well, ships when sinking due to bombardment not only play the sound but also show the animation now, which I did not see before. Does that also work for air units and active air defense? There are quite a few sounds and animations now that I did not observe before, this is really cool!
What I did not test so far is if the special action animation are still played even in FOW, #299.
The good animation seems to repeat after 100 frames (longes possible animation sequence so far). Before it used to pause for around the same time (ca. 10sec. or so). Would it be possible to introduce that again? At the moment the continuous animation distracts a bit , however I would not want to switch it off completely. Since PC power has increased much within the last 20 years, I wonder if we should make good animation the default (currently it is off by default).

williamd67:
Thanks for your kind words.
> Does that also work for air units and active air defense?

I would expect so as it is a generic solution with no knowledge about the kind of unit.
>GoodActor: Would it be possible to introduce that again?

Absolutely, it was not my intention to change that. My mistake is probably that I start counting the pause-time at the start of the animation in stead of at the end.
We can adjust the algorithm in case we like/want that. Currently it waits for 8 seconds + random(0-5 seconds) (this was also the original code). We could make the pause time dependent on the length (number of frames) of the animation per example.

LynxAbraxas:
Since I just had my very first successful space transport (https://github.com/civctp2/civctp2/issues/229#issuecomment-643679008), I wonder if the space launch has a special animation/effect, or does the unit just silently disappear and then reappears?

williamd67:
>  I wonder if the space launch has a special animation/effect

Unfortunately nothing was done when a space plane was launched (it was only hidden). I found a space launch sound but no animation.

I implemented the following: when a space plane is launched, it will animate (move)  one tile in the direction of the destination and will play the space launch sound. After that the plane is hidden. Let me know what you think about this. As always any feedback is welcome ;)

williamd67:
There are no open issues any more on my todo-list. The classes Director, Action, Anim and different actors are refactored or cleaned (removed unnecessary methods and members). The removed members were or only read and never written, or never read and only written.

The list of improvements is part of this communication. Feel free to post any feedback on the functionality or the code.

LynxAbraxas:
> I implemented the following: when a space plane is launched, it will animate (move) one tile in the direction of the destination and will play the space launch sound.

That's great! Tested that and it works great. 

https://github.com/civctp2/civctp2/pull/314/commits/4d39a92307ba1f38d383c092a7dfd80815418a66 solves an issue that was annoying me for a long time but never found the time to investigate, thanks, works as well.

LynxAbraxas:
I have the impression that the map does not focus on units/armies bombarding a target. Could that be added?

LynxAbraxas:
Currently (with https://github.com/civctp2/civctp2/pull/314/commits/10b901cf9cfcb479c95ee43394304a9407e23f84 included) I see the special attack animation moving from the attacker pos to the target pos. While this is quite fancy, it does reveal the position of the attacker, even if the attacker unit itself is invisible, like a slaver or lawyer. So either the animation should only be shown if the unit is visible to the victim or it should be shown stationary only at the target pos.

williamd67:
> So either the animation should only be shown if the unit is visible to the victim or it should be shown stationary only at the target pos.

Good point. I propose the following behavior (please comment):
1. attacker is invisible: sound is played, animation is played stationary (on target); this is nice so that player knows something happened but does not know who caused it;
2. attacker is visible: sound is played, animation is played from attacker;  so that it is clear which attacker performed the action.
3. map stays centered on the animation when player is involved (currently not the case); see also my remark #299 [comment](https://github.com/civctp2/civctp2/issues/299#issuecomment-644410112); else it may move to other events (?)

BTW: I got this behavior (animation from attacker-position) for free when I cleaned the code (which means it was already there but not executed).

williamd67:
> I have the impression that the map does not focus on units/armies bombarding a target. Could that be added?

I will investigate that and add if possible.

LynxAbraxas:
> Good point. I propose the following behavior (please comment):
> 
>     1. attacker is invisible: sound is played, animation is played stationary (on target); this is nice so that player knows something happened but does not know who caused it;
> 
>     2. attacker is visible: sound is played, animation is played from attacker;  so that it is clear which attacker performed the action.
> 
>     3. map stays centered on the animation when player is involved (currently not the case); see also my remark #299 [comment](https://github.com/civctp2/civctp2/issues/299#issuecomment-644410112); otherwise it may move to other events (?)

Yes, that sounds very nice. The visibility of a unit would be related to #315, but should be covered all right if `m_visibility` is the base for determining visibility.
I also like the idea of point 3, if it otherwise should move to other events I guess needs to be tested.

williamd67:
This commit should implement 1 and 2. Item number three needs a little bit more thinking and discussion. I propose to do that item in a separate PR.

williamd67:
> I have the impression that the map does not focus on units/armies bombarding a target. Could that be added?

@LynxAbraxas: I have difficulty to reproduce this scenario. As it seems that bombarding cannot be scheduled like other special attacks. Can you elaborate in which scenario you want this behavior? I assume that you mean 'center map' with focus?

LynxAbraxas:
> Can you elaborate in which scenario you want this behavior?

Well, if a unit (e.g. Catapult, Artillery or Battleship) is set to bombard a city the bombard order is kept over turns until changed. So possibly at the beginning of a turn such a standing order is executed but the map does not center on this event. Similar if ending a turn before all order are set/executed.

LynxAbraxas:
> This commit should implement 1 and 2.

That's great, I'll test.

> Item number three needs a little bit more thinking and discussion. I propose to do that item in a separate PR.

Sure, as you see fit.

LynxAbraxas:
Yes, 1 and 2 are nicely covered with the new commit.
I observe that sometimes the special action is not animated but the sound is played in e.g. this save-game [noAniButSound.zip](https://github.com/civctp2/civctp2/files/4808609/noAniButSound.zip), it can also serve for tests concerning 3, since there are a few expell|pillage animations which do not seem to be related to visible units.

MartinGuehmann:
If you could cherry pick 1d4555acb2359aced2ceb2ccb60f2b9624ab620c, so that it also compiles with Visual Studio.

williamd67:
@MartinGuehmann Thanks for fixing the Visual Studio build.

LynxAbraxas:
Works great! I'd say it's ready for merging.

MartinGuehmann:
Then let's merge.

LynxAbraxas:
Sorry, just noticed a possible issue: When doing a happiness/advertizing attack with a corporate branch, the special effect animation is played even if the attack is not successful. While this can be regarded as correct, since the animation represents the attack, it appears a bit odd to me to see it in case of failure. However that might be due to what I was used over the years.
Another observation: It seem units, cities, goods are drawn multiple times (possibly infinity) when the game is scrolling or the mouse pointer is outside the game window and the game reached the upper or lower scroll limit. This is especially visible for city force fields where the redrawing has an effect like an animation where the opacity of the force field continuously increases.

williamd67:
> It seem units, cities, goods are drawn multiple times (possibly infinity)

That sounds like a bug. I will look into that and try to fix it (if not possible, rollback the changes).

> When doing a happiness/advertizing attack with a corporate branch, the special effect animation is played even if the attack is not successful.

I will investigate this case and see if the code diverts from the other special attacks. As far as I know the attack animation is not played (in other cases) when it is not successful.

LynxAbraxas:
Yes, solved with #326 and #325. Many thanks @williamd67!

> EDIT: I found another place where the success animation was played before it was tested if the action was successful.

I guess you mean the Terrorhack supposingly done by a cyber ninja
https://github.com/civctp2/civctp2/blob/611eb8ec0c306ee6bdb5cc00464e5b83b7167fd6/ctp2_code/gs/gameobj/ArmyData.cpp#L2667-L2681
It seems that ability war removed, or I was not able to find out how to get it, not sure if that happened from CTP1 to CTP2 or later, it might be related to the removal of the subneral-ad unit, which originally was meant to cause unhappiness and the the advertising attack of corporate branches was meant to drain gold, similar to a conversion I think, at least the message and the original docs suggests so:
https://github.com/civctp2/civctp2/blob/611eb8ec0c306ee6bdb5cc00464e5b83b7167fd6/ctp2_data/english/gamedata/info_str.txt#L261

LynxAbraxas:
Another change I only noticed now is that some units before this PR finished their move with a special animation, like the moray striker, which kind of submerged under the water surface at the end of its move. This seems to be bound to the move direction south-east, so it does not happen for the other directions which the unit can end its move in now. @williamd67 do you see a way to make this possible at least for south-west which should just be a mirror of south-east. Another unit where I observed this is the plasma destroyer and the nuclear sub but there might be other units as well. In contrast the fusion tank seems to settle in any direction.

williamd67:
> do you see a way to make this possible at least for south-west which should just be a mirror of south-east?

Yes that can be done (see #328). As the idle animation is only available in south-east direction, other directions are unfortunately not possible.

> In contrast the fusion tank seems to settle in any direction.

I assume that this is part of the move animation and therefor is not restricted to the south-east direction.

LynxAbraxas:
Just noticed that a franchise attack seems to initiate both the moving and the stationary special animation.
And while testing #330 I noticed that apparently formerly a conversion attack played the special animation twice, which might have been a mistake then, but it was timed so nicely that it  had a cool effect of prolonging this very cool special attack animation (I might be biased here, because it is my favorite special attack animation;-)), any way neither is critical at all.

williamd67:
> Just noticed that a franchise attack seems to initiate both the moving and the stationary special animation.

PR (https://github.com/civctp2/civctp2/pull/340/commits/62ed220a36eee427186aeefd58ee936e7692b52e) fixes the two franchise animations.

There is no generic way to prolong the conversion animation. Sorry.

LynxAbraxas:
> There is no generic way to prolong the conversion animation. Sorry.

No problem, I do wonder though why there used to be two (or more, I guess similar to #340, just without the movement).

Issue #238, opened 2020-01-05
Global Sat wonder should not reveal civilian stealth units like Lawyers


LynxAbraxas:
While #210 represents a bug issue, this is a suggestion for a feature issue: 
If the Global Sat wonder would not reveal civilian stealth units, the possessing civ would not be given such a drastic advantage and would still be in the need to use its special units to reveal civilian enemy stealth units (like Lawyers, Corporate Branches, Spies). Currently, with Global Sat it is far too easy to eradicate all enemy Lawyers, Corporate Branches as soon as they show up outside cities.

LynxAbraxas:
Solved with #315.

LynxAbraxas:
NB: New cities are not revealed by GlobSat:

![ss_2020-07-07_00:34:43](https://user-images.githubusercontent.com/21012234/87877989-c2212c80-c9e1-11ea-9eb0-8bdec7b2c4af.png)
![ss_2020-07-07_00:36:29](https://user-images.githubusercontent.com/21012234/87877991-c4838680-c9e1-11ea-99f3-4ef1ec29beec.png)

PR #315, opened 2020-05-25
Let GlobeSat only reveal some units 

LynxAbraxas:
PR to realize #238, with the basic idea that civilian units cannot be distinguished from ordinary civilians, so the Global Sat wonder cannot reveal these.

### DIFF
--- a/ctp2_code/gs/gameobj/Army.cpp
+++ b/ctp2_code/gs/gameobj/Army.cpp
@@ -888,6 +888,11 @@ bool Army::IsCivilian(void) const
 	return GetData()->IsCivilian();
 }
 
+bool Army::IsInvisibleForGlobeSat(void) const
+{
+	return GetData()->IsInvisibleForGlobeSat();
+}
+
 bool Army::IsWounded(void) const
 {
 	return GetData()->IsWounded();
--- a/ctp2_code/gs/gameobj/Army.h
+++ b/ctp2_code/gs/gameobj/Army.h
@@ -308,6 +308,7 @@ public:
 
 	bool CanTransport(void) const;
 	bool IsCivilian(void) const;
+	bool IsInvisibleForGlobeSat(void) const;
 	bool IsWounded(void) const;
 
 	bool ExecutedThisTurn() const;
--- a/ctp2_code/gs/gameobj/UnitData.cpp
+++ b/ctp2_code/gs/gameobj/UnitData.cpp
@@ -1869,7 +1869,7 @@ void UnitData::BeginTurnVision(PLAYER_INDEX player)
 		m_temp_visibility &= ~(1 << player);
 		m_temp_visibility_array.NextTurn(player);
 		m_temp_visibility |= m_temp_visibility_array.GetCurrentVisibility(player);
-		if(g_player[player]->m_hasGlobalRadar && m_army.m_id && !m_army.IsInvisibleForGlobeSat() && GetDBRec()->GetVisionClassStandard()) {
+		if(g_player[player]->m_hasGlobalRadar && m_army.IsValid() && !m_army.IsInvisibleForGlobeSat() && GetDBRec()->GetVisionClassStandard()) {
 			m_temp_visibility |= (1 << player);
 			m_visibility |= (1 << player);
 		}
@@ -2547,7 +2547,7 @@ void UnitData::DoVision(UnitDynamicArray *revealedUnits)
 	if(!IsCity())
 	{
 		m_visibility = (1 << m_owner);
-		if(m_army.m_id && !m_army.IsInvisibleForGlobeSat()) // GlobeSat does not reveal all units (i.e. some units cannot be distinguished from ordinary civilians like the common population or traders)
+		if(m_army.IsValid() && !m_army.IsInvisibleForGlobeSat()) // GlobeSat does not reveal all units (i.e. some units cannot be distinguished from ordinary civilians like the common population or traders)
 		    m_visibility |= g_theWonderTracker->GlobeSatFlags();
 		m_radar_visibility = 1 << m_owner;
 		m_ever_visible |= m_visibility;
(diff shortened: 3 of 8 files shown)


LynxAbraxas:
So far this leads to a segfault on `IsCivilian` of 
https://github.com/civctp2/civctp2/blob/00ac5f37dcf022919b2dddc14bf8db32e52c7a89/ctp2_code/gs/gameobj/UnitData.cpp#L2559
Any idea why?

MartinGuehmann:
So GlobeSat does not reveal settlers, but nuke-planting spies? I thought it was about hiding stealth units. Anyway, I would add a new flag InvisibleForGlobeSat, so that it can be fine tuned.

Actually, what is an ordinary civilian.

LynxAbraxas:
> So GlobeSat does not reveal settlers, but nuke-planting spies? 

Good point, so far civilians are:

```
awk '/{/ {a=$1}; /Civilian/ {print a}' ./ctp2_data/default/gamedata/Units.txt 
UNIT_ABOLITIONIST
UNIT_CLERIC
UNIT_CORPORATE_BRANCH
UNIT_DIPLOMAT
UNIT_EMPATHIC_DIPLOMAT
UNIT_LAWYER
UNIT_SEA_ENGINEER
UNIT_SETTLER
UNIT_TELEVANGELIST
UNIT_URBAN_PLANNER
```

So we could either make Spies civilians as well or "add a new flag InvisibleForGlobeSat". Does a new flag need many changes?

> Actually, what is an ordinary civilian.

With that I meant those that have no tactical meaning, i.e. non-units like the common population or traders. 
NB: GlobeSat does not reveal the current state of trade routes (#256), i.e. you still have to explore their status with units (which is intentional) to avoid that GlobeSat gives a too big advantage and to avoid that it would give info the AI would not be able to consider.

LynxAbraxas:
> Any idea why?

Lead to a segfault because armies might already be gone.

MartinGuehmann:
> So we could either make Spies civilians as well or "add a new flag InvisibleForGlobeSat". Does a new flag need many changes?

I do not consider anybody who plants nukes as a civilian. As a modder you might want to hide other things from GlobeSat, so it should be a new flag.

About the changes, you already have added new flags to const.txt, it is the same for unit.txt. And then you have to duplicate the code for IsCivilian.

LynxAbraxas:
> About the changes, you already have added new flags to const.txt, it is the same for unit.txt. And then you have to duplicate the code for IsCivilian.

I'll give it a try, but it will have to wait a bit, I've very little spare time at the moment.

LynxAbraxas:
While this seems to work in gerneral, Settlers, Urban Planners, Diplomats and Emphatic Diplomats are still visible even if outside any normal vision. Any ideas why that is the case?

LynxAbraxas:
Not sure why this was closed, at least it was unintentionally.

MartinGuehmann:
> While this seems to work in gerneral, Settlers, Urban Planners, Diplomats and Emphatic Diplomats are still visible even if outside any normal vision. Any ideas why that is the case?

I don't know why. But why should Settlers, Urban Planners, Diplomats, and Emphatic Diplomats be hidden from GlobeSat. They are not stealth units.

LynxAbraxas:
> But why should Settlers, Urban Planners, Diplomats, and Emphatic Diplomats be hidden from GlobeSat. They are not stealth units.

2 reasons:
1. They are not stealth but also do not appear militaristic, i.e. in a high-res sat image they would look like normal people. That's why I decided to take out the sea-engineer. 
2. In the scope of the game GlobeSat before this PR is giving so much advantage to the owning civ, that it is hardly possible to still challenge that civ, in addition many advantages seem not usable for the AI (e.g. fly lawyers to every corporate branch popping up on the map to sue it before it can do any attack at all), so in turn it can be seen as a too big advantage for a human player. So all modifications here are meant to balance the game a bit more.

Any way I'm happy to adjust this as you see fits best, but before I'd like it to work in general, i.e. also for Settlers, Urban Planners, Diplomats, and Emphatic Diplomats.

LynxAbraxas:
Found another ref
https://github.com/civctp2/civctp2/blob/24d98c14ed163e80a1789ab0892a478f284d7133/ctp2_code/gs/gameobj/Player.cpp#L5767-L5784
that eventually seems to lead to the GlobSat flags and might have an effect on the visibility of units, which I tried to adjust accordingly in e760bf17cb5b299f4bf559b6f0beb1091d5f7cd0 but the game segfaults that way during the evaluation towards the new turn.

MartinGuehmann:
> Any way I'm happy to adjust this as you see fits best, but before I'd like it to work in general, i.e. also for Settlers, Urban Planners, Diplomats, and Emphatic Diplomats.

Well, you could argue that you could see 10000 or 30000 people from space. In any way, I would treat Settlers, Urban Planners, and Sea Engineers the same.

Diplomats are very visible to foreign states, more visible than citizens. However, it should be more related to the game concepts, like small stealth units. That's a bit easier to grasp as a player, I think.

MartinGuehmann:
> that eventually seems to lead to the GlobSat flags and might have an effect on the visibility of units, which I tried to adjust accordingly in [e760bf1](https://github.com/civctp2/civctp2/commit/e760bf17cb5b299f4bf559b6f0beb1091d5f7cd0) but the game segfaults that way during the evaluation towards the new turn.

Actually, in this case you should use `m_army.IsValid()`. `m_army.m_id` can be anything, either because it is uninitialized or it is an ID of an army that does not exist anymore.

LynxAbraxas:
Yes, true, thanks @MartinGuehmann for that hint, changed accordingly.

> but the game segfaults that way during the evaluation towards the new turn.

Was not due to that change but due to https://github.com/civctp2/civctp2/pull/311#issuecomment-651999928

MartinGuehmann:
Then let's merge this, too.

PR #316, opened 2020-05-31
REJECTED: Technical: Decrease complexity of ifs that contains GetVisiblePlayer

williamd67:
Index of player, returned by GetVisiblePlayer, is used in three major ways:
1. to fetch the visible player pointer
2. to test if an index equals the visible player index
3. to test if something is visible
for all three cases a method has been introduced.

There is also some cleanup in whitespace (especially empty lines I encountered).

### DIFF
--- a/ctp2_code/gs/slic/SlicEngine.cpp
+++ b/ctp2_code/gs/slic/SlicEngine.cpp
@@ -24,15 +24,15 @@
 //
 // Modifications from the original Activision code:
 //
-// - Readded the new slicfunctions of the patch, by Martin Gühmann.
+// - Readded the new slicfunctions of the patch, by Martin GÃ¼hmann.
 // - Added FreeAllSlaves slicfunction by The Big Mc.
 // - Added PlantSpecificGood and RemoveGood slicfunctions by MrBaggins.
 // - Data blanked out at screen in between 2 players in hotseat play.
 // - Great library history cleared between 2 players in hotseat play.
 // - Corrected reported memory leak.
 // - Added database access to all databases in the new database format,
-//   even if it does not make sense, by Martin Gühmann.
-// - New slic functions added by Martin Gühmann:
+//   even if it does not make sense, by Martin GÃ¼hmann.
+// - New slic functions added by Martin GÃ¼hmann:
 //   - CargoCapacity     Gets number of additional units a unit can carry.
 //   - MaxCargoSize      Gets the maximum number of units a unit can carry.
 //   - CargoSize         Gets the current number of units a unit is carrying.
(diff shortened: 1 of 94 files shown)


williamd67:
During my work on Director I encountered regularly places where GetVisiblePlayer was indirectly used. First the index was fetched and than something was done with the index. Especially the visibility calculation `actor->GetUnitVisibility() & (1 << g_selected_item->GetVisiblePlayer())` was regularly duplicated and I wanted it to be hidden in a method.

So I introduced four difference methods, all part of g_selected_item:
1. GetVisiblePlayer that directly returns the pointer to the visible player
2. IsVisiblePlayer(index) that test if index equals the visible player index
3. IsUnitVisible(unit) which test if a unit is visible by the visible player
4. IsPosVisible(pos) which test if a position is visible by the visible player

To have a clear way of working I renamed GetVisiblePlayer to GetVisiblePlayerID and checked all places it was used. In case it was indirectly used I replaced it with one of the four methods.

Unfortunately it was much more work and time than I expected and more changes as well. I still pushed it through and I hope it has enough value to go to master. I think it is easier to read now. I still have ideas to improve it further.

Let me know your opinion.

ctp2_code/gs/slic/SlicEngine.cpp
```diff
@@ -953,37 +940,29 @@ void SlicEngine::AddBuiltinFunctions()
 	m_functionHash->Add(new Slic_GetMostAtRiskCity);
 	m_functionHash->Add(new Slic_GetRoundsToNextDisaster);
 	m_functionHash->Add(new Slic_GetCurrentPollutionLevel);
-	// New slicfunction by The Big Mc
```
MartinGuehmann, line 956: These slic function below are not from CTP2.1, so you may prune the comments but you still need to make that clear that you do not find those in Activision's patch.

MartinGuehmann:
Any comments about this one? Otherwise, I would just check whether it compiles on Windows and merge then if it does, of course once the merge conflict have been fixed.

williamd67:
Nice, in that case I will resolve the merge conflicts.

MartinGuehmann:
I tried to compile it on Windows, that did not work without code code modification. And then I tested it. The AI on screen seemed not to behave as without this. My impression was that it wouldn't explore and bring ships of the coast.

That is a bit unspecific. And the code changes are to many to find the problem. Usually, when I did something like this on the code I ended up of redoing it.

So I would start with cleaning up white space and then I would rename GetVisiblePlayer() to GetVisiblePlayerID(), and whatever else needs to be renamed. This can be done automatically. And then go for function replacing in the functional way, one by one. Maybe I would also not edit too many files for a commit so that I still have a change to nail it if something goes wrong.

williamd67:
>  And then I tested it.

@MartinGuehmann thanks for the effort of testing this. Unfortunately it did not work.

> Usually, when I did something like this on the code I ended up of redoing it.

I agree with your approach as there are so many changes it has to be faultless or rejected. As you found a single error, there may be more and it will be hard to proof otherwise with this all or nothing PR.

I will redo the changes one for one so it may be possible to find the incorrect one. I will also use the current branch to compare the code against as in theory the new and old branch should be equal at the end (except for the found issues). For that reason I will keep this branch sync with master.

williamd67:
There were too many changes that were hard to test. PR is rejected and closed.

PR #317, opened 2020-06-07
Show tooltip when idling mouse over panel button (SDL2-regression)

williamd67:
This will fix the issue that a tool-tip was not shown any more when idling the mouse over a panel button; per example unit-actions such as `Pillage`.


LynxAbraxas:
Yes, works again:
ss_2020-06-07_19:07:37

MartinGuehmann:
Good then we can merge.

PR #318, opened 2020-06-13
Technical: fix memory leak (sort of) while saving game

williamd67:
During saving the game an extra UnseenInstallationInfo-record was added to every unseen cell to store the city owner. This record was not removed.

This caused additional memory to be allocated after every save game (cumulative). (Re-)loading the game removed this additional memory. As the record was stored in a list it was correctly deleted when quitting the game.

This was most noticeable due to auto-save and playing for a long time without reloading or quitting the game.

A side-effect was that the save-game size increased as well (cumulative). This was also resetted when the game was reloaded and saved again.

As in bigger games the number of unseen-cells can be significant (my game had around ~100k unseen cells) the amount of allocated memory grew also significantly ~4Mb per turn.

### DIFF
--- a/ctp2_code/gs/world/UnseenCell.cpp
+++ b/ctp2_code/gs/world/UnseenCell.cpp
@@ -25,18 +25,18 @@
 // Modifications from the original Activision code:
 //
 // - Finished tile improvements are now stored as well in the improvements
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
Indeed that piece of code is missing. I do not expect any problems with this so let's merge.

PR #319, opened 2020-06-14
Corrections to unit dependencies

LynxAbraxas:
Formerly, it was possible that thanks could not be built any more even though fusion tanks were not available yet. In contrast cavalry was not deprecated before fusion.
I also noticed that leviathons could already be built with unified physics whereas they should only become available with plasma weaponry.

### DIFF
--- a/bin/deprecateAdv.awk
+++ b/bin/deprecateAdv.awk
@@ -1,16 +1,16 @@
 # read a Great_Library.txt
 # when a  UNIT_.*_PREREQ is found use the info to extract ObsoleteAdvance from Units.txt into variable OA
-# insert that info 2 lines after "Requires:"
+# insert that info 1 lines after "L:DATABASE_ADVANCES,ADVANCE_" (which is language indep.)
 
 match($0, /.*(UNIT_.*)_PREREQ.*/, s) {
     cmd = sprintf("awk -v var=%s '/{/ {a=$1}; /ObsoleteAdvance/ { if ( a ~ var ) print $2}' ./ctp2_data/default/gamedata/Units.txt", s[1])
     cmd | getline OA
     close(cmd)
     }
-/Requires:/ {
+/L:DATABASE_ADVANCES,ADVANCE_/ {
     found=1
     }
-(found++) == 3 && OA {
+(found++) == 2 && OA {
     printf("Obsolete:\n<L:DATABASE_ADVANCES,%s>%s<e>\n", OA, OA)
     OA=0
     } 1
--- a/bin/deprecateAdv.awk
+++ b/bin/deprecateAdv.awk
@@ -3,7 +3,7 @@
 # insert that info 1 lines after "L:DATABASE_ADVANCES,ADVANCE_" (which is language indep.)
 
 match($0, /.*(UNIT_.*)_PREREQ.*/, s) {
-    cmd = sprintf("awk -v var=%s '/{/ {a=$1}; /ObsoleteAdvance/ { if ( a ~ var ) print $2}' ./ctp2_data/default/gamedata/Units.txt", s[1])
+    cmd = sprintf("awk -v var=%s '/{/ {a=$1}; /ObsoleteAdvance/ { if ( a ~ var ) print $2}' ../../default/gamedata/Units.txt", s[1])
     cmd | getline OA
     close(cmd)
     }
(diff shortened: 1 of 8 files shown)


LynxAbraxas:
Checking the source for this, I found that `ctp2_data/default/gamedata/Units.txt` apparently did not exist before https://github.com/civctp2/civctp2/commit/dee3da6e5edea506778802e7631ca3c0cb155c61:
https://github.com/civctp2/civctp2/commits/99e74c4dfb7a4af6ece4412296e0b2cc20cc9af0/ctp2_data/default/gamedata/Units.txt

@MartinGuehmann do you know why it came in there and how it worked before? Was Units.txt constructed manually?

MartinGuehmann:
> @MartinGuehmann do you know why it came in there and how it worked before? Was Units.txt constructed manually?

What do you mean by "constructed manually"? It comes from one of the first commits. Originally, these weren't even commits but a set of modified files zipped up and attached to a forum post on Apolyton. We had no version control system. Our idea was to go for SourceForge at that time, however it wasn't clear whether Activion's EULA was compatible with SourceForge.

As modders we just included the files we modified, and so we added a file when it modified. That's why we don't have the complete set of Japanese text files and that's why we added some English text files for the scenarios, recently. Also the person who created the repository did not put the original text files in as a base.

So if you are looking for the previous version of Unit.txt you find it on the CD. And at least you copied that part of the CD.

bin/deprecateAdv.awk
```diff
@@ -1,16 +1,16 @@
...
     cmd | getline OA
     close(cmd)
     }
-/Requires:/ {
+/L:DATABASE_ADVANCES,ADVANCE_/ {
     found=1
     }
-(found++) == 3 && OA {
+(found++) == 2 && OA {
     printf("Obsolete:\n<L:DATABASE_ADVANCES,%s>%s<e>\n", OA, OA)
```
MartinGuehmann: This isn't language independent.

bin/deprecateAdv.awk
```diff
@@ -11,6 +11,9 @@ match($0, /.*(UNIT_.*)_PREREQ.*/, s) {
     found=1
     }
 (found++) == 2 && OA {
-    printf("Obsolete:\n<L:DATABASE_ADVANCES,%s>%s<e>\n", OA, OA)
+    oa=OA
+    gsub(/ADVANCE_/, "", oa)
+    gsub(/_/, " ", oa)
+    printf("Obsolete:\n<L:DATABASE_ADVANCES,%s>%s<e>\n", OA, tolower(oa))
```
MartinGuehmann, line 17: The initials should be upper case, in fact you want to take this from the string database, anyway. Since you could name these things whatever you like. And in a localized version those don't fit anyway.

In fact, we could add a default value which would be loaded from the string database. Something like you have in Wikipeda, where you could append the link with an s for instance and it is still part of the link.

bin/deprecateAdv.awk
LynxAbraxas: Why not? `L:DATABASE_ADVANCES,ADVANCE_` has nothing language dependent.

MartinGuehmann: And what about the "Obsolete"? I wouldn't use that word in German, even so somebody claims it exists. And even if the spelling would be "Obsolet". I would translate it with "Veraltet mit:".

LynxAbraxas: I see what you mean. What I meant is that the insertion position is language independent in contrast to the former state where the insertion position was determined by the string `Requires:`. Taking such strings from a database would be nice, but I am not sure if I will find the time to realize that.

bin/deprecateAdv.awk
LynxAbraxas: I see, you mean like using
https://github.com/civctp2/civctp2/blob/11b9a024f4e0b5ace7011f6e53a49f5f7bb83651/ctp2_data/english/gamedata/gl_str.txt#L83
for the English version and otherwise the corresponding language file.

LynxAbraxas:
Instead of 0a11aaec40d2e57029fba21ee99e17c6fd6123e9, this PR would benefit from 855e0706458a8b0c0d17ee9c497099405d2629b6 and 83be17ad1d12bfde7a1c255560c4b61270f08a8a (from #321) as well.

bin/deprecateAdv.awk
LynxAbraxas: Done with https://github.com/civctp2/civctp2/pull/319/commits/c46cea6ae07a7cf67798f7d28351af965b4a28b0.

LynxAbraxas:
> As modders we just included the files we modified, and so we added a file when it modified. 

I see. It's a pity that the original files were not necessarily included in the repo in their unchanged form, but well, that's how it is. We could attach a commit at the very beginning of the repo or amend the very first, but all other commits would get a new SHA.

MartinGuehmann:
> > As modders we just included the files we modified, and so we added a file when it modified.
> 
> I see. It's a pity that the original files were not necessarily included in the repo in their unchanged form, but well, that's how it is. We could attach a commit at the very beginning of the repo or amend the very first, but all other commits would get a new SHA.

That's not really good, since we are referring to these SHAs. In principle the one who set up the repro could have included the files, but this is also long ago.

In fact you can still get the files and compare, it is just not as convenient then doing it from within the repository.  At least at some point I added most of the text files, when I figured out that I also had to edit the ones from the Japanese version. That was then when the guy who provided them was long gone.

LynxAbraxas:
> That's not really good, since we are referring to these SHAs. In principle the one who set up the repro could have included the files, but this is also long ago.

Yes, I would also not like to change the history of master any more, before PRs are merged this is fine, but on master this can cause a lot of confusion. Let's live with the repo as is.

Merge conflicts resolved with a rebase, so ready for merging from my side.

MartinGuehmann:
The build checks have passed, even so this is not reported on GitHub. Anyway, it looks good so let's merge.

LynxAbraxas:
> The build checks have passed, even so this is not reported on GitHub.

Yes, have observed this behavior a few times lately. Possibly some changes in GH backend.

PR #320, opened 2020-06-16
Allow order sounds to be defined as string-identifier

williamd67:
Order sound can be defined in orders.txt. This was done by an integer which was hard to determine. The sound identifier can now be determined by a string-identifier.

This means that you can specify special order sounds in Orders.txt by the string identifier, per example:
```
Sound SOUND_ID_PILLAGE
FailSound SOUND_ID_GENERALFAIL
```
In case this sound would be declared in the current master it will play through the general success sound that is always played. That is solved/changed in #314.


MartinGuehmann:
What happens if you have the Sound and FailSound defined in Orders.txt? I am not sure whether any of the mods it uses, but it would be nice that none would complain if it any did.

And could you also fix the merge conflicts?

williamd67:
> What happens if you have the Sound and FailSound defined in Orders.txt?

Good question and I tested it. In case numerical identifiers are defined in Orders.txt, it will still work as the implementation of SoundRecord is smart enough to support both numerical identifiers as string identifiers so current implementation is backwards compatible.

> And could you also fix the merge conflicts?

Done

MartinGuehmann:
Then we can put this in.

PR #321, opened 2020-06-20
Debug reentry of space plane

LynxAbraxas:
PR to solve #229.

### DIFF
--- a/Scenarios/AE_Mod/scen0000/english/gamedata/Great_Library.txt
+++ b/Scenarios/AE_Mod/scen0000/english/gamedata/Great_Library.txt
@@ -9155,7 +9155,7 @@ Can <L:DATABASE_ORDERS,ORDER_SPACE_LAUNCH>Space Launch<e>
 [END]
 
 [UNIT_SPACE_PLANE_GAMEPLAY]
-The Space Plane is a long-range suborbital <L:DATABASE_ORDERS,ORDER_BOARD_TRANSPORT>Transport<e> vehicle.  When launched from a city, it can <L:DATABASE_ORDERS,ORDER_BOARD_TRANSPORT>Transport<e> up to five units to any other city on the map.  The <L:DATABASE_ORDERS,ORDER_BOARD_TRANSPORT>Transport<e> takes 4 turns to complete.  Although not well suited to transporting short distances, the Space Plane is ideal if you need to quickly get units the other side of the map.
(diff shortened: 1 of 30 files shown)


LynxAbraxas:
Reentries of space planes now work even if the game is saved and loaded while a space plane is in space.

LynxAbraxas:
A space lift only takes 3 turns with the current implementation:
https://github.com/civctp2/civctp2/blob/11b9a024f4e0b5ace7011f6e53a49f5f7bb83651/ctp2_data/default/gamedata/Units.txt#L3530
https://github.com/civctp2/civctp2/blob/11b9a024f4e0b5ace7011f6e53a49f5f7bb83651/ctp2_code/gs/gameobj/ArmyData.cpp#L1735
which is more appropriate than 4 turns as stated in the GL, so c1e0e1eb2907ec1b68394ba6ef83f5b195e73e94 changes the GL files accordingly. However for awk to work properly for all language files, also TXT files need to be normalized to use LF as EOL (855e0706458a8b0c0d17ee9c497099405d2629b6 and 83be17ad1d12bfde7a1c255560c4b61270f08a8a), which needs to be tested if that has any negative effect on windows. @MartinGuehmann could you do that?

LynxAbraxas:
I wonder if we should reduce the move points of a space plane. Currently it can move so far in one turn that it often makes little sense to use a space launch unless one wants to pass over enemy terrain. For non-gigantic maps the use of a space launch reduces further. Considering that it seem comparable to the NASA Space Shuttle, its ability to fly from one city to another on its own is not very realistic any way and in case of saying that it gets transported by a carrier, it would be realistic to say that this can be done only with very few move points per turn, while possibly keeping or adjusting the fuel points.

LynxAbraxas:
Halfed move points to better fit the concept also described in the GL:
`Although not well suited to transporting short distances, the Space Plane is ideal if you need to quickly get units the other side of the map.`
https://github.com/civctp2/civctp2/blob/3294daaf778dc02472a16cdf4252240f67cc9190/ctp2_data/english/gamedata/Great_Library.txt#L9030-L9032

LynxAbraxas:
Tested this for some time now and seems to work as expected now. @MartinGuehmann ready for merge?

MartinGuehmann:
> which is more appropriate than 4 turns as stated in the GL, so [c1e0e1e](https://github.com/civctp2/civctp2/commit/c1e0e1eb2907ec1b68394ba6ef83f5b195e73e94) changes the GL files accordingly.

Why do you hard encode the value and don't use {UnitDB(UnitRecord[0]).SpaceLaunch}?

LynxAbraxas:
> Why do you hard encode the value and don't use {UnitDB(UnitRecord[0]).SpaceLaunch}?

Because it was hard encoded before;-) but I like the idea and changed the commit accordingly.

LynxAbraxas:
Works as expected.

MartinGuehmann:
Then we can merge.

PR #323, opened 2020-06-21
Adjust listing of the trade manager

LynxAbraxas:
So far the list of possible trade routes is based on first to fifth route yielding the best price for each good owned. Since this is not necessarily the best choice due to high cost of trade points, this PR changes the list to show the first fife best routes concerning maximum value per trade point, similar to what the AI now uses since #252.
It makes it much easier to maximize the total trade income with the given number of caravans but still this is a limited amount of the total number of all possible trade routes. So the trade manager is still lacking some means to create an arbitrary trade route, which can be of special interest for domestic trade, e.g. when one wants to boost a newly created city, no matter how many trade points those routes cost.
This could be realized with e.g. a scroll list of all possible source cities (all owned cities that have at least one good) and another scroll list with all other own|specific civ|friendly|all (not at war and not embargo) cities. A drop down list (as uses in some places of the game) would probably not be so suited in case there are many entries in the list.

### DIFF
--- a/ctp2_code/ui/interface/trademanager.cpp
+++ b/ctp2_code/ui/interface/trademanager.cpp
@@ -417,7 +417,7 @@ void TradeManager::UpdateCreateList(const PLAYER_INDEX & player_id)
 
 						for(i = 0; i < m_numCities; i++) {
 							if(valuePerCaravan > maxValuePerCaravan[i]) { // determine best offer
-								for(j = m_numCities - 1; j>= i; j--) {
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
@MartinGuehmann, @williamd67 what do you think of this? It's around for 2 year now. It works as expected, so could go in if wanted, otherwise I would close the PR.

MartinGuehmann:
This can also go in.

Issue #74, opened 2019-01-28
Documentation for newly implemented game concepts


LynxAbraxas:
In relation to #49 (https://github.com/civctp2/civctp2/issues/49#issuecomment-457877924):

Do we have some place where newly implemented game concepts are documented? 
If not, additions from mods should be documented in a kind of manual to not loos track of what has been done and to avoid that the code is the only "documentation".

MartinGuehmann:
I don't think so. And what do you call a mod. I would restrict it to modifying game files, but not source files. We added a few things from classical mods, that are most prominently CityMod and GoodMod. We did not add things from the slic files, since the idea was that this stuff should be implemented in the source.

However, there are still a few things we could include, I thing the stuff from map.txt concerning world generation would be a good addition.

Other things are bit questionable, the people tweaked the priorities in the goals for attack and siege hoping that they would improve the AI. However, that does not help if the logic in the code is flawed.

Beside that #49 addresses something changed by Activision, so if it all then it would be in the official manual if at all.

LynxAbraxas:
Hm, then what is this EMOD that pops up in the code?
Improvements to the AI are not necessarily needed by a gamer, but new options/behaviours like #73 would do well.

> Beside that #49 addresses something changed by Activision, so if it all then it would be in the official manual if at all.

Agree, but if activision missed it, our documentation for new stuff could also include a section like "explanations to activision manual and things that are missing in the original manual"

MartinGuehmann:
> Hm, then what is this EMOD that pops up in the code?

That are modifciations by E (his user name on Apolyton, ekmek on CivFanatics). So he added quite some stuff to the game. However, he is not a programmer, so the things are half done. And I eded up to fix the things or care that they are not in the way.

> Improvements to the AI are not necessarily needed by a gamer, but new options/behaviours like #73 would do well.

That depends, originally one of the big complains about the game was the weak AI, and that was one of our issues:

Fix the AI

Sounds simple but a game AI is quite complex so there are many issues. It least it should be a bit challenging, an AI that you can beat at every wonder is a bit dull. In principle, every game has an AI that is flawed in one respect or in another. Simply because AIs are complex, so you have a lot of chances to screw up.

Of course, a game that runs faster and needs less memory is useful for the gamer, especially if it does not crash because it runs out of memory.

But you are right, we need also a few more goodies for the next release.
 
[...]

PR #324, opened 2020-06-23
Soften smoothen align city information

williamd67:
The purpose of this branch is to soften, smooth-en and align the city-information on the map.

I will create some examples (screenshots) of my idea that can be discussed.


williamd67:
This is first attempt to align the city information. I also made the color of the numbers depending on the color of the player. Dark color gets a text in white, light color gets a text in black. The same is done for the stacking indicator of a unit.

I updated some icons to be able to use the color of the player as frame.

Improvements in this commit are:
* text on light color in black (single draw)
* text in dark color in white (single draw)
* also for shields of units

* capital icon next to name
* sleeping-units icon next to name
* watchfull icon next to name (in color of player)
* city-name does not show empty-queue (yellow); when production is shown (production will show empty-queue)

* capital icon darkened under fog
* sleeping-units darkened under fog (will never happen as only active player shows sleeping units)
* watchful icon darkened under fog
* all other icons darkened under fog (not only player color)
* if needed is the player color darkened (locally) to make the numbers more readable (converted and franchised)
* enslavement icons are shown on top of the name in stead of next of all the other icons

* share frames between elements (single line in stead of double lines) in color of player

* isRioting icon is not shown as it can be determined by the color of the city name (red)

williamd67:
Examples of the new city information rendering:
Screenshot from 2020-06-24 00-41-10
Screenshot from 2020-06-24 00-41-30
Screenshot from 2020-06-24 00-42-48

Screenshot from 2020-06-24 00-53-26
Screenshot from 2020-06-24 00-54-07
Extreme cases (faked):
Screenshot from 2020-06-24 00-47-16
Screenshot from 2020-06-24 00-47-57

Unit with black and white stack-indication:
Screenshot from 2020-06-24 00-57-32
Screenshot from 2020-06-24 00-59-33

williamd67:
I will try your suggestions and post screenshots. Difficulty with not having a black-name box is that we may change the name-color to yellow and red, I expect that that is not always good contrast. Having said that I can also make some part of the box yellow/red to show the state. I will also play with this.

I will try to make it transparent as well. I do expect that it will be harder to read names and numbers.

williamd67:
I think rendering the frames in the same color as owner who caused the city icon, is a winner. It looks indeed better in my opinion.
Screenshot from 2020-06-25 23-26-48
Screenshot from 2020-06-25 23-25-14

LynxAbraxas:
Looks great, will test it.

> I will try to make it transparent as well. I do expect that it will be harder to read names and numbers.

True, probably depends on the amount of transparency, which might be possible to implement adjustable with some keyboard keys, but that's only an idea. If it is hard to realize there's no need to bother with that. It is already much nicer to view in the new state.

williamd67:
@LynxAbraxas I implemented/hacked a version where the city-information is blended with the background. With `c` (was toggle city-names) you can change the alpha: from 255 to 127 in steps of 32.

Personally I do not think it is much better, although around 160 there is a sweet spot where the background shows, and the foreground is still visible. The advantage is that it softens even more as the city-information blends ;) in the background. We could make the toggle between off/on/blended with fixed alpha.

It may also be better when the whole city-box is the player-color (I will try that next).

LynxAbraxas:
> I think rendering the frames in the same color as owner who caused the city icon, is a winner. It looks indeed better in my opinion.

Yes, me too, I like it very much.

> Difficulty with not having a black-name box is that we may change the name-color to yellow and red, I expect that there is not always a good contrast with the player color.

True, and some colors do not go well together.

> I implemented/hacked a version where the city-information is blended with the background. With c (was toggle city-names) you can change the alpha: from 255 to 127 in steps of 32.
> Personally I do not think it is much better, although around 160 there is a sweet spot where the background shows, and the foreground is still visible. The advantage is that it softens even more as the city-information blends ;) in the background. We could make the toggle between off/on/blended with fixed alpha.

I like it very much, and with the opaque rendering during scrolling, I am even growing fond of lower values up to 127. So why not leave it at 5 states as it is now, it is just 2 more key strokes than having 3 states (on/off/blend).

williamd67:
> with the opaque rendering during scrolling

To be honest that is just co-incidence and not planned by me. It indeed looks nice to highlight them during scrolling.

> So why not leave it at 5 states as it is now...

I am fine to leave the 5 states in. @MartinGuehmann what do you  think?

I will improve the code and remove the hacks.

LynxAbraxas:
Playing with this for a while now and am starting to wonder if similar transparency (possibly with its own control key) might also be nice to have for the unit/army shields. Basically those and the city info are elements that do not belong the realistic looking of the map.

williamd67:
> I am fine to leave the 5 states in.

Currently there are 6 states, so I propose to skip the alpha around 220 , and support the following states:
255, 191, 159, 127, 0

LynxAbraxas:
Hm, as you see fit. 6 is also fine for me. Another option would be to give full opacity while e.g. the Shift (or Alt or Alt-Gr) key is presssed.

williamd67:
I implemented the blending of the city-information into the background properly (removed all the hacks). The selected value is also stored and reloaded. The graphics-options screen shows the correct on/off value.

Should we introduce a slider in the graphics-options screen to be able to change the transparency without knowing the key?

williamd67:
> if similar transparency (possibly with its own control key) might also be nice to have for the unit/army shields and decorations

I will put that on my todo list (seperate PR). I expect that we not only do it for the shield but also for the nation flag, medals, veteran status, etc.

LynxAbraxas:
> I implemented the blending of the city-information into the background properly (removed all the hacks). The supported alpha values are 255 (OPAQUE, ON), 191, 159, 127 and 0 (TRANSPARENT, OFF). The graphics-options screen shows the correct on/off value.

That's great and works well.

> The selected value is also stored and reloaded (backwards compatible). 

Not sure about the backwards compatibility. If I save a game with ctp2 including https://github.com/civctp2/civctp2/pull/324/commits/d465cb9f9dba21a00bd027105d2567af77291909 (this PR) and then open that with a ctp2 not including this PR I get a segfault:
```
Backtrace:
/lib/x86_64-linux-gnu/libc.so.6(+0xb16b4)[0x7f69f373a6b4]
/lib/x86_64-linux-gnu/libc.so.6(_IO_vfprintf+0x2143)[0x7f69f36e64d3]
/lib/x86_64-linux-gnu/libc.so.6(vsprintf+0x81)[0x7f69f370a6d1]
./ctp2(_Z20c3errors_ErrorDialogPKcS0_z+0x192)[0x55f0a2170020]
./ctp2(_ZN9ProfileDB5ParseEP8_IO_FILE+0x46d)[0x55f0a2593571]
./ctp2(_ZN9ProfileDB4InitEj+0x1b8)[0x55f0a259306a]
./ctp2(_ZN6CivApp13InitializeAppEP11hinstance_ti+0xad)[0x55f0a217ffd5]
./ctp2(_Z7CivMainiPPc+0xcb)[0x55f0a2179418]
./ctp2(main+0x2c)[0x55f0a21791e4]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7f69f36aab97]
./ctp2(_start+0x2a)[0x55f0a21606ea]
```
during loading, which might come from parsing e.g. the `CITY_NAMES=127` with the expectation to find yes|no.

> Should we introduce a slider in the graphics-options screen to be able to change the transparency without knowing the key?

If you feel up to that challenge, why not. Otherwise it might be enough to document that new functionality somewhere. Not sure where though, see #74. It could just be a new section in the `README.md` or a e.g. `DOC.md`, which would be less fuss than putting it in the LaTeX docs (which have not been updated for a long time) and compiling a PDF.

williamd67:
> Not sure about the backwards compatibility.

Ah, I meant that you could read older save files. I was not thinking about the opposite direction :( Let me think if I can fix that as well. It would mean introducing a new name at least.

>  If you feel up to that challenge, why not.

I will give it a try. It would mean that it is done completely.

LynxAbraxas:
>  I was not thinking about the opposite direction :( 

Well, perhaps the opposite direction is a bit exotic, basically arising from testing PRs and possibly backward compatibility was broken already at other places without us noticing.

Not sure if I can see a difference when https://github.com/civctp2/civctp2/pull/324/commits/1f0e5bf5fe70825531d3a516f4fc1510e8a70e61 is included.

williamd67:
> Not sure if I can see a difference when 1f0e5bf is included.

I do not expect so, it is for mods that have 'religion'-icons. I updated the code so that they are handled in the same ways as the other city icons. I tested it by enforcing some random icons and checked if the icons were aligned/darkened correctly.

williamd67:
Storing and retrieving the graphic-option `ShowCityNameAlpha` is now really backwards compatible (I hope).

This is my last open item for this PR. So I removed `WIP` from the title.

LynxAbraxas:
Many thanks @williamd67 for this other really great PR! I tested https://github.com/civctp2/civctp2/pull/324/commits/342e609a4305c409d0baab18327b7f34a17bc47f for a while and works nicely and is backwards compatible. I'd say good to go in. @MartinGuehmann what do you think?

MartinGuehmann:
Then we can merge.

PR #325, opened 2020-06-24
Only play advertising animation when action is successful

williamd67:
This will fix the remark of @LynxAbraxas that advertisement animations are played even when unsuccessful.


LynxAbraxas:
Sounds great, trying to test this but currently GL's DI registry tells me `certificate signed by unknown authority` for any image I try to push so I guess they are modifying something at the moment that needs some polishing...

LynxAbraxas:
Got it tested, and it works. Many thanks @williamd67 for fixing!

MartinGuehmann:
So let's merge.

PR #326, opened 2020-06-24
Not render actors while scrolling

williamd67:
As map is copied and not rendered during scrolling, rendering actors will cause the actor to be redrawn in (almost) the same position multiple times which causes side-effects. Actors are not rendered during the scrolling actions, which may cause a small jump when the scrolling stops.

I expect this to fix the issue reported by @LynxAbraxas that actors are rendered multiple times on the same location.

### DIFF
--- a/ctp2_code/gfx/spritesys/EffectActor.cpp
+++ b/ctp2_code/gfx/spritesys/EffectActor.cpp
@@ -90,10 +90,6 @@ void EffectActor::Process(void)
 
 	if (m_curAction)
 	{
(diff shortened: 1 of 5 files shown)


LynxAbraxas:
Sounds great, trying to test this but currently GL's DI registry tells me `certificate signed by unknown authority` for any image I try to push so I guess they are modifying something at the moment that needs some polishing...

LynxAbraxas:
Also works nicely, many thanks for this fix as well.

MartinGuehmann:
Then we can merge.

PR #328, opened 2020-06-28
Render idle animation also in mirrored idle facing

williamd67:
This will allow units to execute the idle animation also when facing the mirrored idle position.


LynxAbraxas:
Works nicely as well.

MartinGuehmann:
So then let's merge.

PR #329, opened 2020-06-28
WIP: Hide diplo window if AI accepted proposal

LynxAbraxas:
Currently, the diplomatic window stays open after the AI accepted a proposal, so the evaluation of ther AI moves continues, mostly behind the still open window. This PR is meant to close the window when the AI has accepted the proposal. In case the AI rejects this problem does not arise because the game does not continue directly, instead the human player is asked for giving a thread, which then closes the window.


LynxAbraxas:
Not working with the current changes. Suggestions how to realize this instead are very welcome.

LynxAbraxas:
I won't find the time to work on this in the near future, so closed to serve as a reference. Please feel free to pick it up.

PR #330, opened 2020-06-29
Stop the AI from establishing embassies where it cannot

MartinGuehmann:
Stop the AI from establishing embassies, if it is at war with the target civilization, or if the target civilization has the close embassies wonder.


LynxAbraxas:
Works well, many thanks @MartinGuehmann. Do you want to include some commits of #259?

MartinGuehmann:
The only thing, that would be relevant would be the change to `ORDER_RESULT ArmyData::EstablishEmbassy`. But I did not remove it, because the point of it is to put the wonder that keeps you from establishing embassies into the message.

LynxAbraxas:
Good point, I must have overseen that then. However I think it only determines the first wonder of that player that is not obsolete, i.e. it misses a check if that wonder is the one that prevents establishing embassies.

MartinGuehmann:
> Good point, I must have overseen that then. However I think it only determines the first wonder of that player that is not obsolete, i.e. it misses a check if that wonder is the one that prevents establishing embassies.

Cleaned the code a bit and fixed that, and found an issue with another message. The stuff compiles on Windows, but I haven't tested it, yet. Even so the code should be straight forward about that.

LynxAbraxas:
Works like a charm now:
ss_2020-07-05_13:28:25
ss_2020-07-05_14:11:38
and the diplomat around a barbarian city are moving off to other targets.
Many thanks @MartinGuehmann for the fixes. I'd say good to go in.

MartinGuehmann:
Well, there is an extra space before "Gutenberg's Bible" in the message. Fixed this and some string related issues.

LynxAbraxas:
Looks good to me.

MartinGuehmann:
Then let's merge.

PR #331, opened 2020-07-02
Corrections to trade route mod concerning wonders

LynxAbraxas:
Some checks were missing from #256 concerning trade routes and wonders.


MartinGuehmann:
Then let's merge.

PR #332, opened 2020-07-04
Draw anti aliased borders

williamd67:
Borders can look jiggered especially when zooming out. When the option `Smooth Icon Borders` is selected they become incorrect when zooming out.
Screenshot from 2020-07-04 09-43-42
Screenshot from 2020-07-04 09-44-32

This PR will draw the borders and other lines with a new anti-aliased line draw (16-bit per pixel) based on  [Wu's algorithm](https://en.wikipedia.org/wiki/Xiaolin_Wu%27s_line_algorithm).

BTW: I will rebase this PR with master as soon as #324 is merged to master. I need drawing routines from that branch here.


LynxAbraxas:
Yeah, the drawing at lower res, especially the interpolation in between the highest and the lowest can look bad.
I also noticed that the aa-algorithm used for drawing trade routes was bad (possibly wrong) when working on #159, so decided to draw routes without aa. Perhaps with your new implementation, aa for trade routes could work out well. I also considered drawing the trade routes with some transparency, but did not realize that because it will lead to some brown black when many differently colored trade routes share the same path, still that could be seen as in indication that there are many routes in that segment. What might be interesting is to revive the highlight function for a trade route and use that e.g. in the message of a route being pirated or when a route is selected in the trade manager.

> BTW: I will rebase this PR with master as soon as #324 is merged to master. I need drawing routines from that branch here.

You can, but you don't have to.

williamd67:
@LynxAbraxas, I implemented the trade-routes using the new anti-aliased line-drawings. Please have a look and see if you like the end result.

williamd67:
The most visible change is that the path of an unit (unfinished, or tracking) is done with the new anti-aliased line methods. As the old method darkened the color and added some kind of border the line is now thinner (lighter).

In case it is too thin we could make them two pixels wide in stead of one.

LynxAbraxas:
> @LynxAbraxas, I implemented the trade-routes using the new anti-aliased line-drawing methods. Please have a look and see if you like the end result.

It is definitely much better anti-aliased than it used to be, still I have the impression that the color appears slightly adjusted (which is the nature of aa). I'm not sure which I prefer best, this aa or no aa. @MartinGuehmann what do you think?
In any case it seems their drawing during scrolling is changed, such that now they do not appear dashed during scrolling in the direction of the line (horizontal or vertical).

LynxAbraxas:
> The most visible change is that the path of an unit (unfinished, or tracking) is done with the new anti-aliased line methods. As the old method darkened the color and added some kind of border the line is now thinner (lighter).
> 
> In case it is too thin we could make them two pixels wide in stead of one.

Appears OK to me.

williamd67:
> In any case it seems their drawing during scrolling is changed

This is not directly caused by the AA but by another change I made. I aligned the dashing (pattern) so that when you draw two line-elements by two different draw statements, they appear as a single line. This aligning is based on the x and y coordinates of the line which change during scrolling. I do agree that it is an unwanted effect and will try to solve it. Worst case I will revert the drawing of trade-routes during scrolling.

williamd67:
@LynxAbraxas, I reverted the anti-aliasing trade-routes as I agree with you, that it was not better. I think anti-aliasing works best when there is single color background.

I made some other changes and hope you think they are an improvement:
* lengthened the dash to 7 pixels to distinguish better from the city-influence border;
* thinned the line when zoomed out (one main pixel and one outline in stead of two main pixels and one outline);
* made the outline color the darkened main color in stead of black.

Please have a look and see if you think it is an improvement. You can also play with it yourself, changing these settings should be straight forward.

LynxAbraxas:
Looks really great. I especially like the displaced border lines, which now allow to see the color of adjacent borders. The only thing I find a bit odd looking is the effect that the dashed line of trade routes become continuous when scrolling. How comes the city influence lines do not exhibit this effect?

williamd67:
> The only thing I find a bit odd looking is the effect that the dashed line of trade routes become continuous when scrolling. How comes the city influence lines do not exhibit this effect?

This is caused by the fact that trade-routes are redrawn over the whole screen not only the new scroll area. City-influence line are not redrawn. I am half-way a fix to limit drawing of trade-routes only in the new scroll area. It need some small tweaking still (like wrapping).

williamd67:
This fixes the opaque trade-route redrawing while scrolling.

LynxAbraxas:
Will test the new commits.

> made the outline color the darkened main color in stead of black.

Not sure about this change. It looks nice, but inside FOW the colors are then all the same and for some player colors it can be difficult on some tile types to see the routes. So perhaps black, although a bit boring, might be the better choice here.

LynxAbraxas:
Looks really great now, both zoomed in and out and when scrolling. Many thanks @williamd67 for this PR. 
Very minor: You could add the reason/considerations in the commit message of db6b286, and I would think it is "when zoomed out" in ee3cbbc.

LynxAbraxas:
Hm, is the drawing of dark path lines as in
ss_2020-07-14_23:02:15
related to this PR?

williamd67:
>  is the drawing of dark path lines related to this PR?

I guess so as I did change the drawing of the path-lines to anti-aliased. I also changed recently the drawing method as I duplicated too much code. So I guess it is an issue with the combined method that draws special lines. I will investigate the issue and fix it.

> I would think it is "when zoomed out" 

You are right it should be 'zoomed out'. I will fix the commit comment (when rebasing).

> You could add the reason/considerations in the commit message

Good point. I will update the commit message (when I rebase the branch).

LynxAbraxas:
Yes, solved with https://github.com/civctp2/civctp2/commit/f0da863857032652f41a3d5ebab239fa7c07dffd, will test the new ones.

BTW: I've always been wondering what the special path character `'*'` is indicating.
@williamd67 did you by any chance observe oddities in the code of the A* algorithm that could be the source for #237?

williamd67:
> observe oddities in the code of the A* algorithm that could be the source for #237?

No, I did not need to look into the A* algorithm till now. I did a quick experiment and indeed are helicopters allowed to fly pass enemy cities (at war) and it was attacked.

I will go off-line a couple of weeks. When I am back I will have a look at the A* algorithm. It looks complicated to me. I expected a standard algorithm with a complicated cost-function where the cost-function would differ for different units, such as land, air and sea. As I only had a brief look, I may have looked in the wrong place.

williamd67:
As I do not have any open issues any more for this PR, I removed WIP from the title.

LynxAbraxas:
Seems to work all nicely, for me new behavior of c54fad7ebf3a0b23b015720c4792e833b6ccb17a is a bit odd, but that might just because I'm not used to it.

williamd67:
> for me new behavior of c54fad7 is a bit odd

Ok, play with this for a while, if you do not get used to it or do not like it, let me know as it is now easy to change.

LynxAbraxas:
> Ok, play with this for a while, if you do not get used to it or do not like it, let me know as it is now easy to change.

What I am missing is that the centering on units/cities does not lead to a centered unit/city any more, but I can get used to that.

LynxAbraxas:
I noticed that the bombard dashed line is drawn differently depending on the direction:
ss_2020-07-20_22:24:29
ss_2020-07-20_22:24:46

LynxAbraxas:
> What I am missing is that the centering on units/cities does not lead to a centered unit/city any more, but I can get used to that.

Here a screen shot, where hitting the eye-button of the message leads to a map centered more on another city than the actual one:
ss_2020-07-19_22:50:28

williamd67:
> I noticed that the bombard dashed line is drawn differently depending on the direction

I am really sorry that you had the same mistake twice. This is the same issue as before, i.e. drawing in shadow in stead of real color now only for lines with patterns such dots and dashes. The method is duplicated for performance reasons (if that is needed is questionable) so also the issue was duplicated. My bad.

Regarding the centering: I can change it that centering will allow larger black borders (the centered item will really become the center) and keep the smaller borders when scrolling and clicking on the radar-map, or I could roll it back to previous behaviour.

I will at least ensure that centering will work again.

LynxAbraxas:
> I am really sorry that you had the same mistake twice. This is the same issue as before, i.e. drawing in shadow in stead of real color now only for lines with patterns such as dots and dashes. The method is duplicated for performance reasons (if that is needed is questionable) so also the issue was duplicated. My bad.

No problem, that's what testing is for and I too well know the problem of code duplication, which is often not easy to judge about, but generally I have drifted to avoiding code duplication even if is a performance penalty, it is much easier to maintain.

>  Regarding the centering: I can change it that centering will allow larger black borders (the centered item will really become the center) and keep the smaller borders when scrolling and clicking on the radar-map, or I could roll it back to previous behaviour.

As you see fit.

williamd67:
> As you see fit.

I implemented it a little different and am happy about the result. Any tile can become the center tile, by scrolling or by clicking on the radar-map. Of course, centering by city or unit will also work.

As the behavior of the scrolling, clicking, and centering is consisten it still feels natural to me which was my initial issue (next to the fact that on some zoom-levels not all tiles could be reached).

LynxAbraxas:
Yes, works very well now. Many thanks @williamd67.

MartinGuehmann:
Then let't merge.

PR #335, opened 2020-07-05
Adjustments to messages

LynxAbraxas:
PR introducing messages in case of VictoryEnslavement and some minor adjustments to messages.

### DIFF
--- a/ctp2_code/gs/gameobj/ArmyData.cpp
+++ b/ctp2_code/gs/gameobj/ArmyData.cpp
@@ -8746,17 +8746,7 @@ sint32 ArmyData::Fight(CellUnitList &defender)
 	if(defenderSucks) {
 		if(ta.IsValid()) {
 			g_director->AddAttack(ta, td);
-			g_gevManager->AddEvent(GEV_INSERT_AfterCurrent,
-								   GEV_BattleAftermath,
-								   GEA_Army, m_id,
-								   GEA_MapPoint, pos,
-								   GEA_Unit, ta,
-								   GEA_Unit, td,
-								   GEA_Player, m_owner,
-								   GEA_Player, defense_owner,
-								   GEA_Int, 0,
-								   GEA_End);
-			for(i = 0; i < defender.Num(); i++) {
+			for(i = 0; i < defender.Num(); i++) { // insert unit kill event such that it is exectued after BattleAftermath event
 				CAUSE_REMOVE_ARMY cause = CAUSE_REMOVE_ARMY_DIED_IN_ATTACK;
 				if(defender[i].m_id == td.m_id) {
 					cause  = CAUSE_REMOVE_ARMY_DIED_IN_ATTACK_ON_TOP;
@@ -8768,6 +8758,16 @@ sint32 ArmyData::Fight(CellUnitList &defender)
 									   GEA_Player, m_owner,
 									   GEA_End);
 			}
+			g_gevManager->AddEvent(GEV_INSERT_AfterCurrent, // BattleAftermath event has to be added after unit kill to be executed before
+								   GEV_BattleAftermath,
+								   GEA_Army, m_id,
+								   GEA_MapPoint, pos,
+								   GEA_Unit, ta,
+								   GEA_Unit, td,
+								   GEA_Player, m_owner,
+								   GEA_Player, defense_owner,
+								   GEA_Int, 0,
+								   GEA_End);
 		}
 	}
 	else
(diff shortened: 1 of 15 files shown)


LynxAbraxas:
Works nicely upto 1d47ffeabc995aebf0e7d0b50236f1fd0f214663, but dd50eb3039fa8c9bcb02a127c54d4ecac21bc236 causes a segfault on
https://github.com/civctp2/civctp2/blob/dd50eb3039fa8c9bcb02a127c54d4ecac21bc236/ctp2_code/gs/world/cellunitlist.cpp#L610

```
Backtrace:
./ctp2(_ZNK8UnitData7GetTypeEv+0xc)[0x5582439f52d0]
./ctp2(_ZNK4Unit7GetTypeEv+0x20)[0x558243b8d6ee]
./ctp2(_ZN12CellUnitList20DoVictoryEnslavementEiRK4Unit+0x365)[0x558243db712d]
./ctp2(_ZN14AftermathEvent15GEVHookCallbackE10GAME_EVENTP16GameEventArgList+0x5b7)[0x558243bbca9f]
./ctp2(_ZNK13GameEventHook8ActivateEP16GameEventArgListiRi+0x158)[0x558243aba8f8]
./ctp2(_ZN16GameEventManager12ActivateHookE10GAME_EVENTP16GameEventArgListiRi+0x9e)[0x558243ab7844]
./ctp2(_ZN9GameEvent7ProcessEv+0x5b)[0x558243aba341]
./ctp2(_ZN16GameEventManager11ProcessHeadEv+0xa7)[0x558243ab753d]
./ctp2(_ZN16GameEventManager7ProcessEv+0xa3)[0x558243ab7469]
./ctp2(_ZN6CivApp7ProcessEv+0x15d)[0x5582436922fb]
./ctp2(_Z7CivMainiPPc+0x2ee)[0x558243687b76]
./ctp2(main+0x2c)[0x55824368771f]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7fb1111edb97]
./ctp2(_start+0x2a)[0x55824366e8da]
```
@williamd67 @MartinGuehmann any ideas why `looser` is already gone at that point? 
Or any ideas how else to realize a message to the victim telling which unit is gone (and where it was)?

MartinGuehmann:
If I remember correctly, those units are killed during battle. So they should be gone when AftermathEvent is executed. However, from the code I have seen, it may be assumed that these units are still around. But I wouldn't count on that.

LynxAbraxas:
> If I remember correctly, those units are killed during battle. So they should be gone when AftermathEvent is executed. However, from the code I have seen, it may be assumed that these units are still around. But I wouldn't count on that.

Many thanks @MartinGuehmann for that hint, with the simple change in 7a2cc13 for `ctp2_code/gs/gameobj/combatevent.cpp` it now works as expected:

ss_2020-07-14_21:06:22
ss_2020-07-14_21:01:45

ss_2020-07-14_21:04:58
ss_2020-07-14_21:10:41

ss_2020-07-14_21:24:56
ss_2020-07-14_21:26:08

LynxAbraxas:
@williamd67 while debugging this, I stumbled upon an issue concerning the drawing of stealth units like slavers. As you can see in these screen shots
ss_2020-07-14_21:04:26
ss_2020-07-14_21:04:43
slavers are positioned where the tile under the cursor is highlighted (between Lodz and Radom and east of Poznan) but are draw far off (Geniezno and west of Persis). Do you have an idea why this happens? I currently only observe this when changing the player with the cheat mode, but have observed it also when not using the cheat mode. Here is the save-game if you want to test: 
[stealthUnitPos.zip](https://github.com/civctp2/civctp2/files/4920931/stealthUnitPos.zip).

williamd67:
> slavers are positioned where the tile under the cursor is highlighted (between Lodz and Radom and east of Poznan) but are draw far off (Geniezno and west of Persis). Do you have an idea why this happens?

Not immediately I expect a discrepancy between unit and actor position. Thanks for the save game I will investigate it.

LynxAbraxas:
> Still have to check if this call to `GEV_BattleAftermath` also needs adjustments:
> 
> https://github.com/civctp2/civctp2/blob/be3f50a0118bc5860fb11fed29486c144601e6aa/ctp2_code/gs/gameobj/ArmyData.cpp#L8746-L8772
> 
> However I've currently no idea in which case/situation one would end up there, i.e. how to construct a save-game situation where `defenderSucks == true`:

The only "unit" with 0 defense and 0 attack with default settings is an empty city, which the AI does not seem to attack with an army including a slaver, so I changed the default settings of Eco-rangers and Eco-terrorists to have 0 defense and 0 attack in order to be able to test the changes of a241ae932b748f4f5404cdff887cfb6b53f5bb5f which seem to work all fine:
ss_2020-07-16_09:40:47
ss_2020-07-16_09:45:38
ss_2020-07-16_09:50:05

williamd67:
I found a reason why unit- and actor-position were misaligned. In case of a stealth-unit under some conditions, moving the actor was skipped. I fixed that situation in #338.

LynxAbraxas:

MartinGuehmann:
Then let's merge.

PR #336, opened 2020-07-05
Fix issue with #311

LynxAbraxas:
see https://github.com/civctp2/civctp2/pull/311#issuecomment-651999928 and following comments.

### DIFF
--- a/ctp2_code/ai/strategy/scheduler/scheduler.cpp
+++ b/ctp2_code/ai/strategy/scheduler/scheduler.cpp
@@ -693,7 +693,7 @@ void Scheduler::Match_Resources(const bool move_armies)
 			goal_ptr->Rollback_All_Agents(); // Just roll back but don't report to the build list
 
 			// Actually should be checked in the next cycle, but there still seems to be something wrong.
-			m_goals.erase(goal_iter);
+			goal_iter = m_goals.erase(goal_iter);
 
 			/*
 			// Move to the end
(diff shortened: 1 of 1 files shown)


LynxAbraxas:
@williamd67, @henshall828 what do you think? Please feel free to adjust.

henshall828:
> 
> 
> @williamd67, @henshall828 what do you think? Please feel free to adjust.

The second part of the fix may not be quite right, as it could again lead to the problem with decrementing an iterator at the beginning of the list which leads to goal_iter containing an invalid value. A cleaner way to do this could be to have Repriotize_Goal return the iterator to use, then the code becomes
`goal_iter = Reprioritize_Goal(goal_iter)`
following the same pattern as the erase.

LynxAbraxas:
@henshall828 please create a commit or PR that contains the changes.

ctp2_code/ai/strategy/scheduler/scheduler.cpp
```diff
@@ -725,8 +725,11 @@ void Scheduler::Match_Resources(const bool move_armies)
 					// http://www.cplusplus.com/reference/stl/list/splice.html
 					//or use a decrement
 					// Sort the goal list, move iterator increment herein back
+					tmp_goal_iter = goal_iter;
+					--goal_iter;
```
henshall828, line 729: If goal_iter is the first item in the list of goals, this will cause an error.

henshall828:
> 
> 
> @henshall828 please create a commit or PR that contains the changes.

Please can you take a look and test PR #337 

Many thanks @LynxAbraxas

LynxAbraxas:
Closed in favor of #337.

PR #337, opened 2020-07-05
Fix issues introduced to AI scheduler by #311

henshall828:
In two cases goal_iter would not be pointing to the correct item in the m_goals list in the next loop iteration.

1. Fix iterator when a goal is erased
2. Fix iterator when Reprioritize_Goal moves the current goal to later in the list

see [https://github.com/civctp2/civctp2/pull/311#issuecomment-651999928](#311 (comment)) and following comments and pull request #336 

Also fix the debugging loopCount to only increment when moving on to the next item in the list - not erasing, replacing or repeating the current item.


LynxAbraxas:
Tested this for some turns and seems to solve the issue. Thanks @henshall828 for the fix.

MartinGuehmann:
Then let's merge.

PR #338, opened 2020-07-16
Move stealth-unit actors even when unit is stealthy

williamd67:
This will ensure that when a stealthy unit moves not only the unit but also the actor is moved.


LynxAbraxas:
Yes, this could very well be the problem. Looks good to me.

MartinGuehmann:
Fine, then let's merge this.

Issue #327, opened 2020-06-27
Getting advance from other civ or goody hut changes current researched advance


LynxAbraxas:
If one gets an advance from other civ (steal, diplo, conquer city) or goody hut, the current researched advance is changed. This is not only annoying, but it can also happen that the limited list of advances available for research does not list the advance any more that was already under research. If that was the only one towards a specific goal, this is really annoying.

williamd67:
Indeed, I had the same issue last week as well.

LynxAbraxas:
I'd say that the general randomization of the advances that are listed for research (sub-set of max 4 advances that could be researched in regard to full-filled prerequisites) is fine, but the one already under research should be kept or at least always be part of the listing.

LynxAbraxas:
Probably this code is related:
https://github.com/civctp2/civctp2/blob/4778e5cf0a0c18f55508088d76f7efdc52332293/ctp2_code/gs/gameobj/Player.cpp#L8047-L8065

LynxAbraxas:
Solved with #339

PR #339, opened 2020-07-19
Keep advance that is being researched available in advance selection

williamd67:
Ensure that when a new advance is given by 'goody hut' or due to diplomacy, the advance that is being researched, is still available in the advance selection list (research-dialog).

I expect that this solves #327.

### DIFF
--- a/ctp2_code/gs/gameobj/Advances.cpp
+++ b/ctp2_code/gs/gameobj/Advances.cpp
@@ -26,14 +26,14 @@
 //
 // - Safeguard FindLevel against infinite recursion.
 // - Speeded up goody hut advance and unit selection.
-// - Replaced old civilisation database by new one. (Aug 22nd 2005 Martin Gühmann)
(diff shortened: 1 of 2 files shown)


LynxAbraxas:
Yes, seems to solve #327:
Before exchanging advance for city:
ss_2020-07-19_17:29:12
exchanging advance for city:
ss_2020-07-19_17:30:01
after exchanging advance for city:
ss_2020-07-19_17:30:46

"smart materials" is still available in the list but not the selected any more.
Without this PR former research "smart materials" is not available in the list:

ss_2020-07-19_17:26:06

@williamd67 do you see a way to also keep the formerly researched advance as the one researched/selected?

williamd67:
>  do you see a way to also keep the formerly researched advance as the one researched/selected?

I will have a look as I thought it would automatically keep the current one. I retested it and indeed it does not.

LynxAbraxas:
This is great! Works nicely now. Many thanks @williamd67.

MartinGuehmann:
Looks good let's merge.

PR #340, opened 2020-07-19
Play only one franchise animation (moving)

williamd67:
Before two animations were started both the moving from unit to city, and the stationary within the city. This has been fixed to only one animation is shown, moving if the unit is visible else  stationary.


LynxAbraxas:
Works as expected now.

MartinGuehmann:
Let's merge.

Issue #341, opened 2020-07-20
End-game score not saved in the hight score list


LynxAbraxas:
It seems that the final achieved score is not added to the high score list (under linux/docker):
![ss_2020-07-20_22:49:25](https://user-images.githubusercontent.com/21012234/87986365-50c6a400-cadd-11ea-9eb7-d07b25ec7875.png)
even though the score is not zero:
![ss_2020-07-20_22:54:57](https://user-images.githubusercontent.com/21012234/87986380-58864880-cadd-11ea-908d-c402fafd4b3e.png)

The full screen end-game movie plays nicely, but it seems the music is not started again afterwards (it does for the wonder movies).

LynxAbraxas:
Here is a save-game for testing: 
[sciVic.zip](https://github.com/civctp2/civctp2/files/4950071/sciVic.zip)

Could this need some text file mapping into the docker container?

williamd67:
> Could this need some text file mapping into the docker container?

I do not expect so as I have the same issue on my Ubuntu machine.

LynxAbraxas:
> I debugged it and the keep-score flag is set to false which happens when the 'disclaimer' for the scenario-editor is shown. So in case you edited this game for debugging or other wise, it will not keep the high-score. 

Many thanks @williamd67 for checking. I surely went into the cheat-mode for debugging, so this explains it. Does that also happen in case the scenario-editor is started when adjusting the map for a new game, i.e. creating a scenario?

williamd67:
>  when adjusting the map for a new game, i.e. creating a scenario?

The save-scenario action (button) of the editor resets the keep-score flag to _true_ so that high-scores for scenarios will be stored as long as the editor is not opened while playing the scenario.

BTW: as mentioned before I will be off-line for a couple of weeks

LynxAbraxas:
I see, thanks.

> BTW: as mentioned before I will be off-line for a couple of weeks

OK, looking very much forward to see you back here. I'll have to stop here for some time too.

PR #342, opened 2020-07-21
Start music after showing victory movie

williamd67:


williamd67:
This fixes the fact that music was not started after a victory movie was shown which was mentioned in #341.

LynxAbraxas:
Yes, works well.

MartinGuehmann:
Then let's merge.

PR #343, opened 2020-07-25
Mapping high-score file into docker container

LynxAbraxas:
This PR solves #341:
ss_2020-07-25_01:18:36


MartinGuehmann:
From my point of view, this is finished. @LynxAbraxas do you still want to remove the commits for testing? Or do you thing that they have some value for documentation?

LynxAbraxas:
Yes, finished and I would keep the testing commits as a ref how this was tested.

MartinGuehmann:
Good then we can merge.

PR #345, opened 2020-07-27
Fix compiling and a crash on windows

MartinGuehmann:
Pull request stopped MS Visual Studio from compiling the game and also introduced a crashed, when I was loading a savegame.

I am still not happy with the crash fix since it just reverts the changes in radarmap.cpp from 07c099eb0c2268b00b9e1e56cf64a52b0b918eed.


MartinGuehmann:
Looks good. I haven't really looked into the surface locking, but it is just logical to do it.

MartinGuehmann:
Then let's merge this.

PR #347, opened 2020-08-27
A star improvements

williamd67:


MartinGuehmann:
Is it supposed to change anything, practically? I remember that I tried to remove it but I was for some reason unhappy with it.

You can also define PRINT_COSTS, to check if the costs are different.

williamd67:
> Is it supposed to change anything, practically?

The deprecation in itself does not solve anything. As it in my opinion neither improved nor fixed anything I reasoned that it could be deprecated.

The real fix is in [5bfb66b](https://github.com/civctp2/civctp2/pull/347/commits/5bfb66b126acb6cb2ce2724a24bd66c7ca9071b9) where for all 'future'/'resulting' costs the minimum is chosen. This will ensure that A* always find the shortest path.

> You can also define PRINT_COSTS, to check if the costs are different.

The costs were different and A* stopped to early. It did not investigate all paths which was caused by the fact that some paths got a heavy 'future' penalty and therefore were ignored. This penalty was invalid. I changed it to make the penalty the minimum possible.

MartinGuehmann:
In fc53be2 you actually did not deprecate m_can_space_launch, in fact you removed it. Deprecating means telling people not to use it, because it may be removed in a future version.

williamd67:
Good point. I will change 'deprecate' to 'remove' in future commits (or in case I rebase). I was already using 'deprecate' incorrectly for a while so it changed meaning for me.

LynxAbraxas:
Guess this could be a solution to #237.

williamd67:
I expect it to solve the incorrect paths, mentioned in #237.

It does not solve creating a path close to active-defense. I did not feel completely comfortable to implement that. As player you should be aware of enemy cities and the chance that there is an active unit. Also paths over multiple turns may become an issue. Next to that I cannot foresee the impact on the AI. I have thought about creating an option to support the human player to prevent mistakes.

@LynxAbraxas: welcome back, I missed your feedback, testing and guidance.

williamd67:
I do not have any open tasks any more. It does not solve the issue that paths pass active defense and units will be attacked. This needs more discussion.

MartinGuehmann:
> It does not solve creating a path close to active-defense. I did not feel completely comfortable to implement that. As player you should be aware of enemy cities and the chance that there is an active unit. Also paths over multiple turns may become an issue. Next to that I cannot foresee the impact on the AI. I have thought about creating an option to support the human player to prevent mistakes.

That does not belong into the UnitAstar itself. That is stuff only the AI should deal with. And so it belongs into the RobotAstar. And in fact that is the place where it is implemented.

LynxAbraxas:
> @LynxAbraxas: welcome back, I missed your feedback, testing and guidance.

Thanks @williamd67! I'm sorry to say it was only a short re-visit during holidays, hoping to be back some day with more time again...

LynxAbraxas:
As far as I can tell by now, works find and could go in.
@MartinGuehmann what do you think?

LynxAbraxas:
> In [fc53be2](https://github.com/civctp2/civctp2/commit/fc53be2c782a5eb4cc8ddc0fbc4f6e058f9fe475) you actually did not deprecate m_can_space_launch, in fact you removed it. Deprecating means telling people not to use it, because it may be removed in a future version.

Hm, any idea what `m_can_space_launch` could have been meant for here?

LynxAbraxas:
@MartinGuehmann what do you think?

williamd67:
@MartinGuehmann anything I need to do before this can be merged?

MartinGuehmann:
@williamd67: I don't think right now. First I need to merge #375. For that I have to mark the revision with the regression. And then I have to test it, to see whether there are any problems. That simply has the potential that something goes wrong so I should test.

MartinGuehmann:
That looks good and so can go in.

LynxAbraxas:
> The real fix is in [5bfb66b](https://github.com/civctp2/civctp2/pull/347/commits/5bfb66b126acb6cb2ce2724a24bd66c7ca9071b9) where for all 'future'/'resulting' costs the minimum is chosen. This will ensure that A* always find the shortest path.

Seems this is now: 2d16a4c3b84475cd9e60a8d7d14f6050a4658aab

Issue #285, opened 2020-03-22
GL-CI integration for civctp2 is gone


LynxAbraxas:
It seems on 2020-03-22 GL stopped the possibility to integrate GH into private repos:
https://about.gitlab.com/releases/2020/03/12/free-period-for-cicd-external-repositories/
While civctp2 in general is open/public, the very special licenses of the source code of CTP2 does not allow us to make the GL fork public because we cannot keep the docker registry private, since the docker images hold the files form the CD and these are needed for the integrated testing.

LynxAbraxas:
GL reverted this for now:
https://gitlab.com/gitlab-org/gitlab/-/issues/10361#note_309549475

LynxAbraxas:
Back to disabled without a paid plan: https://gitlab.com/gitlab-org/gitlab/-/issues/10361#note_312582813
I made a post for the alternative configuration (https://gitlab.com/gitlab-org/gitlab/issues/19585#note_311301489), that only the registry is kept private, which would be nicer any way since compile logs then would be public and available for  discussion directly. 
Until that is available, @MartinGuehmann would have to push to the GL repo after a merge on GH for the smoke tests to run because it is not just the integration with GH that is gone but also the option to activate pull for mirroring in private GL repos. 
Perhaps GH has an option to do push mirroring on merges, which could be an alternative.

LynxAbraxas:
With https://gitlab.com/gitlab-org/gitlab/-/merge_requests/54831 in GL-14.x (GL-14.2 is running on gitlab.com now), the DI registry can be kept private while making the project public.
So with that settings, pulling from GH works again (might need to be reactivated on GL because the pull-mirroring is disabled after 14 failures: https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html#how-it-works).
Furthermore, GL-CI/CD minutes are not limited any more (https://docs.gitlab.com/ee/subscriptions/gitlab_com/index.html#ci-pipeline-minutes) since it is a public project created before 2021-07-17.

LynxAbraxas:
@MartinGuehmann Can you change those settings for https://gitlab.com/civctp2/civctp2 ?
1. under Settings > General > Project visibility > choos public  and set "Container registry" to "Only Project Members"
[...]

PR #351, opened 2020-12-29
Debian packages of CTP2 from GL-CI

### LINKED ISSUE #322: Binary packages for Linux
Currently, all milestones for "Linux Release" are [solved](https://github.com/civctp2/civctp2/issues?q=milestone%3A%22Linux+Release%22+). The only thing that is missing I think is the creation of binary packages for Linux that could be used to install CTP2 with a package manager. This should be possible to do with the compilation done in CI. It would need some means to provide a path to the GOG/CD files that we cannot put in the package. Since the Dockerfile for CI uses Ubuntu for building, supporting only DEB-packages for a start seems reasonable. So all this needs is the automatic creation of a DEB-package as part of the CI to become CI-CD (continuous delivery). Since the smoke tests only work for GL, it would be good to use GL CI-CD, which would also offer the option for manual tests as a stage to continuous delivery.

LynxAbraxas:
For the 20th anniversary of CTP2, this PR offers a means to create Debian packages with the GL-CI (resolves #322) thereby providing automated releases for Linux.

With that, the installation of ctp2 only needs two actions (the game files need to be provided before the installation of the dep-package):
```
cp -r Scenarios/ ctp2_data/ ctp2_program/  /opt/ctp2/ # copy game files from the original CD or GoG
apt install ctp2-plain.deb
```
as done in the DF:
https://github.com/civctp2/civctp2/blob/ef6d8f720b9109d14e9a127b465cab861a4a0707/Dockerfile#L57-L66

Both, plain and debug versions get packaged (`ctp2-plain.deb` and `ctp2-debug.deb`) and are used/tested in the GL-CI test-stage.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,3 +14,4 @@
 
 ## files essential for testing
 !ctp2CD/
+!deb/
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -85,7 +85,7 @@ debug:
     - echo $DOCKER_PARAMS
     - echo $CTP_PARAMS
     - mkdir -p $(pwd)/logs/ && chmod a+rw $(pwd)/logs/ # needs to be rw for diUser
-    - XVFBtmp=${SHARED_PATH}/.X11-unix/ HOME=$(pwd) ./run.sh
+    - XVFBtmp=${SHARED_PATH}/.X11-unix/ HOME=$(pwd) tools/run-DI.sh
              $DOCKER_PARAMS
              $IMAGE_TAG:${CI_JOB_STAGE}-${BTYP}
              ./ctp2 $CTP_PARAMS &
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,6 +39,13 @@ variables:
                    --target builder 
                    .
     - docker push $IMAGE_TAG:builder-latest-${CI_JOB_NAME}
+    - mkdir -p deb/ # create artifacts folder
+    - docker run --rm
+                 --volume $(pwd)/debian/:/debian/
+                 --volume $(pwd)/deb/:/deb/
+                 --volume $(pwd)/tools/build-deb.sh:/tools/build-deb.sh:ro
+                 $IMAGE_TAG:builder-latest-${CI_JOB_NAME}
+                 tools/build-deb.sh ${CI_JOB_NAME} # create deb-package from pre-compiled files
     ## 3rd stage build using system-latest and builder-latest for caching
     - apk add --update --no-cache curl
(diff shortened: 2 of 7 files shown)


LynxAbraxas:
As long as #285 is not resolved, this only works with a manual push of master (or any other commit) to GL, or needs a config of a push-event to GL on GH or locally.

Apparently artifacts cannot be made visible for private projects: https://gitlab.com/gitlab-org/gitlab/-/issues/17544
neither it is possible to restrict access to DIs (which would allow to make the GL civctp2 repo public): https://gitlab.com/gitlab-org/gitlab/-/issues/19585

So this would need some means of pushing the created artifacts from GL-CI to GH releases, which might be possible with:
https://github.com/marketplace/actions/github-upload-release-artifacts
https://github.com/actions/upload-release-asset

LynxAbraxas:
Here is the DEB-file from current master (plain, i.e. not debug) as ZIP:
[ctp2-plain.zip](https://github.com/civctp2/civctp2/files/5755926/ctp2-plain.zip)

LynxAbraxas:
@der-gabe may I tempt you with this?

MartinGuehmann:
> Both, plain and debug versions get packaged (`ctp2-plain.deb` and `ctp2-debug.deb`) and are used/tested in the GL-CI test-stage.

So what is the use of including the debug version?

> As long as #285 is not resolved, this only works with a manual push of master (or any other commit) to GL, or needs a config of a push-event to GL on GH or locally.
> 
> Apparently artifacts cannot be made visible for private projects: https://gitlab.com/gitlab-org/gitlab/-/issues/17544
> neither it is possible to restrict access to DIs (which would allow to make the GL civctp2 repo public): https://gitlab.com/gitlab-org/gitlab/-/issues/19585
> 
> So this would need some means of pushing the created artifacts from GL-CI to GH releases, which might be possible with:
> https://github.com/marketplace/actions/github-upload-release-artifacts
> https://github.com/actions/upload-release-asset

That is now a bit unclear for me. Ideally, it should be pushed, automatically. An I would prefer to call those nightlies, because these releases aren't that well tested.

LynxAbraxas:
@MartinGuehmann since #285 is now solved, this PR would offer ready-to-use DEP-packages for each commit to master as GL-CI/CD artifacts.

MartinGuehmann:
So as far as I can, you would still have to put the deb file onto GitHub for an official release. So then this should be fine.

Then let's merge.

LynxAbraxas:
Thanks! The Debian package resulting from the merge can be found here:
https://gitlab.com/civctp2/civctp2/-/jobs/1925765485/artifacts/browse/deb/
Every new merge to master should produce such a Debian package in the plain and debug job as artifact for download and ready for testing and playing if the files from the original CD or GoG are provided as described above.

LynxAbraxas:
Just to note that artifacts apparently are currently not kept for ever, so the above URL does not work any more. The newest would be:
https://gitlab.com/civctp2/civctp2/-/jobs/2687671624/artifacts/browse/deb/
It is always the artifact of the job named `plain` from the `build` stage and can be found in any successful GL-CI pipeline run.

Issue #248, opened 2020-01-16
"Submarine Canyon" tiles not working (appearing in maps)


LynxAbraxas:
It seems that currently the map cannot have "Submarine Canyon" tiles. When I try to insert them, only deep water is put there:
![ss_2020-01-16_10:08:23](https://user-images.githubusercontent.com/21012234/72518252-31f90380-3855-11ea-8dad-57fee9e1344f.png)

LynxAbraxas:
Are "Submarine Canyon" tiles the correspondent of "Trench" tiles in CTP1?
![ss_2020-01-12_19:16:43](https://user-images.githubusercontent.com/21012234/72518344-68cf1980-3855-11ea-865e-bcb9917c13a2.png)

LynxAbraxas:
I have found some submarine canyon tiles on a CTP2 map now, but some forms look as if they come form the CPT1 tile set:
![ss_2020-01-30_16:40:30](https://user-images.githubusercontent.com/21012234/73467100-1cbac380-4383-11ea-96bc-9264c1b048a8.png)
Drawing these with the map editor still does not work for me.

LynxAbraxas:
"Submarine Canyon" (trench) tiles were disabled for map generation in:  8b2beddb5d3497aef2deb2142fee7601764de904

Apparently CTP2 trench tiles are in [gtset565.til](https://github.com/civctp2/civctp2/blob/e1750a0a6e4e4d92ea475d447d6d766841fb72f8/ctp2_data/default/graphics/tiles/gtset565.til) but incomplete (27ef00e36d3c01ec1e9c6ea6bd6c9835a33c6436), possibly could be created and added with "TileEdit" (#185).

MartinGuehmann:
I assume that you have to just adjust the color of the CTP2 trench tiles and then you are fine.

LynxAbraxas:
> I assume that you have to just adjust the color of the CTP2 trench tiles and then you are fine.

Done in #353.

LynxAbraxas:
Solved with #353.

PR #185, opened 2019-08-11
Add tools for CTP2 modders

MartinGuehmann:
A tool that converts bitmaps to a text format that CTP2 can import as a map.
ModTools/BMP2CTP2.zip

ReadZFS reads the ZFS files, shows image files and plays the sound files that are inside, and can extract them.
ModTools/readZFS.zip

Activision's sprite tool that creates sprites for CTP1 and CTP2.
ModTools/unitcreation.zip

What is still missing are SpriteEdit and TileEdit. In their readmes I have:

> The legal stuff: (oh how I hate this but it's got to be there)
> 
>    Distribution:
>    This program may NOT under any circumstances be distributed
>    on any read-only media, or as part of any software package,
>    including modifications to games without explicit written
>    permission from the programmer.

So if I include it, is it then part of a software package? Or should I just ignore it?

PR #353, opened 2021-02-08
Trench tiles for CTP2

LynxAbraxas:
This PR solves #248:
ss_2021-02-08_21:51:47
Also combinations the map generator does not come up with work:
ss_2021-02-08_21:51:04
Still some constellations are not covered even with the fast variety of trench tiles available so far.


LynxAbraxas:
This was made possible by #185, many thanks @MartinGuehmann for recovering those and to their creators of old. Do you see a chance to convince "Martin the Dane"  to contribute the sources of his tools in case the old binaries do not work any more one day? This would also be interesting to be able to solve bugs (like in the export function: 656c8e79f).

MartinGuehmann:
> This was made possible by #185, many thanks @MartinGuehmann for recovering those and to their creators of old. Do you see a chance to convince "Martin the Dane" to contribute the sources of his tools in case the old binaries do not work any more one day? This would also be interesting to be able to solve bugs (like in the export function: [656c8e7](https://github.com/civctp2/civctp2/commit/656c8e79f029b24aa840cdd187017874557a0bf3)).

Well, I could just write to Martin, or you could write to him, his email address should be in the readme file.

Actually, I am not sure whether the binaries still run on Windows 10, for sure what I remember is that they work better with Wine. And thanks to Wine I won't expect them stop working.

MartinGuehmann:
Oh, actually are you happy now with this PR? If yes then I can merge.

LynxAbraxas:
> Oh, actually are you happy now with this PR? If yes then I can merge.

Yes, ready to go in.

MartinGuehmann:
Well, then let's merge.

PR #354, opened 2021-02-10
Fix linking error

laenion:
When trying to build the game I got the following error during linking:

```
/usr/bin/ld.gold: error: gs/slic/.libs/libgsslic.a(lex.yy.o): multiple definition of 'state'
/usr/bin/ld.gold: ui/ldl/lex.ldl.o: previous definition here
```

From what I can see the variable `state` is not used anyway, so just delete it.


MartinGuehmann:
This is a one liner and it compiles so let's merge then.

Issue #333, opened 2020-07-04
Building force-field for water cities not possible


LynxAbraxas:
It seem that regularly, city-walls cannot be built in water cities:
https://github.com/civctp2/civctp2/blob/f4cc17172b67fab52e6c3ceeba16ad94651adc05/ctp2_data/default/gamedata/buildings.txt#L245-L256
So, since force-fields can only be built if city walls exist
https://github.com/civctp2/civctp2/blob/f4cc17172b67fab52e6c3ceeba16ad94651adc05/ctp2_data/default/gamedata/buildings.txt#L395-L406
force-fields cannot be built in water cities. 
However, building the national shield wonder does create force-fields for water cities, so to be consistent it should be possible to either built city-walls in water cities or no force-fields in water cities or building them independently of city walls.

LynxAbraxas:
The other buildings that cannot be built in water cities are no prerequisites for other buildings:
```
awk '/{/ {a=$1}; /CantBuildInSea/ {print a, $0}' ctp2_data/default/gamedata/buildings.txt
IMPROVE_AIRPORT    CantBuildInSea
IMPROVE_BALLISTA_TOWERS    CantBuildInSea
IMPROVE_CITY_WALLS    CantBuildInSea
```

PR #357, opened 2021-04-13
Force-fields not depending on city-walls

LynxAbraxas:
- makes little sense
- allows water-cities to build force-fields

Would solve #333.


MartinGuehmann:
Does the Great Library say anything about that force fields can't be build in sea cities? I don't expect so.

MartinGuehmann:
OK, the Great Library does not say anything about that you cannot build force fields in sea cities. So let's merge.

PR #358, opened 2021-04-13
Build manager dialog improvements

williamd67:
This pull-request fixes small issues in the build manager dialog, such as load queue does not show queues without extension, delete queue does not work under Linux, and other consistency issues.


williamd67:
Before
Screenshot from 2021-04-14 18-08-57

After
Screenshot from 2021-04-14 20-54-16

LynxAbraxas:
Many thanks @williamd67 for improving this. I'm not sure if I'll find the time to take a look and test any time soon (but I would like to).
There used to be a problem with too long list in drop down selection menus of e.g. the diplo-manager (e.g. when offering a city).

williamd67:
I have no open issues any more.

I did not touch the multi-city queue editor. Although it technically works, I find it hard to get an overview what happens in your cities when using it.

LynxAbraxas:
Finally came about to test this. Compiles fine and changes appear to work as described. 
Never used the save/load queue functionality so far, now works fine for me.
Due to the shadow effect, the left side of the orange highlight box might look better if it were still 1 or 2 pixel further to the left.
Many thanks @williamd67 for these tunings.
Can be merged in my opinion. @MartinGuehmann what do you think?

LynxAbraxas:
NB: Created a new issue (#371) for the above mentioned problem with drop-down lists.

MartinGuehmann:
Then let's merge.

PR #359, opened 2021-04-21
Fix overwriting glyph (mostly capital C)

williamd67:
Sometimes character 'C' became incorrect (see pictures). This was caused by the fact that in some situation another glyph was drawn exactly over the edge of the surface, as the test was incorrect. It did not take into account that the width is zero-based.

I could reproduce this by loading the attached save-game. Opening the building-manager and going to the tab 'Wonders', and clicking wonder 'Chichen Itza'. After the fix the overwriting does not happen anymore.
Screenshot from 2021-04-21 22-55-35
Screenshot from 2021-04-21 22-56-15
Screenshot from 2021-04-21 22-56-27
[Genghi-81.zip](https://github.com/civctp2/civctp2/files/6353996/Genghi-81.zip)

### DIFF
--- a/ctp2_code/ui/aui_common/aui_bitmapfont.cpp
+++ b/ctp2_code/ui/aui_common/aui_bitmapfont.cpp
@@ -42,8 +42,8 @@
 //  add
 //      aui_BitmapFont::GlyphInfo *aui_BitmapFont::GetGlyphInfo2( MBCHAR *c )
 //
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
I cannot reproduce it. But I think I have seen it. However, it seems to make no trouble either, so let's just put it in.

PR #362, opened 2021-05-07
Animate goods when opening tile information window

williamd67:


MartinGuehmann:
Looks fine, so let's merge.

PR #363, opened 2021-05-08
Draw fogged tiles while scrolling up

williamd67:
Fogged tiles where not drawn (shown as black tiles) while scrolling up. There was a difference in implementation between non-fogged and fogged tiles. 'bigger than' (non-fogged) versus 'bigger than and equal' (fogged). The latter has been replaced by 'bigger than' as well.

Before
Screenshot from 2021-05-08 13-03-30
After
Screenshot from 2021-05-08 13-08-05


MartinGuehmann:
This fixes it, so let's merge.

PR #364, opened 2021-05-08
Trade route animation improvements

williamd67:
Trade-route good-animations adhere to graphic option: Good animations on/off
Reduce number of trade-route good-animations to one every 5-8 seconds


MartinGuehmann:
Looks good so let's merge.

PR #366, opened 2021-05-09
Fix unfinished path turn-count drawing/calculation

williamd67:
Turn-count calculation of the unfinished path was incorrect.

Before
Screenshot from 2021-05-09 16-29-42
Screenshot from 2021-05-09 16-29-08

After
Screenshot from 2021-05-09 16-34-14
Screenshot from 2021-05-09 16-33-56


MartinGuehmann:
This fixes it, so let's merge.

PR #368, opened 2021-05-23
Great Library: Sort advances based on age

williamd67:
As there are so many advances I have difficulty to determine which advance I would like to research next. I sorted the advances based on age, and added a small header for each age.

This result in the list below:
Screenshot from 2021-05-23 19-20-40

Clicking the age-description will select the first element of that age.

Let me know your opinion about this. Any suggestion/improvement is welcome.

### DIFF
--- a/ctp2_code/ctp/civapp.cpp
+++ b/ctp2_code/ctp/civapp.cpp
@@ -1721,6 +1721,10 @@ sint32 CivApp::InitializeGame(CivArchive *archive)
 		return FALSE;
 	}
 
+	if (archive) { // Ensure that SlicEngine's strings have correct numbering
+		SlicEngine::Reload(g_slic_filename);
+	}
+
 	g_theProgressWindow->StartCountingTo( 610 );
 
 	if(g_isScenario && archive &&
--- a/ctp2_code/ctp/civapp.cpp
+++ b/ctp2_code/ctp/civapp.cpp
@@ -1721,10 +1721,6 @@ sint32 CivApp::InitializeGame(CivArchive *archive)
 		return FALSE;
 	}
 
-	if (archive) { // Ensure that SlicEngine's strings have correct numbering
-		SlicEngine::Reload(g_slic_filename);
-	}
-
 	g_theProgressWindow->StartCountingTo( 610 );
 
 	if(g_isScenario && archive &&
--- a/ctp2_code/ui/interface/greatlibrary.cpp
+++ b/ctp2_code/ui/interface/greatlibrary.cpp
@@ -1752,21 +1752,11 @@ void GreatLibrary::UpdateList( DATABASE database )
 	m_topics_list->Clear();
 	m_listDatabase = database;
 
-	switch ( database ) {
-	case DATABASE_UNITS:
-
 #define HIDE(db, index) (db->Get(db->m_alphaToIndex[index])->GetGLHidden())
 
-		for (index = 0; index < g_theUnitDB->NumRecords(); index++)
-		{
-			if(HIDE(g_theUnitDB, index))
-				continue;
-
-			Add_Item_To_Topics_List(g_theStringDB->GetNameStr(g_theUnitDB->GetName(
-						g_theUnitDB->m_alphaToIndex[index])), index);
-
-		}
-
+	switch ( database ) {
+	case DATABASE_UNITS:
+		AddTopicsBasedOnAge<UnitRecord>(g_theUnitDB);
 		break;
 
 	case DATABASE_SEARCH:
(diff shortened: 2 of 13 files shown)


MartinGuehmann:
I would like to have some space above the age names and also an option to disable this. If I look for a specific advance, I want to have an alphabetic order.

And units, tile improvements, buildings, and wonders have an age, too. Even so only indirectly, but still it comes from their enabling advance.

williamd67:
@MartinGuehmann good improvements which I implemented (see below). Let me know if you have more suggestions/improvements.
Screenshot from 2021-05-29 22-36-55

LynxAbraxas:
Could not test this so far because I get merge conflicts with the other PRs (and compile errors without fixing those).

ctp2_code/ctp/civapp.cpp
```diff
@@ -1721,6 +1721,10 @@ sint32 CivApp::InitializeGame(CivArchive *archive)
 		return FALSE;
 	}
 
+	if (archive) { // Ensure that SlicEngine's strings have correct numbering
+		SlicEngine::Reload(g_slic_filename);
+	}
```
MartinGuehmann, line 1726: Doesn't that reload all there code and reset it all the internal values?

This means the internal state gets lost.

MartinGuehmann:
Why the hell is ctp2_data/english/gamedata/ldl_str.txt the only file with conflicts? Shouldn't have the versions from the other languages have conflict, too?

ctp2_code/ctp/civapp.cpp
williamd67: Thanks @MartinGuehmann, this is a really good catch. It will reset the complete SlicEngine and by that will destroy what just has been loaded from the save-file. I will revert this change and create a separate issue to investigate how we can reload the strings without destroying the current state.

williamd67: Unfortunately, this need to be fixed here, as I added a new string 'Sort by age' and by that new string all save-files are broken (strings are not correctly numbered any more). I will investigate a way to load the new strings without changing the SlicEngine state.

MartinGuehmann: That's why the readme of the Apolyton edition says that you reload slic, if you want to continue your current game. In fact we have been adding new strings all the time and we did not add them add the end, but rather were it made sense for us.

So that means, this is not part of the pull request. Of course, it would be nice to have.

williamd67: Ok, in that case this PR is ready for merging as I reverted the change.

BTW: I found a solution for the renumbering of the strings. I will create a separate PR for that.

ctp2_code/ui/interface/greatlibrary.cpp
```diff
@@ -1747,25 +1762,15 @@ void GreatLibrary::UpdateList( DATABASE database )
 	m_topics_list->Clear();
 	m_listDatabase = database;
 
-	switch ( database ) {
-	case DATABASE_UNITS:
-
 #define HIDE(db, index) (db->Get(db->m_alphaToIndex[index])->GetGLHidden())
 
-		for (index = 0; index < g_theUnitDB->NumRecords(); index++)
```
MartinGuehmann, line 1755: If you could make the variable index local here

ctp2_code/ui/interface/greatlibrary.cpp
```diff
@@ -1877,29 +1855,11 @@ void GreatLibrary::UpdateList( DATABASE database )
...
-						g_theGovernmentDB->m_alphaToIndex[ index ])), index);
-
-		}
-
+		AddTopics<GovernmentRecord>(g_theGovernmentDB);
 		break;
 
 	case DATABASE_TILE_IMPROVEMENTS:
-
-		for (index = 0; index < g_theTerrainImprovementDB->NumRecords(); index++)
```
MartinGuehmann, line 1894: Of course make this local here as well.

Then m_maxIndex = index; can also go at the bottom of this function. Actually, the Windows debug version does not like it, since index may be uninitialized. Actually m_maxIndex is not used elsewhere. So remove it also from the class definition.

ctp2_code/ui/interface/greatlibrary.cpp
williamd67: Done

ctp2_code/ui/interface/greatlibrary.cpp
williamd67: Done

MartinGuehmann:
The file ctp2_data/english/gamedata/ldl_str.txt is missing. Therefore, the sort-by-age-button has no text in the English version.

MartinGuehmann:
This is now fine, let's merge.

PR #369, opened 2021-06-06
Army manager look

williamd67:
This pull-request updates the army-manager-dialog look to be consistent with all the other manager-dialogs.

Before:
Screenshot from 2021-06-06 11-46-03

After:
Screenshot from 2021-06-06 12-08-10

### DIFF
--- a/ctp2_code/ui/interface/armymanagerwindow.cpp
+++ b/ctp2_code/ui/interface/armymanagerwindow.cpp
@@ -24,9 +24,9 @@
 //
 // Modifications from the original Activision code:
 //
-// - Initialized local variables. (Sep 9th 2005 Martin G�hmann)
-// - Standardized code (May 21st 2006 Martin G�hmann)
-// - Added army debug text to the army manager window. (Dec 24th 2006 Martin G�hmann)
+// - Initialized local variables. (Sep 9th 2005 Martin Gühmann)
+// - Standardized code (May 21st 2006 Martin Gühmann)
+// - Added army debug text to the army manager window. (Dec 24th 2006 Martin Gühmann)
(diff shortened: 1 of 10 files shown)


LynxAbraxas:
Could not test this so far because I get merge conflicts with the other PRs.

williamd67:
@LynxAbraxas I resolved the conflicts so that this can be tested (when you have some time).

MartinGuehmann:
Looks good, compiles on Windows. I am just not sure about the list, now it shows the names of the armies, which is a good thing. Before, it showed the number of units in the army which was nice. I am just not sure whether this is necessary.

LynxAbraxas:
Thanks @williamd67 for this PR. Works for me and I very much like the bigger size. Do you see a chance to also print the number of units in the army as done in the old version?
@williamd67 BTW: You can delete the branches of merged PRs (unless You intend to keep them). That way it is easier for other who fetch your repo to see which branches/PRs still need testing. That's why I missed testing this PR for some while (sorry about that).

williamd67:
@LynxAbraxas I will have a look if I can re-introduce the number of units in an army. BTW: I deleted my branches.

LynxAbraxas:
That's great! Thanks!
One thing I noticed that is only indirectly related to Your PR here is, that I sometimes get the army manager opened with a double click on a city where as sometimes it is the city manager. So far I could not make out when I get which.

williamd67:
Screenshot from 2022-09-03 10-13-07

LynxAbraxas:
Thanks @williamd67! That's great!
@MartinGuehmann I'd say ready for merging.

MartinGuehmann:
That's good so let's merge.

PR #372, opened 2021-08-22
Ensure that SlicEngine's strings will be resolved post serialization

williamd67:
In case an older save-game is loaded and the numbering of the strings changed, incorrect messages are shown. This can also happen when the save-game will be loaded with a different language.


MartinGuehmann:
That works, so let's merge.

LynxAbraxas:
So with this we don't need to enter the special code '/reload slic' in the chat-window any more?

PR #375, opened 2021-12-28
Fix two bugs

MartinGuehmann:
Do not show units in cites, if they are fortified (AI only testing), I think that also caused some problem with the AI 
Do not center on dying units, when auto centering is off


LynxAbraxas:
Compiles fine for me. You could add the SHA1 of the commit for the regression reference.

williamd67 (approved):

MartinGuehmann:
OK, I could put the SHA1s into the merge pull request message. I think that is a compromise.

PR #381, opened 2022-07-03
refactor: introduced Attributes-class (replaced defines with constants)

williamd67:
Made aui_Region::m_attributes private and made bit-manipulation explicit via methods.


MartinGuehmann:
Looks good, so can go in.

PR #385, opened 2022-08-27
Update Dockerfile to use newest stable Ubuntu as base DI

LynxAbraxas:
This PR updates the Dockerfile to use newest stable Ubuntu as base DI.
With this we can detect if changes to the newest stable Ubuntu cause build/run problems, especially if setting up a monthly schedule in the GL-CI for the master branch.
General problems are updates to gcc (being more strict about the code consistency) and to packages that include a version number in their name (e.g. libswscale5)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ################################################################################
 # base system
 ################################################################################
-FROM ubuntu:18.04 as system
+FROM ubuntu as system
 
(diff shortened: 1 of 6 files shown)


LynxAbraxas:
Works for Ubuntu jammy (22.04LTS):
https://gitlab.com/LynxAbraxas/civctp2/-/pipelines/624512085

PR #386, opened 2022-08-27
Disable inappropriate sounds (for e.g. special attacks if units are in FOW)

LynxAbraxas:
This PR disables inappropriate sounds (for e.g. special attacks if units are in FOW) and otherwise focuses on the spot for which the sound is played. 
This would solve #299 and might be related to #54.


LynxAbraxas:
Tested this for a while and works as expected. 
@MartinGuehmann What do you think? Ready for merging?

ctp2_code/gs/gameobj/ArmyData.cpp
```diff
@@ -5891,6 +5892,9 @@ ORDER_RESULT ArmyData::InterceptTrade()
 						g_selected_item->ForceDirectorSelect(Army(m_id));
 						return ORDER_RESULT_ILLEGAL;
 					}
+					if (g_player[g_selected_item->GetVisiblePlayer()] == route_owner){ // set pirating sound if route_owner is human player
```
MartinGuehmann, line 5895: Visual Studio 2017 refuses to compile this line. It should only be:
`g_selected_item->GetVisiblePlayer() == route_owner`

LynxAbraxas: Many thanks for testing on Win and pointing this out (odd that it works with gcc on linux).
Fixed with [2f0714d](https://github.com/civctp2/civctp2/pull/386/commits/2f0714dc3591adfdf58b79357122274a50566c70)

LynxAbraxas:
@MartinGuehmann What do you think? Ready for merging?

MartinGuehmann:
In this file I got some diplomats trying to trow parties. The visible player is player 1, it is set up as robot. So when you press enter and the AI moves the units you will see and hear the two diplomats trying to throw parties. In your version, you sea that a couple of times. I guess that matches the number of process cycles. In the original version, you can only see it once.

I could track it to [4cf412c](https://github.com/civctp2/civctp2/pull/386/commits/4cf412c62bcd254898b5cca9bbfaff2541833b60).

Of course, this is an annoyance for automatic AI testing, and should go or resolved otherwise, before I merge.

I don't know why the diplomats trying to throw parties, even so they can't. But that is I guess another problem.

[SanAn-Mexi-120v.Chr.FailedRec.zip](https://github.com/user-attachments/files/17000138/SanAn-Mexi-120v.Chr.FailedRec.zip)

I also noticed the death sounds for the slaver and the abolitionists are not played. You can put it into the scope of this pull request, but it is not necessary, since I don't think that this comes from this PR. (Well, I think I should check again.)

LynxAbraxas:
> Of course, this is an annoyance for automatic AI testing, and should go or resolved otherwise, before I merge.

Thanks, @MartinGuehmann for testing this. Currently, I've no clue where to start resolving this. Do you have any pointers to what should be changed in https://github.com/civctp2/civctp2/pull/386/commits/4cf412c62bcd254898b5cca9bbfaff2541833b60?

MartinGuehmann:
That is quite difficult. The thing actually revealed a bug relating how the AI actually used diplomats. And so I fixed that. In principle there could be other bugs in there, which also could be revealed by this.

I would suggest to revert https://github.com/civctp2/civctp2/commit/4cf412c62bcd254898b5cca9bbfaff2541833b60. Just reverting, no rebasing and no history cleanup. It might be useful to go back to the stage as it is now in this pull request. Once this is out I can merge this pull request.

When you have remove https://github.com/civctp2/civctp2/commit/4cf412c62bcd254898b5cca9bbfaff2541833b60 you can then cherry pick it onto the current master and open for that a new PR. As it might be useful to test it on the current master to see whether it reveals some other bugs in the later game.

However, I think if animations are disabled, it should not wait for actions to complete.

LynxAbraxas:
> I would suggest to revert [4cf412c](https://github.com/civctp2/civctp2/commit/4cf412c62bcd254898b5cca9bbfaff2541833b60). Just reverting, no rebasing and no history cleanup. It might be useful to go back to the stage as it is now in this pull request. Once this is out I can merge this pull request.

OK, just did a revert of this (https://github.com/civctp2/civctp2/pull/386/commits/d7aabf7aabbfa0ccf603dfb0b74f6efa07f11541) and will start a new PR of just the cherry-pick of 4cf412c62bcd254898b5cca9bbfaff2541833b60 on master. 
What should the PR be named?

MartinGuehmann:
Let's see what the commit message was:

> use locking for "stand-alone" effect animations but not for projectiles:
>- without "stand-alone" effect animations are hardly ever visible due to other changes of map center
>- projectile animations should play together with (special) attack animations, so they should not lock

I guess it should be something about projectile and effect animations and to make them visible on the map. IN fact I am not aware of any projectile animations. Maybe that is the reason why I have never noticed these. Obviously it is also about something else then sounds.

Let's weather I still have to test it. Probably, I should do.

MartinGuehmann:
Tested for some 600 turns without any problems, except for those I fixed in the current master. That version is quite slow. That happens if you fix one or two bugs in the path finder.

Anyway, this can now go in.

LynxAbraxas:
Follow up PR is #427

PR #399, opened 2023-02-16
Fix crash when conquering nation with active franchises

jacobgruber:
I ran into a bug where conquering a nation with active franchises would crash the game on the next AI turn.

It seems the city still tries to process the franchise. `CityData::ProcessProduction` and `CityData::RemoveFranchise` both deference `g_player[m_franchise_owner]` which is null after conquering. The franchises seem to be removed after these calculations.

I added null pointer checks to both those calls, which allowed me to get to the next turn. This turn took a lot longer than I thought it would!

The check was also missing from one of the analogous conversion functions (it was present in the other). I surprisingly couldn't reproduce the crash there, but thought it best to add the test anyway.


MartinGuehmann:
No, problem with that. So it can go in.

Issue #361, opened 2021-05-05
Help with assembly code in C3Blitter Apple Silicon


whamtet:
Sorry if this is a stupid question, but I'm having problems with the assembly code in c3blitter.cpp.  There are two sets of assembly to account for windows / linux, but nothing for ARM (Apple Silicon).  

Firstly, what is c3blitter for?  Second, do we still need to have assembly code here?  Aren't modern machines fast enough to just use regular cpp?

Any suggestions much appreciated.

williamd67:
On Linux the assembly code is not used. In case you can ensure that the code between 57 and 78 of c3blitter.cpp is skipped (by #define or commented out), you can comment out all assembly code. I do not know if this will work for you.

whamtet:
Thanks @williamd67 , I'm compiling on Linux actually so this is good to know.  But what about the second section?  L376 - 409?  Is that second function never invoked?

whamtet:
Program seems to run fine with second section commented out, thanks for your suggestions @williamd67

PR #400, opened 2023-02-20
Build ubuntu arm

jacobgruber:
1) Set arm flag

Use existing but unused define to Ignore assmebly in c3blitter.cpp (See #361). Define is only used in that file, and only for hiding assembly.

Left open for future architecture checks if needed.

2) Reorder linker arguments. 
`ld` expects libraries after objects, while currently they are before. This caused problems for me when trying to compile with debugging symbols. The current instructions suggest using `gold` if `ld` fails, which I suspect is at least in part because `gold` is more permissive with the order of linker arguments.

Actually it might be a good idea to clean up the libraries here in general. Libraries are defined several times in a few places, and some of them seem to be unused. Is it even possible to use autotools on a system with DirectX? This task migh turn into a bigger refactor, however, which sound dangerous. Especially since I just got here and am very much learning on the job :)

### DIFF
--- a/ctp2_code/Makefile.am
+++ b/ctp2_code/Makefile.am
@@ -55,7 +55,8 @@ ctp_ctp2_LDADD = \
 	sound/libsound.la \
 	$(CTP2_LIBANET_LDADD) \
 	$(CTP2_LIBOSNOWIN32_LDADD) \
-	$(CTP2_LIBTTF_LDADD)
+	$(CTP2_LIBTTF_LDADD) \
+	$(CTP2_LDADD)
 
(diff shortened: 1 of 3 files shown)


MartinGuehmann:
I guess autotools is open source, so you can expect a Windows version, at least I do, but I did not check it. Sure if you wanna refactor things then  go ahead when you feel comfortable. Just make commits atomic, a small thing per commit, so that we have a chance to debug things if something goes wrong.

So as far as I can see, you want add anything to this PR and it still compiles on my system, so it can go in. And so let's merge.

Issue #401, opened 2023-02-20
Freeze in Build Manager: multiple cities and impossible build request 


jacobgruber:
Game freezes when a applying a queue to multiple cities, where one city is unable to build an item in the queue.

How to reproduce:

1) Have two cities
2) Have Banking
3) Have Bazaar in only one city
4) Open Natl Manager, select both cities
5) Add Bank to queue
6) Overwrite

jacobgruber:
I already fixed this, but I'm too tired right now to figure out how to separate it from my other open PR.

Infinite loop in `EditQueue.cpp`. Maybe nobody has ever used this feature but me.

PR #403, opened 2023-03-01
Fix infinite loop in MultiQueue application

jacobgruber:
See #401 

`customWalk->Next` was not called if the conditional failed, so the same object was checked forver.

Since the conditional body is the same for all cases, I moved the body outside the switch. This saves precious lines and IMO makes it more clear what is going on, especially with all the continues.


LynxAbraxas:
Thanks for this PR (never used this feature before). Looks fine to me.

MartinGuehmann:
Compiles on Linux and Windows, nothing obvious wrong, so can go in.

Issue #402, opened 2023-02-27
SegFaults when processing mouse events in invalid aui_Region


jacobgruber:
I've encounted a handful of segfaults in my most recent game that occur in when `Attributes::IsSet` is called from a null `aui_Region` pointer. I suspect the refactor in #381 has exposed some sloppiness in the AUI code.

In particular it seems to occur when mouse events in `aui_ListBox` when, for reasons currently unknown to me, the code tries to proccess a negative list index. Asserts in some (but not all) functions which take this index as an argument imply it is not a valid value, and that these functions don't seem to do any error handling for it.

So far this has occurred for me in the Tech Selection, Build Manager[1] and Natl Manager windows. It might happen other places as well. Other componets besides ListBoxes might generate invalid regions and try to do things with them.

I'm not sure where the best place is to catch this error since I'm not very familiar with the aui code, though I'm looking into it. @MartinGuehmann @williamd67, do either of you have suggestions?

[1] https://i.imgur.com/FVQK1YK.png

jacobgruber:
P.S. I've found a few other bugs, some of which I've made sloppy quick fixes for and some which I haven't been able to reproduce outside of a specific save game. I'm hesistant to litter this repo with issues/PRs for these as a fresh face. What would you have me do?

MartinGuehmann:
> P.S. I've found a few other bugs, some of which I've made sloppy quick fixes for and some which I haven't been able to reproduce outside of a specific save game. I'm hesistant to litter this repo with issues/PRs for these as a fresh face. What would you have me do?

You can put all the things into one PR about miscellaneous fixes. Just put each fix into one commit. In case one of the fixes introduces a bug and we only notice way later, then it is easier to debug if everything is in its own commit. Of course, things may only show up in a specific save game. But that are things that may still show up for other people and of course they should not show up and thus should be fixed.

About the list boxes, at least I can reproduce it on Linux. However, the list box must be empty and it will only crash when you click on the upper rim. I guess that this is in the general list box code. However, this is something that might have been long in the code. Empty list boxes aren't that common and if at least I would have no reason to click on the upper rim when it is empty.

PR #404, opened 2023-03-01
WIP: Fix Crashes in ListBox Mouse Event Handling

jacobgruber:
Fixes crash caused by clicking on the top border of an empty listbox. See #402 

RelativeY is negative in this case, which bypasses sanity checks which assume a non-negative value. This becomes a negative index, which causes a segfault later upon attempted deference.

Might be nice if some of the assumptions on paramters in this class were more explicit.

After this PR I'll put these small fixes into a single PR.

### DIFF
--- a/ctp2_code/ui/aui_common/aui_listbox.h
+++ b/ctp2_code/ui/aui_common/aui_listbox.h
@@ -333,7 +333,8 @@ protected:
 
 private:
 	sint32 CalculateRelativeY(sint32 y) const {
-		return y - m_y - (m_pane ? m_pane->Y() : 0);
+		sint32 relativeY = y - m_y - (m_pane ? m_pane->Y() : 0);
+		return (relativeY > 0 ? relativeY : 0);
 	}
 
 	sint32 CalculateItemIndexByRelativeY(sint32 relativeY) const {


jacobgruber:
I just found another crash in related code that is not solved by this PR. I this this PR should expand in scope to ensure that edge- and error-cases are handled properly throughout the listbox code.

Essenitally, while this first commit fixes one inconsistency in assumptions about parameters, there seem to be several others.

LynxAbraxas:
> EDIT: After further investigation I think this PR is incomplete. I need to take a moment to learn how all the pieces here want to fit together.

Thanks for the debugging and partial solving. 
Take your time, as long as the WIP: is in the title the PR won't be regarded for merging until you remove the WIP again. Just add commits to the same branch will add them to the PR here as well. 
Looking forward to you findings.

MartinGuehmann:
> > EDIT: After further investigation I think this PR is incomplete. I need to take a moment to learn how all the pieces here want to fit together.
> 
> Thanks for the debugging and partial solving. Take your time, as long as the WIP: is in the title the PR won't be regarded for merging until you remove the WIP again. Just add commits to the same branch will add them to the PR here as well. Looking forward to you findings.

@jacobgruber: Unless the maintainer decides to merge it nevertheless. That's me. :D It is more than a year. I tested it. And it does what it is supposed to do. Even so it is a partial and small fix, it is one bug less and thus still improves the game.

Unless you ask for more time or add soon something. I would say it is good to go in.

PR #405, opened 2023-03-01
Misc bugfixes

jacobgruber:
Just a couple additional null pointer checks to stop some segaults I encountered. I had trouble reproducing both of these outside of specific save files.


LynxAbraxas:
Thanks @jacobgruber, looks fine for me. 
If you don't mind, you can provide the save-game-files here in case similar problems arise e.g. during future debugging.

MartinGuehmann:
Compiles on Windows and Linux. And no obvious problem introduced from this. And that is also expected from the changed. So can go in.

PR #408, opened 2023-05-24
Removing X11 from Linux build

atsb:
X11 on Linux is now legacy and many distributions no longer provide X11 on a default system install.  This completely removes the dependency and allows CTP2 on Linux to use SDL2 for everything, this of course is still cross-platform for BSD systems since SDL2 is available there as well.

This PR also updates the configure.ac file as some content was using a very old version of AutoConf.

All these changes were tested and verified on Ubuntu 22.10.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,21 +1,21 @@
 dnl $Id: configure.ac 966 2009-04-11 11:23:43Z ctplinuxfan $
 dnl
-AC_INIT(ctp2,1.2)
-AC_PREREQ(2.54)
+AC_INIT([ctp2],[1.2])
+AC_PREREQ([2.71])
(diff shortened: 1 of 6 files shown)


LynxAbraxas:
Thanks @atsb for this contrib!
Compiles with GL-CI and also works for the smoke-tests:
https://gitlab.com/LynxAbraxas/civctp2/-/pipelines/1021313118
@MartinGuehmann ready for merging?

LynxAbraxas:
@MartinGuehmann please use merging (as before) for accepting pull requests, not rebasing (as used for this PR), since only merging (with --no-ff) shows what commits belong to a particular merge request.

PR #390, opened 2022-09-01
Debug "nation conquered" message even if nation still exists

LynxAbraxas:
This PR is meant to solve #388.

PR #409, opened 2023-08-27
debug uprising in city without units

LynxAbraxas:
PR #390 was lacking modifications in case the uprising succeeds in a city without units/garrison. This PR solves this and adds proper handling of non-land units for uprises.


LynxAbraxas:
Without baec4c1 this could happen:
just a coracle in Olympia before uprise:
ss_2023-08-27_14:13:44
after successful uprise new civ Jamaicans is created with remaining unit from slave uprising as garrison, still the coracle from the Greeks remains in the new civ's city (preventing Jamaican units to enter the city):
ss_2023-08-27_14:16:31
It is possible for the Greeks to move the coracle out of the city:
ss_2023-08-27_14:16:54

With baec4c1 the coracle is part of the defender army fighting the uprising slave army and in case the slave army wins no Greek units remain in the new civ's city:
ss_2023-08-27_14:43:01

MartinGuehmann:
This debug label has a wired meaning. Somehow it is something like WIP. So I guess you are now happy with it so that I can test it.

LynxAbraxas:
@MartinGuehmann  what do you think?

MartinGuehmann:
Compiles on Windows and Linux, and no obvious problems. So it can go in.

Issue #112, opened 2019-03-10
AI employ diplomats and take their success into account


LynxAbraxas:
So far I never observed that the AI tried to establish an embassy or hold a reception.
Due to this the AI does not know the scientific advances of other players and therefore cannot suggest diplomatic exchange of these.

MartinGuehmann:
So at least, the AI now builds diplomats with 31f0ae6. I still haven't seen the AI using them for holding receptions (throwing parties) or establishing embassies.

LynxAbraxas:
Really cool! I'm wondering though what "missionaries" are. I only know clerics and televangelists or is "missionaries" used for both, since televangelists replace clerics when advancing? I'd love to see some of the AI civs to stay in theology longer ideally till it can build televangelists. They just such a cool game concept special to CTP.

MartinGuehmann:
> Really cool! I'm wondering though what "missionaries" are. I only know clerics and televangelists or is "missionaries" used for both, since televangelists replace clerics when advancing? I'd love to see some of the AI civs to stay in theology longer ideally till it can build televangelists. They just such a cool game concept special to CTP.

I used missionary as a category, to put in those people. A televangelist, evangelizes or proselytizes, basically that is also what the cleric does here CTP2, even so a cleric is not necessarily somebody who evangelize. Simply these people spread the word of their religion, whatever that is.

If you wanna see the televangelists in the game then properly we should allow them either for all governments or at least put the televangelists into more modern governments.

LynxAbraxas:
> put the televangelists into more modern governments.

Hm, is there a modern form of theology? If not, how much work would it be to create one (that would possibly allow more cities).
Putting it in all govs probably renders theology uninteresting.

MartinGuehmann:
> > put the televangelists into more modern governments.
> 
> Hm, is there a modern form of theology? If not, how much work would it be to create one (that would possibly allow more cities).

There are modern theocracies such as Saudi Arabia and the Iran. But well, televangelists are not restricted to theocracies, you also find them in the USA, they are not really known here in Germany.

> Putting it in all govs probably renders theology uninteresting.

Theocracy is for me rather uninteresting, anyway. But of course that could be changed, by modifying it in other ways.

LynxAbraxas:
[...]

PR #410, opened 2023-09-03
Debug AI to build clerics or televangelists (missionaries)

LynxAbraxas:
This PR solves the bug that AI civs that are in theocracy were not able to  built clerics or televangelists (missionaries).
It's a change actually missing in https://github.com/civctp2/civctp2/commit/31f0ae688, relating to #225 and https://github.com/civctp2/civctp2/issues/112#issuecomment-488440836, #190 (#129), which could be closed now.
Still it would be interesting to make clerics or televangelists available to all govs, but at a higher cost or with less success chance.


MartinGuehmann:
Compiles on Windows and Linux. No obvious problems. And does what it claims to do. So can go in.

PR #411, opened 2023-09-29
Randomize target city when freeing slaves

LynxAbraxas:
So far the nearest city is the target when freeing slaves. This leads to oversized cities on the border of the empire. This PR randomizes the target city (like it is done when doing slave raids) as a simple measure to avoid this.


MartinGuehmann:
Compiles on Linux and Windows and runs, no obvious problems, so can go in.

Issue #412, opened 2023-10-15
AVCodecContext error when building on Debian12


rambo919:
> ui/aui_sdl/aui_sdlmovie.cpp: In function ‘int stream_component_open(VideoState*, int)’:
> ui/aui_sdl/aui_sdlmovie.cpp:1581:9: error: ‘AVCodecContext’ was not declared in this scope; did you mean ‘AVIODirContext’?
>  1581 |         AVCodecContext *avctx;
>       |         ^~~~~~~~~~~~~~
>       |         AVIODirContext
> ui/aui_sdl/aui_sdlmovie.cpp:1581:25: error: ‘avctx’ was not declared in this scope
>  1581 |         AVCodecContext *avctx;
>       |                         ^~~~~
> ui/aui_sdl/aui_sdlmovie.cpp:1589:46: warning: comparison of integer expressions of different signedness: ‘int’ and ‘unsigned int’ [-Wsign-compare]
>  1589 |         if (stream_index < 0 || stream_index >= ic->nb_streams)
>       |                                 ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
> ui/aui_sdl/aui_sdlmovie.cpp:1592:17: error: ‘avcodec_alloc_context3’ was not declared in this scope; did you mean ‘avio_alloc_context’?
>  1592 |         avctx = avcodec_alloc_context3(NULL);
>       |                 ^~~~~~~~~~~~~~~~~~~~~~
>       |                 avio_alloc_context
> ui/aui_sdl/aui_sdlmovie.cpp:1596:15: error: ‘avcodec_parameters_to_context’ was not declared in this scope; did you mean ‘avcodec_parameters_copy’?
>  1596 |         ret = avcodec_parameters_to_context(avctx, ic->streams[stream_index]->codecpar);
>       |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>       |               avcodec_parameters_copy
> ui/aui_sdl/aui_sdlmovie.cpp:1599:9: error: ‘av_codec_set_pkt_timebase’ was not declared in this scope
>  1599 |         av_codec_set_pkt_timebase(avctx, ic->streams[stream_index]->time_base);
>       |         ^~~~~~~~~~~~~~~~~~~~~~~~~
> ui/aui_sdl/aui_sdlmovie.cpp:1616:20: error: ‘avcodec_open2’ was not declared in this scope; did you mean ‘avio_open2’?
[...]

Issue #352, opened 2021-01-23
Building on Linux SDL missing


commel:
I am trying to build Master on openSUSE Leap 15.2. SDL2 header are installed:

```
Source package : SDL-1.2.15-lp152.5.3.1.src
Source package : SDL2-2.0.8-lp152.7.4.src
```

but the build still produces this:

```
ui/aui_sdl/aui_sdlmovie.cpp: In member function ‘virtual AUI_ERRCODE aui_SDLMovie::Resume()’:
ui/aui_sdl/aui_sdlmovie.cpp:2102:1: warning: no return statement in function returning non-void [-Wreturn-type]
 }
 ^
ui/aui_sdl/aui_sdlmovie.cpp: In member function ‘virtual AUI_ERRCODE aui_SDLMovie::Process()’:
ui/aui_sdl/aui_sdlmovie.cpp:2119:1: warning: no return statement in function returning non-void [-Wreturn-type]
 }
 ^
ui/aui_sdl/aui_sdlmovie.cpp: In member function ‘bool aui_SDLMovie::HandleMovieEvent(SDL_Event&)’:
ui/aui_sdl/aui_sdlmovie.cpp:2157:6: error: ‘toggle_pause’ was not declared in this scope
      toggle_pause(m_videoState);
      ^~~~~~~~~~~~
ui/aui_sdl/aui_sdlmovie.cpp:2164:8: error: ‘cursor_hidden’ was not declared in this scope
    if (cursor_hidden) {
        ^~~~~~~~~~~~~
ui/aui_sdl/aui_sdlmovie.cpp:2168:47: warning: suggest parentheses around comparison in operand of ‘&’ [-Wparentheses]
     if ((event.motion.type & SDL_BUTTON_LMASK != 0) &&
ui/aui_sdl/aui_sdlmovie.cpp:2169:7: error: ‘is_full_screen’ was not declared in this scope
      (is_full_screen || InsideMovieArea(event.button.x, event.button.y))) {
       ^~~~~~~~~~~~~~
ui/aui_sdl/aui_sdlmovie.cpp:2173:4: error: ‘cursor_last_shown’ was not declared in this scope
    cursor_last_shown = av_gettime_relative();
    ^~~~~~~~~~~~~~~~~
ui/aui_sdl/aui_sdlmovie.cpp:2173:24: error: ‘av_gettime_relative’ was not declared in this scope
    cursor_last_shown = av_gettime_relative();
[...]

PR #413, opened 2023-11-19
changes for FFmpeg-5 API (Dockerfile for building on Debian-12 (bookworm) and Ubuntu-23.10 (mantic))

LynxAbraxas:
This PR addresses the issues of #412 and #352 which arise due to the API changes from FFmpeg-4 (ubuntu:latest, currently 22.04, jammy) to FFmpeg-5 (debian:latest, currently 12.2, bookworm) possibly also for FFmpeg-6 (ubuntu:rolling, currently 23.10 mantic). It also serves as a demo on how to base the GL-CI test on debian instead of ubuntu.


LynxAbraxas:
This PR is marked WIP, because the changes for ubuntu:rolling (currently 23.10 mantic) compile (https://gitlab.com/LynxAbraxas/civctp2/-/pipelines/1077760343) but starting the game from the DI does neither work for the smoke tests (e.g.: https://gitlab.com/LynxAbraxas/civctp2/-/jobs/5569243554) nor locally due to new security measures of Xorg/Xvfb.

The commits upto [e0cff5d](https://github.com/civctp2/civctp2/pull/413/commits/e0cff5d3c6482ae83230dcf2756fcdcaeb3dce70) solve the issues of https://github.com/civctp2/civctp2/issues/412 and https://github.com/civctp2/civctp2/issues/352.

So far the base image for testing the build is ubuntu:latest, which does not point to the newest release of ubuntu (ubuntu:rolling), debian:latest being somewhere in between.
It would not be appropriate to use the static DI-tag ubuntu:23.04 (lunar) which would currently closer correspond to debian:latest.
So when ubuntu:latest becomes ubuntu:23.10, we'd need a solution for the problems with the smoke tests (concerning illegal access, the sound in Xvfb) or decide to build the GL-CI test on debian:latest (which at least used to be more "conservative" concerning packge versions and their dependencies).

LynxAbraxas:
@williamd67: Do you remember which exact versions of `ffplay.c` you used as a base for 80965a59f2e43bcf8f7da5d30c185ceff53d4f97? It would be very helpful to know which commit of `ffplay.c` was used (and stripped) then (and which version FFmpeg was the default at that time in ubuntu) to decide which changes from the current [`ffplay.c`](https://github.com/FFmpeg/FFmpeg/blob/fbc7adba6742c4d33ae525a72881f456281b4c46/fftools/ffplay.c) should be ported to [`aui_sdlmovie.cpp`](https://github.com/civctp2/civctp2/blob/67532c1338ad8c0f8e943a6e04633782543fa06a/ctp2_code/ui/aui_sdl/aui_sdlmovie.cpp). In the comment I just found this:
https://github.com/civctp2/civctp2/blob/67532c1338ad8c0f8e943a6e04633782543fa06a/ctp2_code/ui/aui_sdl/aui_sdlmovie.cpp#L1-L5

LynxAbraxas:
> This PR is marked WIP, because the changes for ubuntu:rolling (currently 23.10 mantic) compile (https://gitlab.com/LynxAbraxas/civctp2/-/pipelines/1077760343) but starting the game from the DI does neither work for the smoke tests (e.g.: https://gitlab.com/LynxAbraxas/civctp2/-/jobs/5569243554) nor locally due to new security measures of Xorg/Xvfb.

Turns out that this were not new security measures  but a SIGILL (illegal instruction) caused by a ud2 assembler instruction which is meant to fail (https://peeterjoot.wordpress.com/2010/05/26/a-fun-and-curious-dig-gcc-generation-of-a-ud2a-instruction-sigill/).
However, this time not due to a problem reported in d9b6fdc2 (though a similar one still existed upto 1867d37) but at:
https://github.com/civctp2/civctp2/blob/2037e527d6d79d840134e93e32f42a02852c4e1d/ctp2_code/ui/aui_sdl/aui_sdlui.cpp#L332

Since the build.log does not show any warnings for this file, it could be that the libsdl2 packages of ubuntu:mantic/nobel have this ud2 assembler instruction built into them. So either this is a bug there or newer libsdl2 ought to be used differently than currently implemented in the ctp2 code.
This should be tracked as its own issue (#417), because since May. 2024 ubuntu:latest points to ubuntu:nobel/22.04 and the monthly CI/CD test fail since then (https://gitlab.com/civctp2/civctp2/-/pipelines) due to the updated FFmpeg-5 API (which this PR is meant to address).
Therefore decided to stick to debian:latest/bookworm/12.7 here as base for the DIs and for testing with CI/CD.

LynxAbraxas:
PR ready for merging since current state compiles fine and passes all CI/CD tests (https://gitlab.com/LynxAbraxas/civctp2/-/pipelines/1447371428)

MartinGuehmann:
Checked it, compiles on my Kubuntu and also on Windows with Visual Studio 2017.

What an experience if you don't use Windows all day. It had to install three updates, of course one after the other, and after each Windows has to read start, which takes a while. Of course, after that some of the Windows spy stuff is switched on again. Windows explorer uses ribbons instead of a proper menu bar. The task bar suddenly got a search bar and whether app, of course without asking me whether I want that crap. And finally, the installer of Visual Studio forgot that Visual Studio was already installed, which caused some trouble using it. So I had to reinstall it. That is really unprofessional.

Anyway, this can go in.

PR #414, opened 2023-12-01
increase trade route value based on squared distance

LynxAbraxas:
This PR increases trade route values based on squared distances such that long trade routes become more precious in regard to their `valuePerCaravan`. This promotes longer trade routes, which in turn are more interesting concering the updated tiles along the route (since #256) and allows the AI to acquire more gold so it is can be more capable of affording special attacks and other cost. The PR also resolves a bug in #323.

### DIFF
--- a/ctp2_code/gs/gameobj/CityData.cpp
+++ b/ctp2_code/gs/gameobj/CityData.cpp
@@ -9453,8 +9453,6 @@ void CityData::FindGoodDistances()
 		m_distanceToGood[i] = 0 + specialdistance;
 	}
 
-//	g_theWorld->FindDistances(m_owner, m_home_city.RetPos(), goodsToFind,
-//	                          FindGoodDistancesCallback, this);
 	g_theWorld->FindDistances(m_owner, m_pos, goodsToFind,
 	                          FindGoodDistancesCallback, this);
 }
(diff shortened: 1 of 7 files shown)


MartinGuehmann:
That slows things down. I couldn't even get through 400 years of testing with automatic AI. The weird thing is when I undid the last commit it still felt kind of slow, even so I cannot see a reason for that in the code.

So for now it is just too slow to get in.

LynxAbraxas:
> That slows things down. I couldn't even get through 400 years of testing with automatic AI. The weird thing is when I undid the last commit it still felt kind of slow, even so I cannot see a reason for that in the code.

Hm, are you sure you used `--enable-quicktraderoutecalc` with `./configure` when compiling? Did you test for Linux or Win? Not sure where this configure option would have to go in Visual Studio configs.

MartinGuehmann:
> > That slows things down. I couldn't even get through 400 years of testing with automatic AI. The weird thing is when I undid the last commit it still felt kind of slow, even so I cannot see a reason for that in the code.
> 
> Hm, are you sure you used `--enable-quicktraderoutecalc` with `./configure` when compiling? Did you test for Linux or Win? Not sure where this configure option would have to go in Visual Studio configs.

How could I? You just pushed this commit a few days ago to GitHub, even so it is a almost a year old. What I did was to revert 8f67737ad78fcfeb0dee3c7fad33efed16dd4d92 and test. It still felt to be slower but I can test again maybe also on current master, which is faster anyway.

I am not sure whether I tested both, I am sure that I tested Windows. To add a new configuration to the VS project is overkill. I could also add it to the configurations but that is also a bit. The only thing I would do is modifying the code, directly, and enable it for testing. 

For now I would assume that the quick trade route calculation should be the default.

MartinGuehmann:
We had somewhere a modification of the Astar, it used a heuristics, which was removed. The rationale was that it did not change anything, I just didn't ask whether it was also true for speed.

PR #419, opened 2024-10-01
AI Slider Settings & Slider Setting Window

MartinGuehmann:
Calculate the slider settings a bit more accurately for the AI. While doing so cleaning up some code. And make the slider setting window display the values correctly.


MartinGuehmann:
So that is out for long enough and no complaints. So I guess that can go in.

PR #421, opened 2024-11-10
Fix the AI 5

MartinGuehmann:
A bunch of AI fixes with two crash fixes in between


MartinGuehmann:
So it's a month and no complaints, so let's merge.

PR #422, opened 2024-11-17
Fix regressions

MartinGuehmann:
Fix two regressions:

The army manager used to have a text field for the order the AI gave to the army when that debugging option is on. There is might be better visible than on the terrain.

The next unit button shows again the other units. It was not obvious that you get there by clicking onto the shield icon. It just does not look like a clickable button.


MartinGuehmann:
No complaints. Well then let's merge.

PR #423, opened 2024-11-17
Clean up for release

MartinGuehmann:
A few final thing to clean up for the first GitHub release


MartinGuehmann:
So no complains after a month then this can go in.

LynxAbraxas:
Sorry, had no time to take a look earlier. 
I'd suggest that with the introduction of `LICENSE.md` in 95cda99493f6f816ff the old `EULA - Source Code for CTP2.rtf` should be deleted such that there is only one license file in the repo to avoid confusion.

MartinGuehmann:
> Sorry, had no time to take a look earlier. I'd suggest that with the introduction of `LICENSE.md` in [95cda99](https://github.com/civctp2/civctp2/commit/95cda99493f6f816ffd67880e39cfd73000f4b91) the old `EULA - Source Code for CTP2.rtf` should be deleted such that there is only one license file in the repo to avoid confusion.

I am bit uneasy with that. The EULA is the original license file, and you don't really want to change that in any way. However, the rtf format is not compatible with browsers. The content should be identical anyway, however I still could add a disclaimer.

LynxAbraxas:
> I am bit uneasy with that. The EULA is the original license file, and you don't really want to change that in any way.

Well, it's it textual content that matters, not the formatting or the file format.
How about making `EULA - Source Code for CTP2.md` a softlink to `LICENSE.md` and removing `EULA - Source Code for CTP2.rtf`?

MartinGuehmann:
OK, I did that in #426.

PR #360, opened 2021-04-24
Radarmap improvements

williamd67:
Some small radar-map improvements

PR #425, opened 2024-12-27
Radarmap improvements (picked up from #360)

MartinGuehmann:
Radarmap improvements picked up from #360. I rebased it onto current master. So far I addressed all my concerns I raised in #360. The only problem left is that it slows down the game.

I narrowed it down to redrawing borders.


MartinGuehmann:
I got sped up, but I still have some holes in the borders. They are caused by units passing them and they are not fully redrawn.

Then territories get merged if you show the diplomatic status. This is fine for the fill out areas as this was the case before, but without the areas the borders are missing between civs with the same status. And the code was refactored in a way, that it is not trivial to put that in. At least if you want to solve it to some extend properly.

MartinGuehmann:
I am down to the holes in the borders. But it is really hard to find where the problem is. It rather happens that I fix other bugs on the way.

MartinGuehmann:
So I have nothing to add to this. So merging this for the final test before release.

PR #426, opened 2024-12-28
Clean up for release2

MartinGuehmann:
Some more things that should go in before release.


LynxAbraxas:
What is the exclusive mode for windows?

MartinGuehmann:
> What is the exclusive mode for windows?

It is an option in DirectX. Basically, nothing can be drawn on the screen that does not come from CTP2. Very lovely if CTP2 hangs than you cannot even alt-tab out, you may still call the task manager, but you can't see CTP2 there to kill it. Basically, this way you can only log out and log in.

The other thing is, it gave me were some visual artifacts , which are not there in non-exclusive mode. When I got freezes of CTP2, I started it from VS only in non-exclusive mode. So that I didn't see the artifacts. A bit weird is that I can't remember the artifacts from before. I also think that this wasn't a problem on Windows 98, XP, and 7.

MartinGuehmann:
I think we have now everything, so merging it for the final test.

LynxAbraxas:
@MartinGuehmann
Thanks for the new release.
NB: Please use git cherry-pick in the future if making use of changes already provided as a commit to avoid loosing the info given in the commit message, i.e. https://github.com/civctp2/civctp2/pull/428/commits/26747c3c155139cac1784f82d8f834d9f00b854d adjust the git commit-message if necessary to point to e.g. the WIP PR that contains the commit, i.e. #428.

PR #427, opened 2024-12-30
Improve viewing of "stand-alone" effect animations

LynxAbraxas:
PR to address suggestions from #386, in particular https://github.com/civctp2/civctp2/pull/386#issuecomment-2560027109:

- without "stand-alone" effect animations are hardly ever visible due to other changes of map center
- projectile animations (net thrown when enslaving, etc) should play together with (special) attack animations (slaver moving to throw net), so they should not lock

### DIFF
--- a/ctp2_code/gfx/spritesys/director.cpp
+++ b/ctp2_code/gfx/spritesys/director.cpp
@@ -331,7 +331,7 @@ private:
  *
  * The combination of the two behaviours will create four kind of actions:
  *  - No Lock && No Animation -> immediate
- *  - No Lock && Animation    -> effect
+ *  - Lock    && Animation    -> effect (EffectActor, similar to active now that locking was enabled here as well)
  *  - Lock    && No Animation -> external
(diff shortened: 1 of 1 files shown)


MartinGuehmann:
I think what you could do is to have it deactivated if unit animations are off. Then it should not slow down my AI tests.

ctp2_code/gfx/spritesys/director.cpp
```diff
@@ -447,7 +447,7 @@ class DQActionEffect : public DQAction {
 
 	// Lock behaviour
 	virtual void Unlock() {}
-	virtual bool IsUnlocked() { return true; }
+	virtual bool IsUnlocked() { return IsAnimationFinished(); }
```
LynxAbraxas, line 450: @MartinGuehmann do you think the delay in your AI test originates from this change?

MartinGuehmann: It could be that. I guess that is something to try out.

PR #431, opened 2025-02-20
Windows64 SDL version

MartinGuehmann:
Sooner or later Windows will pull 32 bit support. So it is time to have a Windows 64 bit version. The paths to that might be either to replace the miles and more sound library by something else or go fully SDL.

So far Visual Studio compiles the SDL version as 32 bit. I cleaned up the preprocessor directives. Basically any input, output, and graphics related code, that has an SDL counterpart should be treated as DirectX code even so it is just Windows.

The single player game seems to work as on Linux. Multiplayer got me a bit longer to compile everything, but I expect it to work. However, to get it to 64 bit a Visual Studio project for the Anet library is missing. Also videos don't play, yet. And I still have to separate the 32 bit assembly code from the other Windows code.

It seems to be a good idea to go for SDL, since with it the game loads faster. I don't know exactly whether it runs faster. The other advantage is that you don't have to implement the music for the GOG version. However you should not have files that are Linux links in your music folder.


MartinGuehmann:
Right now we have 32 and 64 bit SDL compiling, linking and starting. Also 64 bit DirectX, however it shuts down when it can't find 16 bit display modes.

Everything now is in place, videos are playing, ok I just checked the intro movie. :D The libraries for multiplayer, however no multiplayer due to bugs. And a few more bugs, not only those known from the Linux version but also the the Cheat editor does not open in the 64 bit SDL version.

MartinGuehmann:
This PR can now be merged. The windows SDL version is now working. It turned out that the SDL Linux port was not complete. And some stuff from the user interface was missing. MP is now working on Windows, however not on Linux. That would be still nice to get into it before merging.

LynxAbraxas:
@MartinGuehmann  I wonder why you added e.g. the FFmpeg code ([4e023a3](https://github.com/civctp2/civctp2/pull/431/commits/4e023a3eaf6306f1e676b6581b2ce5c7db870c4f)) to the repo instead of adding that as a git submodule pointing to that release? Using a git submodule would avoid bloating the code base of ctp2 and would allow to be sure that that ffmpeg code is not used e.g. on linux.
Should I create a PR to change to git submodules for all libs which are available over git-repos?

MartinGuehmann:
> @MartinGuehmann I wonder why you added e.g. the FFmpeg code ([4e023a3](https://github.com/civctp2/civctp2/pull/431/commits/4e023a3eaf6306f1e676b6581b2ce5c7db870c4f)) to the repo instead of adding that as a git submodule pointing to that release? Using a git submodule would avoid bloating the code base of ctp2 and would allow to be sure that that ffmpeg code is not used e.g. on linux.

At that point I didn't really know about submodules ans later I didn't change it. And it gives us a bit more control about the code. However, conceptionally you are right.

> Should I create a PR to change to git submodules for all libs which are available over git-repos?

In principle yes. However, there is a FFmpeg-n6.1.2/SMP/ subdirectory, which I created to allow Visual Studio to compile it. I couldn't find lib and dll files for that, so in the end I decided to compile it from source. The other idea was to be able to compile ARM64 versions, which unfortunately I can't because one build step is creating assembler files matching the platform. Unfortunatly, the assembly file make tool can only create 84x/64x assembly code. At least that is how I understood it.

Ideally, the SMP subfolder should stay in place, alternatively it can be moved one folder up but that the paths in the .vcxproj files in that folder and the paths in the /ctp2_code/ctp/ctp2.sln file. This can be done with a text editor.

In one of the other libraries, they changed the interfaced. I did not bother how to adapt the ctp2 code to that, but just changed that back to one of the previous versions. And later I did not bother to change that back. That's actual also only something specific to the windows code.

I think freetype-1.3.1 was already added back during the early Linux development on GitHub. We use it on Linux and I also added it to Windows. I think I also created the VS project file for it.

LynxAbraxas:
I see. Well another option to git-submodules would be git-subtree, which imports the history of the sub-repo, optionally squashed, keeps track of the commit of the sub-repo and is better suited to make adjustments to the imported code specific for the super project.
Yes freetype was included (I think) before git came into existence.

Since all is included into master already, I would say this topic just should be kept in mind when new modifications of code in included sub-repos is needed. The more changes that accumulate in sub-repo code the more difficult will it be to change to git-submodules or git-subtrees in future (noticed that while trying to port the ffplay code snippets to newer ffmpeg in #440 and in particular https://github.com/civctp2/civctp2/pull/440/commits/d1e5ab7e908d0c20a41eac8a1b0d5e992dbd3d85).

PR #434, opened 2025-04-01
Flatpak Build support

another-sapiens:
This PR adds a flatpak manifest and runner script,  and instructions to build the game using flatpak. This should allow to streamline the building process across Linux distributions and avoid a mismatch between libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,22 @@ The executables these configurations generate are in [your source path]/ctp2_cod
 
 CTP2 can directly be run from Visual Studio, to do so press F5 if you want to run it on the debugger, which will be slow and does not make sense for a Final version. To run without debugging, press Ctrl F5. If necessary, CTP2 will be built, first. Some of the map plugins may fail to build the first time in a clean project. Just try compile those again.
 
-## Building on Linux
+## Building on Linux...
+
+**Using flatpak**
+
+Flatpak should streamline the process, as the libraries are managed by the manifest, and the commands are the same on all distros.
+
+To rehearse:
+
+1. Clone this repository and unzip it
+2. Paste the CD/GoG game data over the unziped directory, _without replacing_ the downloaded repo files
+3. Run `flatpak install org.flatpak.Builder org.freedesktop.Sdk/x86_64/24.08 org.freedesktop.Platform/x86_64/23.08` to satisfy the dependencies.
+4. Run `flatpak run org.flatpak.Builder --install --user --force-clean "build/" "net.apolyton.civctp2.yaml"`. This will take a while to compile everything.
+5. Finally, `flatpak run net.apolyton.civctp2` will run the game!
+
+
+**Using the autogen as-is**
 
 You will probably need GCC 5.x or later to build. The code doesn't seem to build on GCC 4.8.
 
--- a/net.apolyton.civctp2.yaml
+++ b/net.apolyton.civctp2.yaml
@@ -1,0 +1,73 @@
+app-id: net.apolyton.civctp2
+runtime: org.freedesktop.Platform
+runtime-version: '23.08'
+sdk: org.freedesktop.Sdk
+command: run-ctp
+finish-args:
+  - --share=ipc
+  - --socket=wayland
+  - --socket=fallback-x11
+  - --socket=pulseaudio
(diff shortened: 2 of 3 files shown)


MartinGuehmann:
The first thing I got was:
```
error: org.freedesktop.Sdk/x86_64/23.08 not installed
Failed to init: Unable to find sdk org.freedesktop.Sdk version 23.08
```
However, the instruction to install the dependencies just include: 
org.freedesktop.Sdk/x86_64/24.08
org.freedesktop.Platform/x86_64/23.08

But not: 
org.freedesktop.Sdk/x86_64/23.08

Question: Is only org.freedesktop.Sdk/x86_64/23.08 necessary?

Once I had installed this, I went on and got this:

```
In file included from /usr/include/fcntl.h:342,
                 from main.c:16:
In function ‘open’,
    inlined from ‘my_mkstemp’ at main.c:529:7,
    inlined from ‘open_tmpfile’ at main.c:583:7:
/usr/include/x86_64-linux-gnu/bits/fcntl2.h:50:11: error: call to ‘__open_missing_mode’ declared with attribute error: open with O_CREAT or O_TMPFILE in second argument needs 3 arguments
   50 |           __open_missing_mode ();
      |           ^~~~~~~~~~~~~~~~~~~~~~
/usr/include/x86_64-linux-gnu/bits/fcntl2.h:50:11: error: call to ‘__open_missing_mode’ declared with attribute error: open with O_CREAT or O_TMPFILE in second argument needs 3 arguments
```

That come from compiling byacc. Actually, this is not necessary since I can install byacc via app, the same is also true flex. Of course you can still download and install those. However, I don't see the other dependencies in the script.

Another odd thing is that you copy the folder ctp2_code. That's not needed, you only have to copy the executable, the Linux map file, and map plugins to their locations in the ctp2_program folder.

another-sapiens:
Runtime 23.08 is now end-of-life, so I updated to runtime-version 25.08, which gave me some trouble.

The runtime/SDK dependency itself can be taken care of by adding `--install-deps-from=flathub`. So the final command to build is;
`flatpak run org.flatpak.Builder  --install-deps-from=flathub --install --user --force-clean "build/" "net.apolyton.civctp2.yaml"`
so we can skip that separate installation step.

As for the other dependencies, you ask why I included byacc and not the others. This was because the SDK already ships most of the ones we needed. However it seems since updating from 23 to 25 we run into trouble, some deps were trimmed and some seem to be too new? Whatever the cause, we do need to include them ourselves now. Here's what I did for each:

- byacc: I now just use a .deb and extract the binary to reduce compile times. The SDK has bison but it really does have to use byacc.
- ffmpeg: The SDK has it but it doesn't compile, may be too new? I disable everything and only enable file, avi, ogg, indeo5, pcm_s16le and vorbis. Exactly what's needed can be determined by running ffprobe on the files we need to open (.avi movies and .ogg tracks)
- SDL2_image and mixer: I think this was trimmed from the 25.08 SDK and only the base SDL is there, so we compile it.

For the module that builds the actual game, I now source it from git instead of the "current directory". This means that, in theory, you could just download the .yaml manifest, paste it where the game data is, and when built it will automatically clone this repository. Regarding the cflags and cxxflags, I had to remove `-fuse-ld=gold` because the gold linker seems to have been removed from the latest SDKs. And I had to add -std=gnu89; not really sure about this, I tried not adding it or adding others like c11, gnu99, etc. But it only works like that now, don't ask me why. I also specify `--build=x86_64-pc-linux-gnu` because I was getting errors with freetype.

Lastly for the game-data module, I now only use `ctp2_data`, `ctp2_program` and `Scenarios` directories as sources, and included the runner script inside the manifest instead of a separate file so it's cleaner.


With all this, the game finally "works" again on 25.08, as it compiles, opens and somewhat plays. But I can't save, and if a wonder movie ends on its own the game crashes. Not sure what I'm doing wrong.

Honestly I might be punching a bit above my weight here. I'm not really familiar with C++ and I had to recompile more than I'd like to admit. I did what I could, but I think I'll leave it gere for now. Hopefully someone more knowledgeable/patient will find this useful in the future. 

If you're reading this in the future and think this is you, feel free to use my .yaml and re-open this.

PR #439, opened 2025-12-20
More cleanup

MartinGuehmann:
Frankly, I don't have a good idea how to call it. Just a few more things to fix before release.

A few more compiler warning fixes, domestic trade routes now have a higher value then 1 (some bug), and trade routes can again need more than one caravan. Probably, a few more things.


LynxAbraxas (approved):

MartinGuehmann:
Make this mergeable again. So let's merge this.

PR #443, opened 2026-01-11
avoid extra centering on map during "file injunction"

LynxAbraxas:
change missing in #386


MartinGuehmann:
OK, that's fine. Let's merge.

PR #445, opened 2026-06-29
Compensate for SDL 2.x bug and prevent a double free

### LINKED ISSUE #438: I get a segfault while first start
Hi, 

First I want to thank for your hard work to make this possible. 

So the problem is, after I compiled everything and wanted to start it for the first time I've got 
an "bodo@donald:~/CTP2/ctp2_code/ctp> ./ctp2
        Segmentation fault (core dumped)" in the terminal. 

I'm running it or tried to run it on an Opensuse 15.6-64 bit machine. 
I have no idea how to fix it. 
What can I do to give you more info about this problem? 

The message above is all I get in the terminal. 

Thx
Bodo

### LINKED ISSUE #441: spurious crashes since #437 was merged
Collection of spurious (not reproducible) crashes/hangs. #437 was suspected to cause these, but possibly they originate from library updates in ubuntu/debian since about mid 2025.

claw-333:
The `CivSound` constructor has SDL load an audio file into an SDL `Mix_Chunk`. After loading the audio data, SDL 2.8.2 resizes the chunk buffer with `realloc()`. If the audio content is zero length, then `realloc(ptr, 0)` may be equivalent to `free(ptr)`. As a result, the `Mix_Chunk` returned to `CivSound` has 0 in the `alen` field and a pointer to free memory in the `abuf` field. Later, the freed pointer is incorrectly freed again by SDL's `Mix_FreeChunk`.

To prevent SDL from freeing twice, this commit sets `abuf` to `NULL` when `alen = 0`.

The behavior of `realloc(ptr, 0)` is implementation dependent. `realloc(ptr, 0)` may be implemented to return a valid pointer, in which case this commit will result in a memory leak; or, it may return `NULL` and _not_ free `ptr`. While we can distinguish between some of these behaviors in the configure script, determining SDL's behavior is challenging. Setting `abuf` to NULL is the conservative approach: a memory leak is better than a double free.

### DIFF
--- a/ctp2_code/sound/civsound.cpp
+++ b/ctp2_code/sound/civsound.cpp
@@ -77,14 +77,13 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
     if (!fname) {
         return;
     }
-#if defined(USE_SDL)
+
 	// "NULL.WAV" contains no audio data, and with this file SDL 2.x
 	// returns a pointer that it has already freed, which it later
 	// tries to free again in Mix_FreeChunk.
 	if (strcasecmp(fname, "NULL.WAV") == 0) {
 		return;
 	}
-#endif
 
     size_t      l_dataSize = 0;
     m_dataptr   = g_SoundPF->getData(fname, l_dataSize);
--- a/ctp2_code/sound/civsound.cpp
+++ b/ctp2_code/sound/civsound.cpp
@@ -59,6 +59,11 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
     m_associatedObject(associatedObject)
 
 {
+	// Initialize fields in case of early return.
+	m_soundFilename[0] = 0;
+	m_dataptr = NULL;
+	m_datasize = 0;
+
     const char *fname;
 	if(soundID < 0)
 		fname = NULL;
@@ -70,9 +75,6 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
     m_isLooping = FALSE;
 
     if (!fname) {
-        m_soundFilename[0] = 0;
-        m_dataptr = NULL;
-        m_datasize = 0;
         return;
     }
 #if defined(USE_SDL)
@@ -80,18 +82,18 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
 	// returns a pointer that it has already freed, which it later
 	// tries to free again in Mix_FreeChunk.
 	if (strcasecmp(fname, "NULL.WAV") == 0) {
-		m_soundFilename[0] = 0;
-		m_dataptr = NULL;
-		m_datasize = 0;
 		return;
 	}
 #endif
 
-    strcpy(m_soundFilename, fname);
-
     size_t      l_dataSize = 0;
-    m_dataptr   = g_SoundPF->getData(m_soundFilename, l_dataSize);
-    m_datasize  = static_cast<sint32>(l_dataSize);
+    m_dataptr   = g_SoundPF->getData(fname, l_dataSize);
+	if (!m_dataptr) {
+		return;
+	}
+
+    strcpy(m_soundFilename, fname);
+    m_datasize = static_cast<sint32>(l_dataSize);
 
 #if !defined(USE_SDL)
 	m_hAudio = AIL_quick_load_mem(m_dataptr, m_datasize);
--- a/ctp2_code/sound/civsound.cpp
+++ b/ctp2_code/sound/civsound.cpp
@@ -87,7 +87,14 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
     // Argh, audio format mismatch!!!
 	m_Audio = Mix_QuickLoad_RAW((Uint8 *) m_dataptr, (Uint32) m_datasize);
 # else
-    m_Audio = Mix_LoadWAV_RW(SDL_RWFromMem(m_dataptr, static_cast<int>(m_datasize)), 1);
+	m_Audio = Mix_LoadWAV_RW(SDL_RWFromMem(m_dataptr, static_cast<int>(m_datasize)), 1);
+	// Compensate for possible SDL bug: if alen=0, then abuf may have been freed by
+	// realloc inside SDL. Since the behavior of realloc(ptr, 0) is implementation
+	// dependent, conservatively assume that realloc has freed the memory, and set abuf
+	// to NULL to prevent it from being freed again by SDL's Mix_FreeChunk.
+	if (!m_Audio->alen) {
+		m_Audio->abuf = NULL;
+	}
 # endif
 #endif
 }
--- a/ctp2_code/sound/civsound.cpp
+++ b/ctp2_code/sound/civsound.cpp
@@ -36,8 +36,6 @@
 //
 //----------------------------------------------------------------------------
 
-#include <strings.h>
-
 #include "c3.h"
 
 #include "civsound.h"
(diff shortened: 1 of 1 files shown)


ctp2_code/sound/civsound.cpp
MartinGuehmann: This returns NULL if the file does not exists. So m_dataptr should be checked for NULL and if NULL then it should return like if 0 == fname above.

ctp2_code/sound/civsound.cpp
```diff
@@ -87,7 +87,14 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
     // Argh, audio format mismatch!!!
 	m_Audio = Mix_QuickLoad_RAW((Uint8 *) m_dataptr, (Uint32) m_datasize);
 # else
-    m_Audio = Mix_LoadWAV_RW(SDL_RWFromMem(m_dataptr, static_cast<int>(m_datasize)), 1);
+	m_Audio = Mix_LoadWAV_RW(SDL_RWFromMem(m_dataptr, static_cast<int>(m_datasize)), 1);
+	// Compensate for possible SDL bug: if alen=0, then abuf may have been freed by
+	// realloc inside SDL. Since the behavior of realloc(ptr, 0) is implementation
+	// dependent, conservatively assume that realloc has freed the memory, and set abuf
+	// to NULL to prevent it from being freed again by SDL's Mix_FreeChunk.
+	if (!m_Audio->alen) {
```
MartinGuehmann: This should not happen if we leave the function early when m_dataptr is NULL. So `Assert(false);` should go in in case there is another problem, which we can catch in the debug version.

MartinGuehmann:
I think the function should not even try to allocate memory if the file does not exist. Potential memory leaks aren't that nice either. And if we just leave we should be fine. I added two comments to the code above to direct you.

And it should undo commit 70379053bca406fb7da914de7f4290cc4884f1cf, the thing with disabling the optimization was because of this issue.

ctp2_code/sound/civsound.cpp
claw-333: Done.

ctp2_code/sound/civsound.cpp
claw-333: Done.

claw-333:
I agree, however the result is not null. Although “NULL.WAV” is not in the filesystem, it is in the “ctp2_data/default/sound/sound.zfs” archive as “null.wav” (lookup is case insensitive). So the "missing" file does exist, it just doesn't have any sound data.

This appears to be the only such file, so I hard-coded the troublesome filename, and only for SDL.

Other possible approaches:

- Add “pretend this file doesn’t exist” method to ProjectFile, and call it when initializing g_SoundPF.
- hack sound.zfs to remove "null.wav".
- hack sound.zfs to add trivial non-empty sound data to "null.wav".

> And it should undo commit https://github.com/civctp2/civctp2/commit/70379053bca406fb7da914de7f4290cc4884f1cf, the thing with disabling the optimization was because of this issue.

It's not clear to me that the changes in this PR are related to compiler optimizations.
I crashed here with optimizations disabled. Instead, I suspect optimizations made it more likely for the heap damage to affect something that mattered. If double frees trigger that behavior, then there are several other possible triggers. With these fixes, and other pending double-free fixes elsewhere, I have been using -O2 and -O3 reliably.

Also, https://github.com/civctp2/civctp2/pull/449 should be committed before re-enabling optimizations, because it is clearly related to optimizations.

ctp2_code/sound/civsound.cpp
```diff
@@ -70,28 +75,25 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
     m_isLooping = FALSE;
 
     if (!fname) {
-        m_soundFilename[0] = 0;
-        m_dataptr = NULL;
-        m_datasize = 0;
         return;
     }
 #if defined(USE_SDL)
```
MartinGuehmann: Even if the MMS/DirectX version may handle this correctly, it is still not a good design. So this code should be used by the MMS/DirectX version.

MartinGuehmann:
> Other possible approaches:
> 
> * Add “pretend this file doesn’t exist” method to ProjectFile, and call it when initializing g_SoundPF.
> * hack sound.zfs to remove "null.wav".
> * hack sound.zfs to add trivial non-empty sound data to "null.wav".

I think the problem is solved. No need to complicate it more.

> > And it should undo commit [7037905](https://github.com/civctp2/civctp2/commit/70379053bca406fb7da914de7f4290cc4884f1cf), the thing with disabling the optimization was because of this issue.
> 
> It's not clear to me that the changes in this PR are related to compiler optimizations. I crashed here with optimizations disabled. Instead, I suspect optimizations made it more likely for the heap damage to affect something that mattered. If double frees trigger that behavior, then there are several other possible triggers. With these fixes, and other pending double-free fixes elsewhere, I have been using -O2 and -O3 reliably.

Well, the commit message says: disable optimization which causes spurious segfaults, double free and other crashes (tested -O1, -O2, -O3):

Well, that was a suspicion at that time. as far as i can see the segfaults where all sound related and that is what this PR adresses.
 
> Also, #449 should be committed before re-enabling optimizations, because it is clearly related to optimizations.

So far I only encountered one problem with pools and nulling that was optimized away and made trouble, for that I just disabled gcc optimization for the destructor. So far I haven't looked so deeply onto PR #449. But when to merge it shouldn't be WIP anymore.

MartinGuehmann:
OK, I just checked #449 is another pool than the one that made trouble but of course this way is the cleaner design and all pools then should be treated like that.

claw-333:
> OK, I just checked #449 is another pool than the one that made trouble but of course this way is the cleaner design and all pools then should be treated like that.

Removed "WIP" from https://github.com/civctp2/civctp2/pull/449. Which was the other pool?

ctp2_code/sound/civsound.cpp
claw-333: Done.

ctp2_code/sound/civsound.cpp
```diff
@@ -36,6 +36,8 @@
 //
 //----------------------------------------------------------------------------
 
+#include <strings.h>
```
MartinGuehmann: Visual Studio does not know that.

ctp2_code/sound/civsound.cpp
```diff
@@ -57,28 +59,40 @@ CivSound::CivSound(const uint32 &associatedObject, const sint32 &soundID)
...
-        m_datasize = 0;
+    if (!fname) {
         return;
     }
 
-    strcpy(m_soundFilename, fname);
+	// "NULL.WAV" contains no audio data, and with this file SDL 2.x
+	// returns a pointer that it has already freed, which it later
+	// tries to free again in Mix_FreeChunk.
+	if (strcasecmp(fname, "NULL.WAV") == 0) {
```
MartinGuehmann: And again this is not portable, the Visual Studio compiler refuses to compile this. Must the comparison really case sensitive?

MartinGuehmann:
Well, this does not compile on Windows, so it can't go in.

ctp2_code/sound/civsound.cpp
MartinGuehmann: And actually, this should also go after the general header c3.h

claw-333: Done. I removed this include file. gcc doesn't complain without it here, so it must already be included.

ctp2_code/sound/civsound.cpp
claw-333: Done. I was being cautious because the archive look-up is case-insensitive. However, there is no case-insensitive comparison in the common library, and creating one just for this case seems excessive. Added a comment.

claw-333:
Fixed. Please try again.

MartinGuehmann:
- [x] Compiles on Windows
- [x] Compiles on Linux
- [x] No obvious problems on Windows
- [x] No obvious problems on Linux

So can go in.

PR #447, opened 2026-07-02
Fix implicit function decl warnings in yacc code

claw-333:
lex and yacc do not declare all of their external functions, and this wasn't a problem with old C compilers. Modern compilers are pickier.

The implicit function warning was disabled for the MS compiler in the relevant files; however, gcc considers this an error. This commit disables the error for gcc.


claw-333:
Removed "WIP" because this change blocks successful building of dependent PRs.

MartinGuehmann:
This compiles on Linux and Windows, and no problems during AI turns. So let's merge.

PR #449, opened 2026-07-08
Assign null to pointer in destructor

claw-333:
Compiler optimizations skip assigning NULL to object fields in the destructor because it assumes the destructor will only be called once, and therefore no code will ever look at the modified field.

In this case, Pool manages the memory, and the memory is not actually freed. When the program quits, the destructor is called again, and the pointer in the unmodified field is freed again. "volatile" prevents the compiler from skipping the assignment.

### DIFF
--- a/ctp2_code/gs/slic/SlicSegment.cpp
+++ b/ctp2_code/gs/slic/SlicSegment.cpp
@@ -262,6 +262,41 @@ SlicSegment::SlicSegment(CivArchive &archive)
 	Serialize(archive);
 }
 
+namespace {
+
+//----------------------------------------------------------------------------
+//
+// Name       : ClearPointerField
+//
+// Description: Helper function for destructor
+//
+// Parameters : pointer of any type T
+//
+// Globals    : none
+//
+// Returns    : none
+//
+// Remark(s)  : Assigns NULL to a pointer, even with compiler optimizations
+//              enabled.
+//
+//              With optimizations, the compiler may skip the assignment in
+//              the destructor because it assumes the destructor will only be
+//              called once, and therefore no code will ever look at the new
+//              value.
+//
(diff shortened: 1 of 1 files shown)


claw-333:
Removed "WIP" per https://github.com/civctp2/civctp2/pull/445#issuecomment-4918323553

MartinGuehmann:
> Removed "WIP" per [#445 (comment)](https://github.com/civctp2/civctp2/pull/445#issuecomment-4918323553)

Did you remove it because of the comment or did you remove because of the comment and you think it can go in?

claw-333:
I removed it because the comment said "But when to merge it shouldn't be WIP anymore", and I think it can go in.

Apologies if I misunderstood the comment and/or the "WIP" protocol.

MartinGuehmann:
> and I think it can go in.

Well, that is the important part. i just need to find some time to check it. Basically, i run the game for a few 100 turns in AI only mode to see everything is fine. I don't expect here any surprises but well I already had surprises in the past. Of course, afterwards I merged code that was looking fine.

And there is still 70379053bca406fb7da914de7f4290cc4884f1cf, I don't care into which PR the undoing goes.

claw-333:
I appreciate the time and effort needed for vetting changes.

> And there is still https://github.com/civctp2/civctp2/commit/70379053bca406fb7da914de7f4290cc4884f1cf, I don't care into which PR the undoing goes.

I will create an undo PR after this PR and https://github.com/civctp2/civctp2/pull/445 are committed.

MartinGuehmann:
By the way you don't have to merge master into this branch every time that gets updated. This clutters just the version history with the branches. of course, this may depend also on your git client.

MartinGuehmann:
This compiles on Windows and Linux, and no obvious side effect. So let's merge.

PR #450, opened 2026-07-16
Fix astar and other things along

MartinGuehmann:
## Summary

Started as a fix for an A* endless-loop/heap-corruption bug in the pathfinding code, then grew into a broader pass over crashes and performance issues found while actually playing the game (profiled with `perf` across multiple play sessions) and a subsequent read-through of `ai/strategy/goals/`. All fixes were verified individually (targeted testing, ASan/UBSan scratch programs for the more invasive data-structure rewrites, and multi-hundred-turn play sessions) before being folded in here. This work was done with Claude Code including this description.

### A* pathfinding
- Fix heap corruption in `DAPriorityQueue::Remove` for non-root indices.
- Fix an operator-precedence bug that let A* searches silently ignore their node cutoff.
- Add a reopen-count assert to help catch future endless A* loops early.
- Scale `RobotAstar2`'s and the general A* node cutoff with map size (`Astar_MaxSearchNodes()`), instead of a fixed constant that could be too small on large maps or too large on small ones.
- Fix `Path::FlattenAstarList`'s endless-loop check firing far too late.
- Fix an uninitialized read in `UnitAstar::ComputeValidMoveCost`.
- Delete unused `AstarQ.cpp/.h` and `BFS.cpp/.h` (dead code, no longer referenced anywhere).
- Fix some GCC compiler warnings surfaced while working through this code.

### Crash fixes (found while playtesting)
- Fix null-pointer crash in `GreatLibrary::SortByAgeCallback`.
- Fix buffer overflow reading/writing `netf.def` in `NETFunc` (a combined-size `fread`/`fwrite` call was overrunning the destination buffer; split into per-field calls).
- Fix out-of-bounds access when a new player reuses a dead player's slot.
- Harden `main_CivExceptionHandler` against deadlocking when a signal (SIGSEGV etc.) fires while the process is already inside `malloc`, by making the handler's fallback path async-signal-safe.

### Performance (found via `perf` profiling)
- Fix `aui_DirtyList::Minimize`'s rect-merging algorithm, which was O(n³) (it restarted its scan from the head of the list after every merge); now O(n²).
- Replace `tech_Memory<T>`'s linear "used-slot" scan with an O(1) intrusive free list. This was the single largest self-time hotspot in the first profiling pass (60%+), used by several core containers (`aui_DirtyList`, `tech_WLList`, `aui_UI`'s dirty-rect tracking, ...).
- Wire up `QuadTree`'s garbage collector, which existed and was fully implemented but was never actually called from anywhere. Without it, `Vision::m_unseenCells` (the fog-of-war quadtree) accumulated degenerate single-child branches without bound over a long session. Now invoked once per turn from `Player::BeginTurn`.

### Found via code audit (not a reported crash)
- Fix out-of-bounds/null `g_player[target_owner]` access in `Goal::IsTargetImmune` (three separate call sites: Great Wall immunity, incite-revolution pretest, steal-technology pretest — each could dereference `g_player[]` with an unassigned/invalid index). Found by reading through `ai/strategy/goals/` with Claude Code rather than from an observed crash.

### Investigated, no fix needed
- `GameObj`/`ObjPool`'s hash-of-BSTs object lookup structure: ids are assigned in strictly increasing order per object type, so each of the 1024 hash buckets degenerates into a linear chain rather than a balanced tree — but bucket count bounds chain length to (live objects of that type)/1024, which stays small for realistic object counts. Confirmed via profiling to be a negligible, appropriately-scaled cost (~0.83% self-time), not a bug.
- A reported SDL-related hang/lag (`Ctrl+C`'d out of, not reproducible on reload) was traced through `aui_SDLUI::SDLDrawScreen`, `aui_UI::Process`/`Draw`, and `CivApp::Process`; no unbounded loop or obvious defect found on that path.

## Test plan
All items below have already been tested and verified:
- [x] A* pathfinding: confirmed no more endless loops/heap corruption in extended play (robot and human players, multiple map sizes).
- [x] `tech_Memory<T>` rewrite: verified with an ASan/UBSan/LeakSanitizer scratch program against realistic usage patterns (including `tech_WLList`) before landing; 600-turn play session with no crashes afterward.
- [x] `aui_DirtyList::Minimize` and `QuadTree` GC: confirmed reduced/eliminated hotspots via before/after `perf` comparison across sessions of increasing length.
- [x] Crash fixes: each reproduced from an actual in-game crash/stack trace, then confirmed gone on replay from the same or a fresh save.
- [x] Compiles on Linux
- [x] Compiles on Windows


MartinGuehmann:
Changed to WIP, I still have to compile and test it on Windows.

MartinGuehmann:
Now also checked on Windows.

MartinGuehmann:
So no complaints, then it is time to merge.

PR #451, opened 2026-07-26
Revert commit that disabled optimizations.

claw-333:
This reverts commit 70379053bca406fb7da914de7f4290cc4884f1cf:
 "disable optimization which causes spurious segfaults, double free and
  other crashes (tested -O1, -O2, -O3)"

Fixed in https://github.com/civctp2/civctp2/pull/445


MartinGuehmann:
OK, looks good, and can go in.

PR #452, opened 2026-07-28
Fix scheduler/pathfinding thrashing and stuck-flag bugs found via /rnd stress-testing and chasing asserts

MartinGuehmann:
## Summary

Most of these were found by chasing down the root cause of Asserts firing in debug builds during extended `/rnd` (debug auto-turn) sessions, rather than from a generic crash report or gameplay complaint — the Assert usually pointed at the exact mechanism (a consistency check failing, a safety cap being hit, a stuck flag), which made root-causing them tractable instead of guesswork.

A common thread runs through a good chunk of this batch: **stale state that quietly drifts out of sync with reality, then trips an Assert once it's used**. `Astar::DecayOrtho` re-decayed a node's *old* stored cost instead of the freshly computed one; `Diplomat::m_desireWarWith` cached a value from before a player slot got reused; the Sue/UndergroundRailway flag stayed set because it was cleared asynchronously instead of synchronously; `ForceDirectorSelect` got silently ignored because of a stale "selected since" flag; and the `/rnd` loop advanced turns before its own queued events had actually drained. In each case, the fix is the same shape — refresh or synchronously update the state at the point it actually changes, instead of letting a stale copy linger until something downstream asserts on the mismatch. A second, smaller cluster is genuine scheduler/pathfinding logic bugs (the outbidding margin, the ZOC visibility change, the transport reporting gaps) found the same way, by reading what the Assert-adjacent diagnostics logged.

## Scheduler / goal-matching fixes

- Fix stuck `k_UDF_USED_SPECIAL_ACTION_JUST_NOW` flag for Sue and UndergroundRailway special actions (made `AddSpecialActionUsed` synchronous instead of deferred to an async GEV handler).
- Fix `Gold::AddIncome` assert when a converted city's net gold goes negative — branch on sign before calling `AddIncome`/`SubIncome`, matching the existing idiom elsewhere in `CityData.cpp`.
- Require a decisive lead before `Scheduler::Match_Resources` evicts an already-committed agent from a goal — without a margin, two near-tied goals sharing the same one or two agents could evict each other's committed agent over noise-level score changes (confirmed via logs: two `GOAL_ESTABLISH_EMBASSY` goals endlessly swapping the same two diplomats, neither ever actually moving).
- Refresh `Diplomat::m_desireWarWith` cache on player slot reuse — a reused player slot could leave a stale cached value from whoever previously occupied it, tripping a consistency assert.
- Fix `ForceDirectorSelect` being silently ignored whenever a manual selection had happened previously that turn.

## `/rnd` turn-advance and diagnostics

- Fix `/rnd`'s auto-turn loop advancing before its own events drain (also wait on `GameEventManager::EventsPending()` and `DirectorImpl`'s action queue, not just player sync).
- Defer, rather than drop, `AddBeginScheduler` for a not-yet-current player.
- Rename `DIAG:` log prefixes to describe what they actually diagnose (`SPECIAL_ACTION:`, `PLAYER_SYNC:`), and move the `/rnd` `PLAYER_SYNC` diagnostics to `k_DBG_SCHEDULER_DETAIL` since they were flooding the log.
- Keep the previous turn's autosave around as a debugging aid.

## Pathfinding: robot ZOC and transport logic

- Let robot (AI) pathfinding ignore Zone-of-Control on tiles it can't currently see — a human player wouldn't have that omniscient knowledge either, and the real, non-visibility-based ZOC rule still applies at actual move-execution time as a safety net (`ArmyData::MoveIntoCell`), so an army that turns out to be wrong just gets its orders cleared and re-planned, same as it always did for occupied destinations.
- Split `Agent::FindPathToBoard` into task-specific wrappers (`FindPathToPickUpCargo`, `FindPathToBoard`, `FindPathToGoalWhileLoaded`), and rename `move_intersection` to `move_union` in `GotoGoalTaskSolution` (it's computed via OR, not AND).
- Replace duplicated city-or-field target-name logic with `Goal::GetTargetName`.
- Add an Inspect/Continue message box for `GotoGoalTaskSolution` pathfind failures, for interactively inspecting a stuck army/goal without derailing the rest of the game.
- Report when a transport is already in position to board cargo — two silent-return paths in `Goal::GotoTransportTaskSolution` never called the code that shows on-screen status text, making a correctly-waiting transport look idle.
- Log diagnostics for `GotoGoalTaskSolution`'s `FindPathToBoard` failures, later enriched with goal name, target unit count, and land/water/adjacency info.

## A* engine

- Fix `Astar::DecayOrtho` decaying a node's stale stored cost instead of the freshly computed one — this made every reopen of an orthogonal neighbor shrink its cost by another 5% regardless of the real edge cost, always looking like an improvement and reopening the same tile until it hit the reopen-count safety cap.
- Scale the reopen-count safety cap by the *square* of the search distance (first tried linear scaling, which wasn't enough headroom) — a long search over open, uniform-cost terrain legitimately ties a huge number of tiles, so multiple tiles getting reopened dozens of times is normal, not a runaway.
- Filter `k_DBG_ASTAR` diagnostics by player and army/city, not just by debug channel — added `Astar::GetOwner()`/`GetArmyId()` virtuals, overridden in `UnitAstar`, `CityAstar`, and `TradeAstar`, so a single debug player/army/city can be isolated instead of drowning in every player's every search.

## Misc

- Keep the AI from ordering armies that can't entrench (e.g. ships) to entrench anyway — a pre-existing `CanEntrench()` guard had been commented out; restored it, and dropped the disabled sleep fallback since sleep is a human-UI convenience the AI scheduler never checks.
- Document `--enable-ffmpeg4movies` in the Linux build instructions — movie playback silently does nothing without it, which looked like a bug rather than a missing build flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,8 @@ CFLAGS="$CFLAGS -O0 -fuse-ld=gold" CXXFLAGS="$CXXFLAGS -O0 -fuse-ld=gold" ./conf
 make -j$(nproc)
 ```
 
+Movie playback (e.g. the intro movie) requires passing `--enable-ffmpeg4movies` to `configure`, in addition to having the ffmpeg libraries installed - it is off by default because the FFmpeg API still changes a lot. Without it, the game builds and runs fine, it just won't play any movies.
+
 If you want to build a debug version it is:
 
 ```
--- a/ctp2_code/ai/ctpai.cpp
+++ b/ctp2_code/ai/ctpai.cpp
@@ -2196,21 +2196,21 @@ void CtpAi::ExecuteOpportunityActions(const PLAYER_INDEX player)
 		if (army->NumOrders() > 0)
 			continue;
 
-	//	if(army.CanEntrench())
-	//	{
+		if(army.CanEntrench())
+		{
 			// We need to find something more interesting to do here
 			g_gevManager->AddEvent( GEV_INSERT_AfterCurrent,
 									GEV_EntrenchOrder,
 									GEA_Army, army.m_id,
 									GEA_End);
-	/*	}
-		else
-		{
-			g_gevManager->AddEvent( GEV_INSERT_AfterCurrent,
-									GEV_SleepOrder,
-									GEA_Army, army.m_id,
-									GEA_End);
-		}*/
+		}
+		// No sleep fallback here: this function only ever runs for robot
+		// players (see the IsRobot() check above), and sleeping is purely
+		// a human-UI convenience (skip this unit in the "next unmoved
+		// unit" cycle) that also hides the unit on the map. The scheduler
+		// re-evaluates every army's goals every turn regardless of sleep
+		// state, so it buys the AI nothing - just leave a non-entrenching
+		// idle army without a new order.
 	}
 
 	CtpAi::SpendGoldToRushBuy(player);
--- a/ctp2_code/ai/ctpai.cpp
+++ b/ctp2_code/ai/ctpai.cpp
@@ -1059,6 +1059,18 @@ void CtpAi::RemovePlayer(const PLAYER_INDEX deadPlayerId)
 	}
 
 	AgreementMatrix::s_agreements.ClearAgreementsInvolving(deadPlayerId);
+
+	// The agreement-matrix clear above can change what ComputeDesireWarWith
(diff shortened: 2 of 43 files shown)


MartinGuehmann:
So two weekends have passed, time to merge.

PR #456, opened 2026-08-17
Fix a wide range of AI-related assertion failures

MartinGuehmann:
## Summary

This branch is the result of a systematic investigation into assertion failures, predominantly in the AI. Two sources fed the investigation: a full Windows Final/Logging build's debug log output (306 log files, 1617 total assertion lines, deduplicated by file:line and ranked by frequency, worked through systematically from most to least frequent), and separately, assertions found in Linux debug/logging builds that were picked out for investigation directly rather than via that same frequency-ranked process. Each assert was investigated in turn: root-caused and fixed where the mechanism could be confirmed, or given a targeted diagnostic (logging the exact state at the point of failure) where it couldn't yet be confirmed from static analysis alone, so the next real occurrence gives the data needed to finish the fix.

## Major fix groups

- **`Squad_Strength` cache desync** (`3c4bcb654`, `fb43490f9`, `dac25c190`, `9fb5c8d2a`, `74f5ab40f`, `d42020478`): several distinct bugs in how `Goal::m_current_attacking_strength` stays in sync with an agent's cached squad strength across `Commit_Agent`/`Rollback_Agent` - double-adding, stale-cache refreshes, and turn-boundary staleness, all producing `Assert(NothingNeeded())` failures.
- **Unit/army removal idempotency** (`975a14df8`, `9fb5c8d2a`, `ec1b052bc`, `ac31ad158`, `1d79d1be2`): `Unit::KillUnit` made idempotent against redundant kill events; `CTP2Combat::AddDeadUnit` guarded against double-insertion; full insert/remove audit-trail logging added to `World`'s unit reference tracking.
- **Refueling** (`4737f7102`, `dbed60b76`, `1f82bd002`, `91a0a2f5b`): `GetNearestRefuel` now actually honors its own fuel-range check on return instead of ignoring it; added per-army consecutive-failure tracking for `RefuelAirplane`'s pathing failures.
- **City garrison accounting** (`9c6638548`, `229fa12b6`, `3203d5276`): `CtpAi::ComputeCityGarrisons`'s consumer now fully recomputes a city's garrison instead of an incremental subtract that could underflow.
- **Goal lifecycle** (`fdc167d04`, `07d198ec8`): goals are now retired when their target city/army dies outright (e.g. flood) instead of lingering and re-tripping `Get_Target_Pos`'s dead-target assert every subsequent turn.
[...]

Issue #334, opened 2020-07-04
Underwater cities can currently be invaded from troop ship


LynxAbraxas:
Underwater cities can currently be invaded with marines from troop ships:
![ss_2020-07-05_00:43:02](https://user-images.githubusercontent.com/21012234/86522213-d2ba9a00-be5a-11ea-81b0-e2a2f13f18c3.png)
However, considering the text in the GL
https://github.com/civctp2/civctp2/blob/f4cc17172b67fab52e6c3ceeba16ad94651adc05/ctp2_data/english/gamedata/Great_Library.txt#L7500-L7506
it seems that should only be possible when using a crawler for transporting. Otherwise the low movement points of a crawler hardly make it any interesting.

PR #457, opened 2026-08-21
Allow AI transporters to reach and unload at tunnel-connected undersea cities

MartinGuehmann:
Branched from `FixTheAI6`, which has its own separate, not-yet-merged PR (#456). This PR should not be merged before that one — merging it will also bring in #456's commits.

## Summary

Undersea cities connected to a tunnel improvement were effectively unreachable for the AI: Transporters could neither find path to the tunnel nor unload their cargo there. This was tracked down and fixed through a chain of pathfinding, continent-tracking, and order-conversion bugs, confirmed via a mix of playtest-log analysis and live testing.

- A tunnel tile is simultaneously land and water, which the old single-value continent field couldn't represent. Added a `ContinentIDs` struct tracking land and water continent numbers separately per cell, so the tunnels are correctly classified as land or water depending whether water or land part of the path is calculated. Also fixed `PretestDest_SameWaterContinent`'s movement-type gate, broken by an earlier refactor into checking the wrong bits entirely.
- Fixed several `TransportPathCallback` bugs this exposed: Cargo standing on a tunnel was mistaken for still being aboard ship (wrongly re-blocking the crossing); a destination could become structurally unreachable from every approach angle when the cargo couldn't capture/enter the target city directly (an exemption for this was tried, found to produce a bad result, and reverted in favor of fixing the real tunnel-vs-open-water confusion instead); and the tunnel-hop danger-cost surcharge was restricted to the tile actually adjacent to the destination, instead of re-applying on every hop along a long corridor.
- Added a diagnostic logging continent IDs on both sides of a move near the map's X-wrap seam, while investigating a transport that kept failing to path to a tunnel-connected city — ruled out as the cause, but kept since it's cheap and generically useful.
- Continent renumbering after a tunnel/canal is built now happens the same turn, right before the AI scheduler runs, instead of being silently skipped for anything other than city founding.
- Retired the legacy single-value continent field, now fully superseded by `ContinentIDs` (kept only as inert padding for save-file byte-layout compatibility).
- Non-submarine transporters can no longer deliver land units directly onto a water-tile city (civctp2/civctp2#334); the AI's own transporters selection now knows this and only offers submarine transporters (e.g. the Crawler) for such targets.
- Even after a valid path was found, the transport still wouldn't unload at a tunnel-connected undersea city — extended the existing "carrier can't enter, but its cargo can" unload-conversion (previously air-transport only) to water-tile targets.
- While comparing the sea-transport and airlift pathing callbacks for other gaps, ported over a real one: airlift transports had no preference for routing away from hostile units while still in flight, unlike sea transports.

## Also included

Found via the same playtest-log/live-testing work, not tunnel-specific:

- Skip `KillATrader()` when a player's trader pool is already empty (trade-route cancellation assert).
- Set `found` in the airfield branch of `GetNearestRefuel` — was silently false even when an in-range airfield existed.
- `StrategyRecord`'s garrison-count fields now have real defaults instead of leaving callers' locals uninitialized when a strategy doesn't set them.
- Two `configure.ac` warnings and a sign-compare warning fixed.
- Debug cell text (e.g. `/showconts`) now draws on top of tile improvements/cities so that this info is not hidden by tile improvements or cities anymore.


MartinGuehmann:
OK, then it is time to merge.

